PR #26, opened 2015-03-24
Aws init

granders:
This pull request adds logic to automate some of the steps previously outlined in the readme.

This logic also makes it easier to create a docker image for driving tests from Jenkins.


ewencp:
@granders as long as you verify this works from scratch on a freshly allocated instance, it LGTM.

PR #27, opened 2015-03-31
Move tests

granders:
For the most part, this just moves files to the muckrake repo.

Added setup.py and uploaded to pypi.


setup.py
```diff
@@ -0,0 +1,12 @@
+from setuptools import find_packages, setup
+
+setup(name="ducktape",
+      version="0.1",
+      description="Distributed system test tools",
+      author="Ewen Cheslack-Postava",
```
ewencp: Doesn't matter much, but I don't need to take *all* the blame here :)

ducktape/cluster/remoteaccount.py
```diff
@@ -153,3 +153,12 @@ def __str__(self):
             r += self.user + "@"
         r += self.hostname
         return r
+
+
+def http_request(url, method, data="", headers=None):
```
ewencp: Not crazy about this being here. Doesn't look like it has anything to do with the remote account functionality? Could we put it in a different utility module instead? Or better yet, maybe start a mixin class for HTTP/REST utilities since we'll probably need a bunch of them, and this can form the core?

ewencp:
@granders I left a couple of comments, but this looks good to merge.

In the future, you should wait to upload anything to PyPI until it's merged. It's obviously not a problem with this version which is very quickly going to be out of date anyway, but you should make sure when you push this that you also tag v0.1 in git (on the actual commit that was uploaded to PyPI, not the final merged version of this patch) and make sure to git push --tags.

PR #28, opened 2015-04-08
Added basic test discovery.

granders:
@ewencp 
A pass at reasonable test discovery behavior within the specified base directory.

To verify for now, I'm just using the following snippet in (any file really) in muckrake:

from ducktape.tests.test import TestLoader
import sys

    if __name__ == "__main__":
        logging.basicConfig(level=logging.DEBUG)
        args = sys.argv
        base_dir = sys.argv[1] if len(sys.argv) > 1 else "."

        print base_dir
        loader = TestLoader()
        test_classes = loader.discover(base_dir)

        print test_classes

### DIFF
--- a/ducktape/tests/test.py
+++ b/ducktape/tests/test.py
@@ -14,7 +14,12 @@
 
 from ducktape.cluster import VagrantCluster
 from ducktape.logger import Logger
+import importlib
 import logging
+import os
+import inspect
+import re
+
 
 class Test(Logger):
     """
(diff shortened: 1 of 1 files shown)


ducktape/tests/test.py
```diff
@@ -47,3 +52,55 @@ def run_standalone(cls):
...
+
+
+class TestLoader(Logger):
+
+    def is_test_file(self, file_name):
+        """test_*.py or *_test.py"""
+        pattern = "(^test_.*\.py$)|(^.*_test\.py$)"
+        return re.match(pattern, file_name) is not None
+
+    def discover(self, base_dir):
```
ewencp: A short doc about what is discovered (either here or at the class level) would be useful. Helps for anyone who is reading the code or might want to extend it a bit.

ducktape/tests/test.py
```diff
@@ -47,3 +52,55 @@ def run_standalone(cls):
...
+                path = os.path.abspath(os.path.join(pwd, f))
+                path_pieces = path[:-3].split("/")
+                while len(path_pieces) > 0:
+                    module_name = '.'.join(path_pieces)
+                    # Try to import the current file as a module
+                    try:
+                        module = importlib.import_module(module_name)
+                        self.logger.info("Successfully imported " + module_name)
+                    except Exception as e:
+                        self.logger.warn("Could not import " + module_name + ": " + e.message)
```
ewencp: This generates a lot of noise at the WARN level. This should probably be at DEBUG or even TRACE since it looks like this is going to attempt to load a bunch of things that aren't modules.

ducktape/tests/test.py
```diff
@@ -47,3 +52,55 @@ def run_standalone(cls):
...
+                        test_classes.extend(self.get_test_classes(module))
+                    except Exception as e:
+                        self.logger.warn("Error getting test classes from module: " + e.message)
+
+        self.logger.info("Found these test classes: " + str(test_classes))
+        return test_classes
+
+    def get_test_classes(self, module):
+        """Return list of any all classes in the module object."""
+        module_objects = map(lambda x: getattr(module, x), [obj_name for obj_name in dir(module)])
```
ewencp: Why [obj_name for obj_name in dir(module)] instead of just dir(module)?

You also might be able to just use the __dict__ attribute, i.e. module_objects = module.__dict__.values(). Not a bit deal, but might make the intent clearer.

ducktape/tests/test.py
```diff
@@ -47,3 +52,55 @@ def run_standalone(cls):
...
+        return filter(lambda x: self.is_test_class(x), module_objects)
+
+    def is_test_class(self, obj):
+        """An object is a test class if it's a leafy subclass of Test.
+        """
+        return inspect.isclass(obj) and issubclass(obj, Test) and len(obj.__subclasses__()) == 0
```
ewencp: Is there a potential ordering issue here with module import and checking __subclasses__? Since these methods are called as you're iterating over all the files, you may have only imported a subset of them when it's invoked. Does this work if, e.g., the file containing KafkaTest is processed before any of the tests that use it? You may be getting away with this right now if things are processed in alphabetical order because the camus tests pulll in everything else.

granders: Agreed - I changed this so that all modules are imported before test classes are pulled out. This helps simplify the code somewhat as well.

ewencp:
@granders Nice, much cleaner now. Eventually you'll probably want to move what's currently at INFO to DEBUG just because people don't expect test discovery to be logged normally (and for a big project that list of discovered test classes could get *very* large), but not critical now. LGTM.

PR #29, opened 2015-04-10
Test runner

granders:
@ewencp @Ishiihara 
This is probably ready for some initial feedback.


ducktape/__main__.py
```diff
@@ -0,0 +1,131 @@
+from ducktape.tests.test import TestLoader
```
ewencp: You can just comma-separate imports from the same module

ducktape/__main__.py
```diff
@@ -0,0 +1,131 @@
...
+from ducktape.tests.test import SerialTestRunner
+from ducktape.tests.test import SimpleStdoutReporter
+from ducktape.tests.test import SimpleFileReporter
+from ducktape.tests.test import TestSessionContext
+from ducktape.cluster.vagrant import VagrantCluster
+
```
ewencp: There should probably be some documentation in here somewhere, at least a brief comment describing what code is in here.

ducktape/__main__.py
```diff
@@ -0,0 +1,131 @@
...
+import os
+import sys
+import random
+import time
+
+
+def mock_test_method():
```
ewencp: Can we just move the mock stuff to an actual mock test, then have a very simple unit test for this? We'll probably want to remove the randomness, but then we could actually test that we're discovering and running the expected set of tests.

ducktape/__main__.py
```diff
@@ -0,0 +1,131 @@
...
+
+def main():
+    if len(sys.argv) != 2:
+        print "Usage: ducktape <directory_name>"
+        sys.exit(1)
+
+    if not os.path.isdir(".ducktape"):
+        os.makedirs(".ducktape")
+
+    test_session_id = get_test_session_id()
```
ewencp: Maybe move most of this into a dedicated function so it can be tested?

ducktape/tests/test.py
```diff
@@ -64,7 +50,10 @@ def discover(self, base_dir, pattern=DEFAULT_TEST_FILE_PATTERN):
         - Discover test classes within each test module. A test class is a subclass of Test which is a leaf
           (i.e. it has no subclasses).
         """
-        test_files = self.find_test_files(base_dir, pattern)
+        if os.path.isfile(base_dir):
```
ewencp: This is exactly the type of thing that could use a unit test -- just a simple dummy test that you can run either by specifying the parent module or the file directly and make sure both do the right thing.

ducktape/tests/test.py
```diff
@@ -137,3 +129,168 @@ def is_test_class(self, obj):
         return inspect.isclass(obj) and issubclass(obj, Test) and len(obj.__subclasses__()) == 0
 
 
+class TestReporter(object):
```
ewencp: You'll probably just want to move these classes into their own files, at least by type (Reporters, Loaders, etc).

ducktape/tests/test.py
```diff
@@ -137,3 +129,168 @@ def is_test_class(self, obj):
...
+class SimpleReporter(TestReporter):
+    def header_string(self):
+        header = ""
+        header += "Test run with session_id " + self.results.test_session_context.test_session_id + "\n"
+        header += self.pass_fail(self.results.get_aggregate_success()) + "\n"
+        header += "------------------------------------------------------------\n"
+
+        return header
+
+    def result_string(self, result):
```
ewencp: This probably doesn't matter much yet, but you probably want the lower level interface to be some sort of write-to-stream interface (i.e. write to a file) so you're not just collecting a ton of string data in memory. If you end up needing to collect it all in memory you can use something like StringIO to provide a file-like interface from which you can then grab the full output.

At a minimum, you probably want to change these to collect all the substrings in an array and then do a join. Concatenating like this gets expensive quickly since all the data is copied every time the string is extended.

ducktape/tests/test.py
```diff
@@ -137,3 +129,168 @@ def is_test_class(self, obj):
...
+
+        return True
+
+    def __iter__(self):
+        for item in self.results_list:
+            yield item
+
+
+class TestRunner(Logger):
+    DEFAULT_TEST_RUN_ID_FILE = ".ducktape/test_run_id"
```
ewencp: I saw this in another file somewhere. I think this needs to be defined in one place and then all the code that uses the value needs to get the setting from the same place.

setup.py
```diff
@@ -6,6 +6,9 @@
       author="Ewen Cheslack-Postava",
       author_email='ewen@confluent.io',
       platforms=["any"], 
+      entry_points={
+        'console_scripts': ['ducktape=ducktape.__main__:main'],
```
ewencp: I actually haven't setup a script like this, so I'm not sure, but can we/would it be better to put it under its own module instead of in the `__init__.py` file for the ducktape module? Usually it's not a great idea to put too much code into an `__init__.py` file because they're really intended for configuration of the module and so that code isn't very easily discoverable. I'm not sure how the location interacts with this console_scripts setting, but it looks like we could probably put that code anywhere we want.

granders: Hm maybe I'm missing something - how is this code block related to __init__.py?

ducktape/__main__.py
granders: For some reason the PEP style guide recommends newlines for each:
https://www.python.org/dev/peps/pep-0008/#imports

setup.py
ewencp: Ooops, I read that too fast. I was just wondering what the "right" way to make these types of commands work. Is putting a `__main__.py` in the top-level module the normal way to expose a command? I know that lets you do `python <module_name>`, but I wasn't sure if that was the common pattern for exposing things you want to run as just `<module_name>` (or really `<arbitary_binary_name>`).

ducktape/__main__.py
ewencp: I was referring to the second half of the first rule. Quoting the PEP:

> It's okay to say this though:
> 
> `from subprocess import Popen, PIPE`

So I was just saying it might be nicer to have

```
from ducktape.tests.test import TestLoader, SerialTestRunner, SimpleStdoutReport, SimpleFileReporter, TestSessionContext
```

In any case, it's not a big deal, unless we adopt a standard format, different people's IDEs are probably going to end up formatting stuff like this in a variety of ways.

setup.py
granders: I'm fairly sure it can be an arbitrary module.

I guess the advantage is that by using __main__.py, I can run ducktape with either command
python -m ducktape
ducktape

ewencp: Ok, I think I'd just recommend leaving it as is, but trying to keep the amount of code in `__main__.py` from now on.

ducktape/__main__.py
granders: Oops gotcha I misread your original comment.

granders:
@ewencp 
Latest push is mostly refactoring.

Moved logic from __main__ to main.py (__main__ now makes a call to command_line.main.main() to keep __main__ as small as possible)

Broke out the classes in test.py into multiple modules.

Added very simple unit test - configured discovery patterns to use "check" instead of "test" to reduce potential confusion.

ducktape/unit_tests/tests/check_loader.py
```diff
@@ -0,0 +1,40 @@
...
+
+from ducktape.tests.loader import TestLoader
+
+import os
+
+
+class CheckTestLoader(object):
+
+    # Assume base_dir is the unit_test directory
+    BASE_DIR = os.path.abspath(os.path.curdir)
```
ewencp: I noticed you can't run the tests from another directory, and I'm assuming this is why. It'd be better to get this `BASE_DIR` variable dynamically, either from pytest or assuming an offset from this file.

ewencp:
@granders Left one more minor comment, but otherwise LGTM.

PR #30, opened 2015-04-15
Clean up unit test setup.

ewencp:
Move unit tests to their own top-level tests/ directory so we don't end up with
any naming conflicts. Move the loader tests from ducktape/unit_tests/tests/ to
tests/loader/. Future tests can be categorized in directories under
tests/. Setup tests to be runnable via setup.py, which will also auto-install
pytest if necessary. Update README and simplify the loader test so it only
relies on the layout of the particular test files instead of the entire
repository layout.

@granders, this is pretty small if you ignore moving all the files around, makes the tests work in a more standard way (and isolated from the main source code). Only impact on outstanding work should be moving any new tests to the right place.


tests/loader/check_loader.py
```diff
@@ -14,31 +14,11 @@
...
-    root_ducktape_path = "/".join(pwd_vector)
-    return os.path.abspath(root_ducktape_path)
-
+import os.path
 
 def discover_dir():
     """Return the absolute path to the directory to use with discovery tests."""
-    base_dir = find_ducktape_base_dir()
-    return os.path.join(base_dir, "ducktape", "unit_tests", "tests", "resources", "loader_test_directory")
+    return os.path.join(os.path.dirname(os.path.abspath(__file__)), "resources", "loader_test_directory")
```
granders: Aha, I knew there had to be a better way...

granders:
Thanks for the cleanup! These changes should also merge fairly easily into my branch. LGTM

PR #31, opened 2015-04-15
Log output

granders:
@ewencp @Ishiihara 

I decided to keep this pull request smallish by only including logic to pipe test logs and session logs to appropriate places in the file system. Service log collection will be in another branch. To try these changes against muckrake, you'll need to check out the log-output branch in the muckrake repo.

I converged on using "context" classes as a mechanism for passing around loggers. There are now three context classes: TestContext, for the smallest testable unit, SessionContext for 'global context', ServiceContext for individual service instances.

One sneaky bit - Service instances set the the logger attribute on node.account even though these are 'global' variables shared through the session_context.cluster. So when nodes are allocated to a Service, the node.account logging behavior matches the logging behavior of the Service to which it's allocated.


ducktape/command_line/main.py
```diff
@@ -15,11 +15,11 @@
 from ducktape.tests.loader import TestLoader
 from ducktape.tests.runner import SerialTestRunner
 from ducktape.tests.reporter import SimpleStdoutReporter, SimpleFileReporter
-from ducktape.tests.session_context import TestSessionContext
+from ducktape.tests.session_context import SessionContext
 from ducktape.cluster.vagrant import VagrantCluster
 from ducktape.command_line.config import ConsoleConfig
 from ducktape.tests.session_context import generate_session_id, generate_results_dir
-from ducktape.unit_tests.mock import swap_in_mock_run, swap_in_mock_fixtures
+from tests.mock import swap_in_mock_run, swap_in_mock_fixtures
```
ewencp: Are these going to be permanent or are these mock things going away sometime soon? If they're permanent we should move them out of the tests/ module.

ducktape/command_line/main.py
```diff
@@ -60,15 +60,23 @@ def main():
     # in which all test results will be stored
     session_id = generate_session_id(ConsoleConfig.SESSION_ID_FILE)
     results_dir = generate_results_dir(session_id)
-    test_session_context = TestSessionContext(session_id, results_dir)
+    cluster = VagrantCluster()
```
ewencp: Not related to this patch, but should we file an issue to add command line options to choose the cluster type? Not immediately necessary for us, but we'll need it before we can suggest using this in Kafka. This type of small task seems like the kind of thing we should record issues for as we encounter them to make sure they aren't forgotten.

ducktape/command_line/main.py
```diff
@@ -60,15 +60,23 @@ def main():
...
+
+    # Make directory in which results will be stored
     if os.path.isdir(results_dir):
         raise Exception(
             "A test results directory with session id %s already exists. Exiting without overwriting..." % session_id)
     os.mkdir(results_dir)
+    latest_test_dir = "latest-test-results"
```
ewencp: This is definitely a bike-shedding comment, so feel free to ignore, but do we have to append "-test-results" to everything? If we want to label all this stuff as being test results, why not group it all under a single directory called "results" or "test-results", i.e. so the latest one would be something like results/latest symlinked to results/2015-04-15--1. Also has the benefit of being easier to nuke all your old test results.

ducktape/services/service.py
```diff
@@ -25,18 +25,24 @@
...
+        self.cluster = service_context.cluster
+        self.logger = service_context.logger
 
     def start(self):
         """Start the service running in the background."""
         self.nodes = self.cluster.request(self.num_nodes)
         for idx, node in enumerate(self.nodes, 1):
+
+            # Remote accounts utilities should log where this service logs
+            node.account.set_logger(self.logger)
```
ewencp: Should there be a corresponding call in stop() or some other method that clears out the logger. That would offer some opportunity for doing error checking by testing whether the logger was already set (which could signal some race or framework error when we start doing parallel tests, or some user error if they aren't cleaning up properly; in the latter case, we might actually want set_logger() to pass in extra information so that if we catch an error we can also log which services/tests are conflicting).

ducktape/tests/runner.py
```diff
@@ -13,24 +13,75 @@
...
+            raise
+
+
+def get_test_case(test_class, session_context):
+    """Create test context object and instantiate test class.
+    :type test_class: ducktape.tests.test.Test.__class__
+    :type session_context: ducktape.tests.session_context.SessionContext
+    :rtype test_class
+    """
+    session_context.logger.debug(str(session_context))
```
ewencp: Should there be actual messages here, or are these just leftover from debugging?

ducktape/tests/session_context.py
```diff
@@ -83,4 +134,75 @@ def generate_results_dir(session_id):
     :type session_id: str
     :rtype: str
     """
-    return session_id + "-test-results"
\ No newline at end of file
+    return session_id + "-test-results"
+
+
+def mkdir_p(path):
```
ewencp: This function is duplicated. It should probably be refactored into some utils module. That module should probably be clearly named to indicate it's for local utilities since the stuff in RemoteAccount and utilities like this are going to be easy to mix up.

ducktape/tests/session_context.py
```diff
@@ -83,4 +134,75 @@ def generate_results_dir(session_id):
...
+    try:
+        os.makedirs(path)
+    except OSError as exc: # Python >2.5
+        if exc.errno == errno.EEXIST and os.path.isdir(path):
+            pass
+        else:
+            raise
+
+
+class TestContext(Logger):
```
ewencp: It's a bit odd for this to be in session_context.py. If you want to group things together to avoid the proliferation of files, I'd group TestContext into test.py and rename this file session.py (which makes it available for any other small session-related classes that we might end up creating).

ducktape/tests/session_context.py
```diff
@@ -83,4 +134,75 @@ def generate_results_dir(session_id):
...
+        self.logger.setLevel(logging.DEBUG)
+
+        mkdir_p(self.results_dir)
+        fh = logging.FileHandler(os.path.join(self.results_dir, "test_log"))
+        fh.setLevel(logging.DEBUG)
+        # create console handler with a higher log level
+        ch = logging.StreamHandler(sys.stdout)
+        ch.setLevel(logging.INFO)
+        # create formatter and add it to the handlers
+        formatter = logging.Formatter('[%(levelname)-6s - %(asctime)s - %(module)s - %(funcName)s - lineno:%(lineno)s]: %(message)s')
```
ewencp: Was this formatting constant the same as the one before? Should it only be defined in one location that would be reasonable for everyone to use (e.g. a SessionContext class field, since SessionContext is for globals anyway? or SessionContext instance field, which can be used here since it has self.session_context)?

ewencp:
@granders This looks good. One comment that didn't fit anywhere in the code you wrote: what about unit tests? It's much better to have simple unit tests here which can be run quickly rather than having to rely on running muckrake stuff (and we need a reasonable way for other people to contribute, which means having tests they can make sure they don't break if they submit patches). For this, it should just be a matter of writing a very simple test + service that do some simple logging and then the unit test runs the test and verifies a) the layout of the directories and b) that the right stuff got logged to the right files.

ewencp:
Oh, one other item -- how about updating the README? Just a quick description under a section about how to run the tests that describes the layout of the test output?

ducktape/command_line/main.py
granders: My intention was for this to be temporary, but it's a pretty useful sanity check, so arguably it could ultimately remain.

ducktape/command_line/main.py
granders: Agreed

ducktape/tests/runner.py
```diff
@@ -41,34 +92,34 @@ def run_all_tests(self):
...
-            print self.__class__.__name__ + ": running " + test.__class__.__name__
+            self.logger.info(self.__class__.__name__ + ": running " + test.__class__.__name__)
             test.run()
+            self.logger.info(self.__class__.__name__ + ": successfully ran " + test.__class__.__name__)
         except Exception as e:
             raise e
         finally:
             if hasattr(test, 'tearDown'):
-                print self.__class__.__name__ + ": tearing down " + test.__class__.__name__
+                self.logger.inf(self.__class__.__name__ + ": tearing down " + test.__class__.__name__)
```
ewencp: Found this error when testing with muckrake  -- inf -> info. Good candidate for more unit testing :)

ewencp:
@granders also looks like the changes to RemoteAccount (adding set_logger) got lost?

ducktape/services/service.py
granders: added

ducktape/tests/runner.py
granders: deleted - just debugging stuff

ducktape/tests/session_context.py
granders: moved, created 'local_filesystem_utils' module

ducktape/tests/session_context.py
granders: seems reasonable - moved

ducktape/tests/session_context.py
granders: Moved to command_line.config module

I'm not super happy about the current state of log configuration, but maybe this is ok for now.

ducktape/tests/runner.py
granders: got it

ducktape/services/service.py
granders: Interestingly this immediately caught the fact that no Service subclass was calling super(...).stop() in its own stop method

ducktape/command_line/main.py
granders: removed "-test-results" but punted on grouping - placed this in issue #37

README.md
```diff
@@ -9,6 +9,39 @@ triggering special events (e.g. bouncing a service), collect results (such as
...
+            log.info
+            log.debug
+        service_logs - log(s) collected from service nodes go here
+            <service_name>[__<num>]/<instance_id>  - e.g. kafka_service__1/3 - means 1st kafka service in test on
+            metadata
+    ...
+```
+
+Test Discovery
+--------------
```
ewencp: Is this intentionally empty?

README.md
```diff
@@ -87,7 +120,7 @@ fields to get the information they want.
 Unit Tests
 ----------
 
-You can run the tests via the setup.py script:
+It's a good idea to write and run unit tests on the ducktape framework itself. You can run the tests via the setup.py script:
```
ewencp: Yes it is :) We might want to split the readme into user and developer sections since right now they're kind of mixed together.

ducktape/services/service.py
```diff
@@ -25,18 +25,29 @@
...
         """Start the service running in the background."""
         self.nodes = self.cluster.request(self.num_nodes)
         for idx, node in enumerate(self.nodes, 1):
+
+            # Remote accounts utilities should log where this service logs
+            if node.account.logger is not None:
+                raise RuntimeError(
+                    "logger was not None on service start. There may be a concurrency issue, " +
+                    "or some service which isn't properly cleaning up after itself. " +
+                    "Service: %s, node.account: %s" % (self.__class__.__name__, str(node.account)))
```
ewencp: This tells you which Service ran into the problem, but doesn't necessarily help you track down the offending Service/Test. How about logging something to node.account.logger at critical and mentioning that in this message, e.g. say that you can grep the logs for 'MISBEHAVING TEST' or something so they know how to track down the problem?

ewencp:
Finally verified on some real tests, so this LGTM.

PR #36, opened 2015-04-16
Add HTML reporter

Ishiihara:
@ewencp @granders This PR adds the HTML reporter to ducktape. Note that the running time and log file are currently mocked and will be modified once log patch in committed.

### DIFF
--- a/ducktape/command_line/main.py
+++ b/ducktape/command_line/main.py
@@ -14,7 +14,7 @@
 
 from ducktape.tests.loader import TestLoader
 from ducktape.tests.runner import SerialTestRunner
-from ducktape.tests.reporter import SimpleStdoutReporter, SimpleFileReporter
+from ducktape.tests.reporter import SimpleStdoutReporter, SimpleFileReporter, HTMLReporter
 from ducktape.tests.session_context import TestSessionContext
 from ducktape.cluster.vagrant import VagrantCluster
 from ducktape.command_line.config import ConsoleConfig
@@ -88,6 +88,10 @@ def main():
     reporter = SimpleStdoutReporter(test_results)
     reporter.report()
     reporter = SimpleFileReporter(test_results)
+    reporter.report()
+
+    # Generate HTML reporter
+    reporter = HTMLReporter(test_results)
     reporter.report()
 
     if not test_results.get_aggregate_success():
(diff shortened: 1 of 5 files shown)


ewencp:
LGTM once the changes you mentioned are made. One question I have is whether we plan to keep both the HTMLReporter and SimpleFileReporter? If we do, HTMLReporter will probably always be dependent on the SimpleFileReporter since the html will link to files generated by the SimpleFileReporter. Not necessarily a problem, but it does mean that an option that controls the type of reporter would be more complicated to implement (needs to know implementation details of the classes rather than just know how to instantiate and use one of them specified by name).

granders:
Thanks @Ishiihara!
@ewencp Right now as far as I can tell, these two reports are generating independent views of the same data. As I understand it the links will point to the log files generated while the tests run. Log files are generated independently, and are written regardless of which reporter is used.

Perhaps 'SummaryReporter' would be a more accurate name for the class.

granders:
@Ishiihara @ewencp 
Paths to the individual per-test log files are only available in my current pull request (log-output branch) - perhaps it makes sense to just merge this and then link things together once I merge in log-output?

ewencp:
Ah right, that makes sense since the logger is just setup to pipe to a file.

I think this setup will work for now, but it could be a bit brittle -- the report will rely on details implemented elsewhere with no way to know if they change. Would it make sense to file an issue to later address this by to better integrating these two? I'm thinking that the right solution might be to let the reporter control all the output, which in this case would mean that when loggers are setup, that process is actually delegated to the Reporter so it can control what happens with it (e.g. gets put on disk on its own, gets embedded in an HTML report, etc.). On the other hand, this could make the Reporter have to interact with a bunch of stuff since the location of service files that are downloaded from worker nodes would also logically be under the Reporter's control since it also needs to link to them.

ewencp:
@granders Up to you how these get merged and linked up since right now it's not really a problem to have not-quite-complete features checked in.

granders:
@ewencp Hm I think I see your point - you seem to be envisioning the reporter as something which intercepts and routes all messages, whereas I had originally conceived of it as strictly a 'summarize results' mechanism, with a clear enough interface on the TestResult class to make it simple to do so.

 Yeah maybe file an issue - it's certainly worth discussing the tradeoffs.

PR #39, opened 2015-04-21
Verify input paths exist during test discovery.

ewencp:
Print an error and exit with a non-zero status code to indicate that the tests
failed since specifying a non-existent path is probably an error on the user's
part.

### DIFF
--- a/ducktape/tests/runner.py
+++ b/ducktape/tests/runner.py
@@ -27,7 +27,7 @@ class TestRunner(object):
         self.results = TestResults(self.session_context)
         self.logger = session_context.logger
 
-        self.logger.debug("Instantiating %s" + self.__class__.__name__)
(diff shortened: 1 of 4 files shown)


ewencp:
@granders Should be a trivial review and I think addresses the issue in confluentinc/muckrake#19.

ducktape/tests/runner.py
```diff
@@ -27,7 +27,7 @@ def __init__(self, session_context, test_classes, cluster):
         self.results = TestResults(self.session_context)
         self.logger = session_context.logger
 
-        self.logger.debug("Instantiating %s" + self.__class__.__name__)
+        self.logger.debug("Instantiating " + self.__class__.__name__)
```
granders: Thank you!

granders:
LGTM

PR #40, opened 2015-04-24
Add support for jinja2 templates.

ewencp:
Templates can either be rendered directly from strings or from files loaded from
a templates/ directory relative to the class. This is integrated and tested with
Service and Test classes, so both can provide templates/ directories to collect
template files alongside the services/tests themselves.


ewencp:
@granders This provides pretty barebones but functional templating for services and tests. jinja2 is way nicer than manually loading files yourself and then using string interpolation since it gives you a bunch of other logic and filters, this setup lets you get access to any fields in the service/test without having to build up a set of template parameters, and we can set this up easily to automatically group template files relative to the service/test file.

One thing I wanted to do but couldn't figure out how to make work is allowing tests to override service templates just by providing a new template with the same name in its own templates directory. This is doable with by using Loaders that search through a series of directories, but requires services to have some context from the test and passing it through the context seemed to violate your attempt to keep those two isolated. Not critical to get that feature in now though, it's just convenient sometimes since it lets you override some configs on a per-test basis without requiring the Service to parameterize everything for you.

tests/templates/service/__init__.py
```diff
@@ -0,0 +1 @@
+__author__ = 'ewencp'
```
granders: It's handy to update the template so this doesn't crop up all the time (you can stick the copyright notice here)

Pycharm -> preferences ->  editor -> file and code templates -> python script

ewencp: Fixed.

ducktape/services/service.py
```diff
@@ -26,13 +26,14 @@
...
 
-
-class Service(object):
-    def __init__(self, service_context):
+class Service(TemplateRenderer):
+    def __init__(self, service_context, *args, **kwargs):
         """
         :type service_context: ServiceContext
         """
+        super(Service, self).__init__(*args, **kwargs)
```
granders: Is this necessary? Isn't TemplateRenderer essentially a mixin?

granders:
I find it a little disconcerting to add TemplateRenderer functionality to Service and Test by inheriting from TemplateRenderer, but I guess it's essentially just a mixin so I don't have any major beef with the approach.

LGTM

ducktape/services/service.py
ewencp: super() invokes the next constructor in the MRO. This can be affected by the exact set of classes and order the a subclass inherits from this one.  It seems to be a matter of preference whether using super() and multiple inheritence is a good idea. [Here](https://rhettinger.wordpress.com/2011/05/26/super-considered-super/) is one article that gives practical advice about how to use it (this pattern, combined with the keyword-arg stripping approach + the possibility of another mixin/base class in a subclass of Service should explain why I included this here).

This could go either way. There are also people who think it's a [bad idea to use it](https://fuhm.net/super-harmful/). The motivation for using mixins here are two-fold. First, I'm imagining that Service could easily become a behemoth of utility methods, most of which doesn't need to assume much about the class it is mixed into. A lot of those could be kept as separate classes and mixed in as needed by the user, or just kept separate in order to keep things organized internally. I dislike delegation in this particular instance because it would require a ton of boilerplate code just to dispatch calls to the TemplateRenderer we stored as a field. Second, using mixins allows reuse of that code, which I did here by allowing both services and tests to have templates.

Another alternative would be to make TemplateRenderer a standalone class that has to be instantiated to use templates. They just seem like they'll be needed for pretty much every Service, so it seemed worth including via an easy to use method call rather than having to instantiate a different TemplateRenderer manually in every service you write.

granders: Thanks for the links.

Very minor comment, since I don't have a strong opinion here - perhaps just be consistent wrt to super call between Service and Test (Service __init__ method makes call to super, but Test.__init__ doesn't)

PR #41, opened 2015-04-24
Adjust sys.path to work automatically with tests passed on the command line.

### LINKED ISSUE #32: test discovery problem due to module not being in sys.path
@ewencp and @Ishiihara have both found ducktape unable to discovery tests when running in muckrake. 

The root of the problem was that <path_to_muckrake> was not in sys.path, so attempts to import muckrake modules failed. It's easy to work around this by explicitly setting PYTHONPATH to include <path_to_muckrake> but this would be pretty annoying for users.

Ewen's suggested approach seems simple and reasonable:
(in loader.py) maybe something like walking up the directories until we stop seeing __init__.py files (do this once for each input path), then add whatever set of paths result from that to sys.path?

ewencp:
Fixes #32.

### DIFF
--- a/ducktape/command_line/main.py
+++ b/ducktape/command_line/main.py
@@ -43,6 +43,22 @@ def parse_args():
     args = parser.parse_args()
     return args
 
+def extend_import_paths(paths):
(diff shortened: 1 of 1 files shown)


granders:
LGTM

PR #42, opened 2015-04-27
Add tests for cluster classes and remove java_home and kafka_home.

ewencp:
This also fixes a bug in the Vagrant parser caused by a typo Hostname instead of
HostName, which inadvertently caused SSH commands to work in an unexpected way.


ewencp:
@granders Just some more cleanup after splitting off ducktape/muckrake -- we need to fill in some unit tests for existing code and also finish cleaning out Kafka-specific stuff. This PR tries to address that for the cluster code.

granders:
Thanks for the cleanup. 

Tests pass and LGTM. 

I tried a small experiment but wasn't able to successfully run a muckrake test using LocalHostCluster, so I'm guessing there may be some debugging to do (e.g. some mechanism hand out different ports for a given service), but it would be really cool to be able to just fire up a bunch of machines on localhost.

ewencp:
@granders Yup, I realized that's an issue with the current muckrake tests. I think the only issue is the ports, although in muckrake it probably will also have issues with filesystem layout because those tests basically assume we're running in an Ubuntu VM at the moment.

I'm not sure about the right fix for muckrake. Randomly choosing ports is easy, but after that long bug in Kafka, I think it's just a bad idea unless you can guarantee the port range will be free. Besides that, not guaranteeing a dedicated host per service instance means you have to be able to parameterize other configs that could conflict by node id (e.g. storage space). Honestly, I personally think running these tests directly on localhost is generally a bad idea (especially if we keep steps like "kill all java processes" in place), but I also get that some people want that to make debugging a bit simpler.

granders:
@ewencp The assumption of isolation (with respect to processes, ports, directories etc) is pretty pervasive, so I wonder if it's even worth having LocalhostCluster at all?

ewencp:
Is it pervasive anywhere in ducktape or just in muckrake? muckrake probably isn't too horrible to fix because you can make them parameterized by node index -- after getting in templates, most of the work might be in editing the config templates. I'm actually more worried about the differing layouts of files, which seems like it's much harder to handle unless you make all your tests do everything within one fixed directory that is included in the cluster's config.

In any case, some of this is something we have to answer by asking the community -- I am not sure what some people will feel are requirements for these tests, especially if they want to be able to try to reproduce issues locally for debugging.

granders:
Good point - primarily muckrake, in the specific Service implementations.

ewencp:
@granders Turns out this merged very easily only a minor conflict due to some reformatting. Passes the ducktape tests and ran it over some muckrake tests too, so I'm just going to merge based on your previous LGTM.

PR #43, opened 2015-04-28
Service logs

granders:
Added mechanism to pull service logs from service nodes, but still need to add mechanism for toggling collection on/off.

Refactored Service to have a few more lifecycle methods:
allocate_nodes
start, start_node
wait_until_alive
stop, stop_node
clean, clean_node
free, free_node

This allows the test runner to have control over log collection, cleanup and cluster resource allocation even when Exceptions are thrown by the test.

Added ServiceRegistry class - Test classes must now 'register' services during construction. The test runner only allocates nodes to registered services.

Test gets a default tearDown method but no default setUp.

### DIFF
--- a/ducktape/cluster/remoteaccount.py
+++ b/ducktape/cluster/remoteaccount.py
@@ -163,10 +163,11 @@ class RemoteAccount(HttpMixin):
         try:
             subprocess.check_output(cmd, shell=True, stderr=subprocess.STDOUT)
         except subprocess.CalledProcessError as e:
-            if allow_fail:
-                return
             self.logger.warn("Error running remote command: " + cmd)
             self.logger.warn(e.output)
+
+            if allow_fail:
+                return
             raise e
 
     def __str__(self):
--- a/ducktape/cluster/remoteaccount.py
+++ b/ducktape/cluster/remoteaccount.py
@@ -52,7 +52,6 @@ class RemoteAccount(HttpMixin):
                             "Either the service failed to start, or there is a problem with the url. "
                             "You may need to open Vagrantfile.local and add the line 'enable_dns = true'.")
 
-
     def ssh_command(self, cmd):
         r = "ssh "
         if self.user:
@@ -150,8 +149,8 @@ class RemoteAccount(HttpMixin):
         except subprocess.CalledProcessError as e:
             if allow_fail:
                 return
-            print "Error running remote command: " + cmd
-            print e.output
+            self.logger.warn("Error running remote command: " + cmd)
+            self.logger.warn(e.output)
             raise e
 
     def __str__(self):
--- a/ducktape/cluster/remoteaccount.py
+++ b/ducktape/cluster/remoteaccount.py
@@ -89,6 +89,14 @@ class RemoteAccount(HttpMixin):
             self.ssh(cmd, allow_fail)
 
     def scp_from_command(self, src, dest, recursive=False):
+        if type(src) == list:
+            if len(src) == 0:
+                src = ""
+            elif len(src) == 1:
+                src = src[0]
+            else:
+                src = "\{" + ",".join(src) + "\}"
+
         r = "scp "
         if self.ssh_args:
             r += self.ssh_args + " "
@@ -104,6 +112,14 @@ class RemoteAccount(HttpMixin):
         return self._ssh_quiet(self.scp_from_command(src, dest, recursive))
 
     def scp_to_command(self, src, dest, recursive=False):
+        if type(src) == list:
+            if len(src) == 0:
+                src = ""
+            elif len(src) == 1:
+                src = src[0]
+            else:
+                src = " ".join(src)
+
         r = "scp "
         if self.ssh_args:
             r += self.ssh_args + " "
--- a/ducktape/cluster/remoteaccount.py
+++ b/ducktape/cluster/remoteaccount.py
@@ -31,7 +31,7 @@ class RemoteAccount(HttpMixin):
 
     @property
     def local(self):
-        "Returns true if this 'remote' account is actually local. This is only a heuristic, but should work for simple local testing."
+        """Returns true if this 'remote' account is actually local. This is only a heuristic, but should work for simple local testing."""
         return self.hostname == "localhost" and self.user is None and self.ssh_args is None
 
     def wait_for_http_service(self, port, headers, timeout=20, path='/'):
@@ -66,7 +66,7 @@ class RemoteAccount(HttpMixin):
         return self._ssh_quiet(self.ssh_command(cmd), allow_fail)
 
     def ssh_capture(self, cmd):
-        '''Runs the command via SSH and captures the output, yielding lines of the output.'''
+        """Runs the command via SSH and captures the output, yielding lines of the output."""
         ssh_cmd = self.ssh_command(cmd)
         proc = subprocess.Popen(ssh_cmd, shell=True, stdout=subprocess.PIPE, stderr=subprocess.STDOUT)
         for line in iter(proc.stdout.readline, ''):
@@ -89,36 +89,35 @@ class RemoteAccount(HttpMixin):
             self.ssh(cmd, allow_fail)
 
     def scp_from_command(self, src, dest, recursive=False):
-        if type(src) == list:
-            if len(src) == 0:
-                src = ""
-            elif len(src) == 1:
-                src = src[0]
-            else:
-                src = "\{" + ",".join(src) + "\}"
+        if self.user:
+            remotehost = self.user + "@" + self.hostname
+        else:
+            remotehost = self.hostname
+
+        if isinstance(src, basestring):
+            src = remotehost + ":" + src
+        else:
+            # assume src is iterable
+            # e.g. "ubuntu@host:path1 ubuntu@host:path2"
+            src = " ".join([remotehost + ":" + path for path in src])
 
         r = "scp "
         if self.ssh_args:
             r += self.ssh_args + " "
         if recursive:
             r += "-r "
-        if self.user:
-            r += self.user + "@"
-        r += self.hostname + ":" + src + " " + dest
+
+        r += src + " " + dest
         return r
 
     def scp_from(self, src, dest, recursive=False):
-        """Copy something from this node."""
+        """Copy something from this node. src may be a string or an iterable of several sources."""
         return self._ssh_quiet(self.scp_from_command(src, dest, recursive))
 
     def scp_to_command(self, src, dest, recursive=False):
-        if type(src) == list:
-            if len(src) == 0:
-                src = ""
-            elif len(src) == 1:
-                src = src[0]
-            else:
-                src = " ".join(src)
+        if not isinstance(src, basestring):
+            # Assume src is iterable
+            src = " ".join(src)
 
         r = "scp "
         if self.ssh_args:
@@ -158,9 +157,10 @@ class RemoteAccount(HttpMixin):
         os.remove(local_name)
 
     def _ssh_quiet(self, cmd, allow_fail=False):
-        '''Runs the command on the remote host using SSH. If it succeeds, there is no
-        output; if it fails the output is printed and the CalledProcessError is re-raised.'''
+        """Runs the command on the remote host using SSH. If it succeeds, there is no
+        output; if it fails the output is printed and the CalledProcessError is re-raised."""
         try:
+            self.logger.debug("Trying to run remote command: " + cmd)
             subprocess.check_output(cmd, shell=True, stderr=subprocess.STDOUT)
         except subprocess.CalledProcessError as e:
             self.logger.warn("Error running remote command: " + cmd)
--- a/ducktape/cluster/remoteaccount.py
+++ b/ducktape/cluster/remoteaccount.py
@@ -101,6 +101,7 @@ class RemoteAccount(HttpMixin):
         return r
 
     def scp_from(self, src, dest, recursive=False):
+        """Copy something from this node."""
         return self._ssh_quiet(self.scp_from_command(src, dest, recursive))
 
     def scp_to_command(self, src, dest, recursive=False):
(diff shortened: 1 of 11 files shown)


ducktape/services/service.py
```diff
@@ -52,21 +60,71 @@ def start(self):
...
+        for node in self.nodes:
+            self.logger.debug("Starting node %d on %s" % (self.idx(node), node.account.hostname))
+            self.start_node(node)
+            self.wait_until_alive(node)
+
+    def start_node(self, node):
+        """Start service process(es) on the given node."""
+        raise NotImplementedError("Subclasses must implement clean_node.")
+
+    def wait_until_alive(self, node):
```
ewencp: Can you explain why this should be separate from start_node? The only use of it I see in this or the muckrake patch is during start/restart, but since restart presumably uses stop/start except in very unusual cases, wouldn't this all be simpler without wait_until_alive?

That said, I think it could be useful for a subclass like DaemonService to separate out start_process and wait_until_alive since since start_process might always be a different command for subclasses but wait_until_alive might follow a common pattern (e.g. checking pid liveness, looking for something in the log, etc).

ducktape/services/service.py
```diff
@@ -52,21 +60,71 @@ def start(self):
...
+    def clean(self):
+        """Clean up persistent state on each node - e.g. logs, config files etc.
+        Subclasses must override clean_node.
+        """
+        for node in self.nodes:
+            self.clean_node(node)
+
+    def clean_node(self, node):
+        """Clean up persistent state on this node - e.g. service logs, configuration files etc."""
+        raise NotImplementedError("Subclasses must implement clean_node.")
```
ewencp: Question here is whether we enforce good practice by raising this as the default or make writing tests easier by allowing services that can handle old state without a problem (e.g. they just overwrite old logs) ignore writing this method. Probably not a big deal since they can always override and leave it empty, but something to think about.

ducktape/services/service_registry.py
```diff
@@ -0,0 +1,63 @@
...
+
+    def stop_all(self):
+        """Stop all currently registered services in the reverse of the order in which they were added.
+
+        Note that this does not clean up persistent state or free the nodes back to the cluster.
+        """
+        for service in reversed(self.values()):
+            try:
+                service.stop()
+            except Exception as e:
```
ewencp: You might want the exceptions you're catching here to be BaseException. There are a few that won't be caught by Exception. See https://docs.python.org/2/library/exceptions.html#exception-hierarchy

Actually, KeyboardInterrupt in particular might need careful thought and handling -- it's unlikely we want to kill the process immediately in most cases, but we'll probably need to coordinate a couple of things. When shutting down a test, we'd probably like for that shutdown to finish (so we need to handle the exception). But then the test runner ideally wouldn't finish running any tests that were left. Might just want to change these to BaseException now and file a ticket for graceful ctrl-c handling.

ducktape/tests/loader.py
```diff
@@ -29,6 +31,28 @@ class TestLoader(object):
     def __init__(self, session_context):
         self.logger = session_context.logger
 
+    def parse_discovery_symbol(self, discovery_symbol):
```
ewencp: Update unit tests for the new syntax? I think this should just be a couple of lines to add a new test method to the existing tests.

ducktape/tests/reporter.py
```diff
@@ -36,9 +36,11 @@ class SimpleReporter(TestReporter):
     def header_string(self):
         """Header lines of the report"""
         header_lines = [
+            "=========================================================================",
             "Test run with session_id " + self.results.session_context.session_id,
             self.pass_fail(self.results.get_aggregate_success()),
-            "---------------------------------------------------------------------"
+            "==========================="
```
ewencp: Ok, definitely a nit here, but I think I will go crazy if these lines demarcating the report sections are not the same length :)

ducktape/tests/test.py
```diff
@@ -32,17 +33,85 @@ def __init__(self, test_context):
...
+            self.copy_service_logs()
+            self.services.clean_all()
+
+    def free_nodes(self):
+        self.services.free_all()
+
+    def copy_service_logs(self):
+        """Copy logs from service nodes to the results directory."""
+        for service in self.services.values():
+            if not hasattr(service, 'logs'):
```
ewencp: Is this `logs` field documented anywhere? How as a Service writer do I know how to specify these?

ducktape/tests/test.py
```diff
@@ -32,17 +33,85 @@ def __init__(self, test_context):
...
+                self.test_context.logger.debug("Won't collect service logs from %s - no 'logs' attribute." %
+                    service.__class__.__name__)
+                return
+
+            log_dirs = service.logs
+            for node in service.nodes:
+                for log_name in log_dirs.keys():
```
ewencp: If you break this up into checking if the logs should be collected, then collecting them, then you should be able to scp them all at once, right? It looks like they're all landing in the same directory, so that should be safe and will be faster.

Eventually we'll also want to parallelize the collection from all the hosts, but for now just getting all at once is going to be a big improvement for even just 2 or 3 log files, especially if they are small.

ducktape/tests/test.py
```diff
@@ -32,17 +33,85 @@ def __init__(self, test_context):
...
+                    if self.should_collect_log(log_name, service, node):
+                        dest = os.path.join(
```
ewencp: This doesn't change while on the same node. You can compute the path and check if the directory exists only once assuming you make the change suggested right above here.

ducktape/tests/test.py
```diff
@@ -32,17 +33,85 @@ def __init__(self, test_context):
...
+                            self.test_context.logger.warn(
+                                "Error copying log %(log_name)s from %(source)s to %(dest)s. \
+                                service %(service)s: %(message)s" %
+                                {'log_name': log_name,
+                                 'source': log_dirs[log_name],
+                                 'dest': dest,
+                                 'service': service,
+                                 'message': e.message})
+
+    def mark_log_for_collect(self, log_name, service, node=None):
```
ewencp: It looks like these are intended to be used by Test subclasses? We should make sure anything users are expected to call on a regular basis has at least brief documentation.

ducktape/tests/test.py
```diff
@@ -32,17 +33,85 @@ def __init__(self, test_context):
...
+
+    def mark_log_for_no_collect(self, log_name, service, node):
+        if node is None:
+            # By default, mark all nodes for collection
+            for n in service.nodes:
+                self.log_collect[(log_name, service.__class__.__name__, n.account.hostname)] = False
+        else:
+            self.log_collect[(log_name, service.__class__.__name__, node.account.hostname)] = False
+
+    def should_collect_log(self, log_name, service, node):
```
ewencp: So if I'm interpreting this right, the model is that Services specify all logs that a test might be interested in keeping, and tests blacklist anything they don't want? If we're doing this for Kafka, do we always want opt-in? I'm wondering if defaulting to always collecting the data logs (which might be very large) and having to repeat the filtering logic in every test that doesn't need them is a good idea?

ducktape/services/service.py
granders: Hmm I guess it's true that some tests are just a thin wrapper around a service that runs some java test class, and this makes writing such services slightly more painful.

However given that users can really easily mock it with a 'pass', I don't think it makes it that inconvenient, and at least forces you to *decide* to ignore it.

ewencp: Sounds fine, just trying to keep an eye out for places where we can make things as simple as possible for test authors.

ducktape/services/service.py
granders: discussed offline - removed wait_until_alive

ducktape/services/service_registry.py
granders: Changed to BaseException

ducktape/tests/test.py
granders: The original idea was to provide finer-grained options. The use-case I had in mind was to have the ability to collect the Kafka 'service log' without collecting the kafka 'data log' (or dynamically choose whether to collect the data logs).

Maybe all or none is good enough, I'm not sure.

ewencp: You can still do that, I'm just trying to reduce the number of scp processes since they each have overhead. Right now you have

```
for each service:
  for each node:
    for each log specified by the service
      if should_collect(log):
        scp(node, log, log_dir)
```

This runs num_logs * num_nodes scp commands. I'm suggesting reorganizing this as follows:

```
for node:
  node_logs = []
  for each service:
    for each log specified by the service:
      if should_collect(log):
        node_logs += log
  scp(node, node_logs, log_dir)
```
You get exatly the same type of filtering, you just figure out everything you want to pull from each node and run a single scp command per node.

ducktape/tests/loader.py
granders: added

granders:
@ewencp 
Ok ready for another round.

ducktape/cluster/remoteaccount.py
```diff
@@ -90,6 +89,14 @@ def kill_process(self, process_grep_str, clean_shutdown=True, allow_fail=False):
             self.ssh(cmd, allow_fail)
 
     def scp_from_command(self, src, dest, recursive=False):
+        if type(src) == list:
```
ewencp: It might be better to switch this test around: instead of checking if it's a list, check if it's not `isinstance(src, basestring)` and then treat it as an iterable. This should work better with other types you could easily encounter. Ideally it would work with iterators too, which don't support `len()`. I think the `len(src) == 1` case should just work using the same code as the `else`, it'll just include the extra `{}`. Not sure about the empty one, but I'm not sure that should even be allowed since it probably produces a command that generates an error anyway since an empty src will result in too few arguments.

ducktape/cluster/remoteaccount.py
```diff
@@ -90,6 +89,14 @@ def kill_process(self, process_grep_str, clean_shutdown=True, allow_fail=False):
...
+            if len(src) == 0:
+                src = ""
+            elif len(src) == 1:
+                src = src[0]
+            else:
+                src = "\{" + ",".join(src) + "\}"
```
ewencp: Cool! I was unaware of this in bash!

One potential drawback is for people who aren't using bash as their shell. Not sure if we want to worry about that now, but it may become a problem.

ducktape/services/service.py
```diff
@@ -12,36 +12,63 @@
...
+    and check logs/output from the service, which can be helpful in writing tests
+    or benchmarks.
+
+    Services should generally be written to support an arbitrary number of nodes,
+    even if instances are independent of each other. They should be able to assume
+    that there won't be resource conflicts: the cluster tests are being run on
+    should be large enough to use one instance per service instance.
+    """
+
+    def __init__(self, context, num_nodes):
```
ewencp: Might want to add a small docstring that at least explains the `context` argument. I think the explanation just below here is fine, and maybe just add that in most cases, the context will be your test class instance (i.e. self in your Test).

We haven't really talked about how to handle docs for ducktape yet, but writing a new Service is probably the first thing any user needs to do, so they'll hit this pretty quickly.

ducktape/services/service.py
```diff
@@ -12,36 +12,63 @@
...
+
+        # Provides a mechanism for locating and collecting log files produced by the service on its nodes.
+        # logs is a dict with entries that look like log_name: {"path": log_path, "collect_default": boolean}
+        #
+        # For example, zookeeper service might have self.logs like this:
+        # self.logs = {
+        #    "zk_log": {"path": "/mnt/zk.log",
+        #               "collect_default": True}
+        # }
+        self.logs = {}
```
ewencp: You might not want to always set this here. In a lot of cases, there will be a fixed set of logs for a given service, in which case it might make more sense to set it at the class level, e.g. something like

```
class Zookeeper(Service):
    logs = { "zk_log": {"path": "/mnt/zk.log", "collect_default": True}
```

You can make this work well while still providing a default here (rather than having to always check `hasattr(self, 'logs')`) using a pattern like this example:

```
>>> class Base(object):
...     def __init__(self):
...             if not hasattr(self, 'logs'):
...                     self.logs = ''
...
>>> class Foo(Base):
...     logs = 'foo.log'
...
>>> Base().logs
''
>>> Foo().logs
'foo.log'
```

ducktape/services/service.py
```diff
@@ -12,36 +12,63 @@
...
+
+    def allocate_nodes(self):
+        """Request resources from the cluster."""
+        if self.allocated:
+            self.logger.warn("Requesting nodes for a service that has already been allocated nodes.")
+            self.logger.warn("Returning without requesting new nodes.")
```
ewencp: Any reason this shouldn't throw an exception?

ducktape/services/service.py
```diff
@@ -52,21 +79,71 @@ def start(self):
...
+        """Start the service on all nodes."""
+        self.logger.info("Starting service: %s" % self.__class__.__name__)
+
+        if not self.allocated:
+            self.allocate_nodes()
+
+        self.logger.debug("Killing processes and attempting to clean up before starting nodes.")
+        for node in self.nodes:
+            # Added precaution - kill running processes
+            self._kill_running_processes(node)
```
ewencp: This was the call I was talking about getting rid of here, or moving to a utility. Its implementation just kills all java processes.

This is especially important if we make the LocalhostCluster work :)

ducktape/services/service.py
granders: I put it at the class level
Putting `log = {}` it in the base Service class also renders `hasattr` unnecessary

ducktape/cluster/remoteaccount.py
granders: Ah I didn't realize this was bash-specific.

ducktape/cluster/remoteaccount.py
granders: Done. A little more robust and concise this way.

ducktape/services/service.py
granders: done

ducktape/services/service.py
granders: Hmm, I see no real problem with throwing an exception.

granders:
@ewencp Ready for another round

ewencp:
LGTM.

PR #47, opened 2015-05-04
Vagrant cluster ip

granders:
A few updates:

- Vagrant cluster checks for externally_routable_ip to allow us to get away without modifying /etc/hosts on driver machine

- Now handles keyboard interrupt better - still cleans up, but then runs no more tests.

- Cleaned up reporting a little bit, added run time to the output

Also, miscellaneous small refactoring here and there.


ewencp:
@granders I'm getting a lot of test errors with the current version.

ducktape/command_line/main.py
```diff
@@ -36,10 +35,12 @@ def parse_args():
                         help='one or more space-delimited strings indicating where to search for tests')
     parser.add_argument("--collect-only", action="store_true", help="display collected tests, but do not run")
     parser.add_argument("--debug", action="store_true", help="pipe more verbose test output to stdout")
+    parser.add_argument("--exitfirst", action="store_true", help="exit after first failure")
```
ewencp: `--exit-first` instead? `--collect-only` has a separating dash

ducktape/command_line/main.py
```diff
@@ -108,8 +107,12 @@ def main():
         print test_classes
         sys.exit(0)
 
+    # Initializing the cluster is slow, so do so only if
```
ewencp: Oh, this makes so much sense! I wondered why test discovery would always take so long.

ducktape/tests/test.py
```diff
@@ -35,27 +35,57 @@ def __init__(self, test_context, *args, **kwargs):
...
         """
         if hasattr(self.test_context, 'services') and type(self.test_context.services) == ServiceRegistry:
-            self.test_context.services.stop_all()
-            self.copy_service_logs()
-            self.test_context.services.clean_all()
+            exceptions = []
+            try:
+                self.test_context.services.stop_all()
+            except BaseException as e:
+                exceptions.append(e)
```
ewencp: Isn't it probably bad if we're seeing exceptions here? Right now they're just being swallowed. It seems like we should at least log them?

granders:
@ewencp 
Ready for another pass.
teardown and handling of KeyboardInterrupt is a little more simpler now. Fixed unit tests as well

ewencp:
LGTM

PR #50, opened 2015-05-22
Support multiple tests per Test class.

### LINKED ISSUE #49: multiple test methods per class
test writers should be able to write test classes that have multiple test methods

With this in place, the examples we present to the community can be both cleaner and look more familiar to those with junit background.

ewencp:
Test methods must be name test.* or .*test. The test loader now supports
matching up to the test method in the format
file.py::TestClass.test_method. Test contexts are now preferred when logging
some information since only having the test class name is no longer
sufficient. Fixes #49.

### DIFF
--- a/README.md
+++ b/README.md
@@ -26,10 +26,11 @@ Use
 ---
 ducktape discovers and runs tests in every path provided. 
 
-    ducktape <relative_path_to_testdirectory>        # e.g. ducktape dir/tests
-    ducktape <relative_path_to_file>                 # e.g. ducktape dir/tests/my_test.py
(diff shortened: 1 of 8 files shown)


ewencp:
See also confluentinc/muckrake#31. @granders this was pretty straightforward, but we do now have to use TestContext in many places rather than just referring to the Test for naming info because just having the Test class name isn't sufficient.

Also, passing the test (and now test context) around a lot in SerialTestRunner seemed a bit wasteful since that class is guaranteed to only have one active test at a time, so I simplified that by keeping that state in a field on the test runner.

granders:
@ewencp  Thanks, LGTM

PR #52, opened 2015-05-30
Add option to control cluster type.

### LINKED ISSUE #33: Add command-line hook to specify type of cluster
Right now command_line.main fires up a VagrantCluster, but it should be possible to choose.

ewencp:
To allow setting it on a per-project or per-user basis, this also adds support
for loading configs from config files in addition to the command line. Fixes #33.

### DIFF
--- a/ducktape/command_line/config.py
+++ b/ducktape/command_line/config.py
@@ -16,11 +16,14 @@ import os
 
 
 class ConsoleConfig(object):
+    # Directory for project-specific ducktape configs and runtime data
+    DUCKTAPE_DIR = ".ducktape"
+
     # Store various bookkeeping data here
-    METADATA_DIR = ".ducktape"
+    METADATA_DIR = os.path.join(DUCKTAPE_DIR, "metadata")
 
     # Default path, relative to current project directory, to the project's ducktape config file
-    PROJECT_CONFIG_FILE = os.path.join(METADATA_DIR, "config")
+    PROJECT_CONFIG_FILE = os.path.join(DUCKTAPE_DIR, "config")
 
     # Default path to the user-specific config file
     USER_CONFIG_FILE = '~/.ducktape/config'
--- a/ducktape/command_line/config.py
+++ b/ducktape/command_line/config.py
@@ -26,7 +26,7 @@ class ConsoleConfig(object):
     PROJECT_CONFIG_FILE = os.path.join(DUCKTAPE_DIR, "config")
 
     # Default path to the user-specific config file
-    USER_CONFIG_FILE = '~/.ducktape/config'
+    USER_CONFIG_FILE = os.path.join('~', DUCKTAPE_DIR, 'config')
 
     # Default cluster implementation
     CLUSTER_TYPE = "ducktape.cluster.vagrant.VagrantCluster"
--- a/ducktape/command_line/config.py
+++ b/ducktape/command_line/config.py
@@ -19,6 +19,15 @@ class ConsoleConfig(object):
     # Store various bookkeeping data here
     METADATA_DIR = ".ducktape"
 
+    # Default path, relative to current project directory, to the project's ducktape config file
+    PROJECT_CONFIG_FILE = os.path.join(METADATA_DIR, "config")
+
+    # Default path to the user-specific config file
+    USER_CONFIG_FILE = '~/.ducktape/config'
+
+    # Default cluster implementation
+    CLUSTER_TYPE = "ducktape.cluster.vagrant.VagrantCluster"
+
     # Track the last-used session_id here
     SESSION_ID_FILE = os.path.join(METADATA_DIR, "session_id")
 
--- a/ducktape/command_line/main.py
+++ b/ducktape/command_line/main.py
@@ -113,6 +113,8 @@ def main():
 
     setup_results_directory(results_dir, session_id)
     session_context = SessionContext(session_id, results_dir, cluster=None, args=args)
+    for k,v in vars(args).iteritems():
+        session_context.logger.debug("Configuration: %s=%s", k, v)
 
     # Discover and load tests to be run
     extend_import_paths(args.test_path)
--- a/ducktape/command_line/main.py
+++ b/ducktape/command_line/main.py
@@ -39,8 +39,10 @@ def parse_args():
                         help='one or more space-delimited strings indicating where to search for tests')
     parser.add_argument("--collect-only", action="store_true", help="display collected tests, but do not run")
     parser.add_argument("--debug", action="store_true", help="pipe more verbose test output to stdout")
-    parser.add_argument("--cluster", action="store", default="ducktape.cluster.vagrant.VagrantCluster",
-                        help="Cluster class to use to allocate nodes for tests")
+    parser.add_argument("--config-file", action="store", default=ConsoleConfig.PROJECT_CONFIG_FILE,
+                        help="path to project-specific configuration file")
+    parser.add_argument("--cluster", action="store", default=ConsoleConfig.CLUSTER_TYPE,
+                        help="cluster class to use to allocate nodes for tests")
     parser.add_argument("--exit-first", action="store_true", help="exit after first failure")
 
     project_configs = []
@@ -49,7 +51,7 @@ def parse_args():
         project_configs = list(itertools.chain(*[line.split() for line in open(project_config_file).readlines()]))
 
     user_configs = []
-    user_config_file = os.path.expanduser('~/.ducktape/config')
+    user_config_file = os.path.expanduser(ConsoleConfig.USER_CONFIG_FILE)
     if os.path.exists(user_config_file):
         user_configs = list(itertools.chain(*[line.split() for line in open(user_config_file).readlines()]))
 
--- a/ducktape/command_line/main.py
+++ b/ducktape/command_line/main.py
@@ -30,7 +30,7 @@ def parse_args():
     """Parse in command-line and config file options.
 
     Command line arguments have the highest priority, then user configs specified in ~/.ducktape/config, and finally
-    project configs specified in <ducktape_metadata_dir>/config.
+    project configs specified in <ducktape_dir>/config.
 
     :rtype: argparse.Namespace
     """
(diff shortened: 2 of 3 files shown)


ewencp:
@granders Addition of the option to control cluster type is pretty trivial, but since the default choice might vary by org/project, I also added support for loading configs from the project directory and from a file in the user's home directory. I think my main question for review here is whether putting it under <project>/.ducktape/config is kosher -- that's the METADATA_DIR, but I'm not sure if that's really a good location (especially if it is expected to be .gitignored usually).

granders:
I'm leaning away from METADATA_DIR and instead toward something like "ducktape.cfg" in cwd.  Here is my mini thought process

pros 
- convenient
- reasonably intuitive

cons
- so what is metadata directory for? Tracking metadata about test runs. Right now this is just the last used session id, but it could include other things (?) like run statistics (?). If this is indeed its purpose, then having configs here is odd.
- awkward given .gitignore issue
- might want to specify different metadata directory in config file(?)... but the config file lives in the metadata directory?

alternative
- we freely read and write files in the cwd, so having "ducktape.cfg" in cwd seems reasonable, and easy to use
- this does add one more file (we'd have results/ , ducktape.cfg, and .ducktape) but I don't see a problem with this

ducktape/command_line/main.py
```diff
@@ -24,10 +23,15 @@
...
 import sys
-
+import importlib
+import itertools
 
 def parse_args():
-    """Parse in command-line options.
+    """Parse in command-line and config file options.
+
+    Command line arguments have the highest priority, then user configs specified in ~/.ducktape/config, and finally
```
granders: Hm, is adding extra state outside of the project that useful? I guess if you're using it with several different projects it's nice to have ducktape work the way you want out-of-the-box is nice.

ducktape/command_line/main.py
```diff
@@ -35,9 +39,23 @@ def parse_args():
                         help='one or more space-delimited strings indicating where to search for tests')
     parser.add_argument("--collect-only", action="store_true", help="display collected tests, but do not run")
     parser.add_argument("--debug", action="store_true", help="pipe more verbose test output to stdout")
+    parser.add_argument("--cluster", action="store", default="ducktape.cluster.vagrant.VagrantCluster",
+                        help="Cluster class to use to allocate nodes for tests")
```
granders: Very minor - there's an arbitrary convention of lower case initial letter for help descriptions. 

Either way is fine by me as long as they're consistent.

ducktape/command_line/main.py
```diff
@@ -35,9 +39,23 @@ def parse_args():
...
     parser.add_argument("--exit-first", action="store_true", help="exit after first failure")
 
-    args = parser.parse_args()
+    project_configs = []
+    project_config_file = os.path.join(ConsoleConfig.METADATA_DIR, 'config')
+    if os.path.exists(project_config_file):
+        project_configs = list(itertools.chain(*[line.split() for line in open(project_config_file).readlines()]))
+
+    user_configs = []
+    user_config_file = os.path.expanduser('~/.ducktape/config')
```
granders: Whatever convention is chosen for location and name of config, the path/name should probably live in ConsoleConfig in config.py

ducktape/command_line/main.py
```diff
@@ -109,7 +127,14 @@ def main():
 
     # Initializing the cluster is slow, so do so only if
     # tests are sure to be run
-    session_context.cluster = VagrantCluster()
+    try:
+        (cluster_mod_name, cluster_class_name) = args.cluster.rsplit('.', 1)
+        cluster_mod = importlib.import_module(cluster_mod_name)
+        cluster_class = getattr(cluster_mod, cluster_class_name)
+        session_context.cluster = cluster_class()
```
granders: do we want the cluster to be a subclass of Cluster? 

I guess the test runner will still (probably) fail if it doesn't walk and talk like a Cluster, but fail-fast and all that jazz.

ducktape/command_line/main.py
```diff
@@ -35,9 +39,23 @@ def parse_args():
                         help='one or more space-delimited strings indicating where to search for tests')
     parser.add_argument("--collect-only", action="store_true", help="display collected tests, but do not run")
     parser.add_argument("--debug", action="store_true", help="pipe more verbose test output to stdout")
+    parser.add_argument("--cluster", action="store", default="ducktape.cluster.vagrant.VagrantCluster",
```
granders: This could be a future issue, but it could be useful to specify config file with a flag.

granders:
I suppose whether it's kosher is really up to us. My concept of metadata dir is not terribly clear other than the rough idea I mentioned in the comment.

As for having it also live in the home directory - although this kind of thing isn't unusual, it seems more relevant if your program is getting used all over the place (like an editor)

ducktape/command_line/main.py
ewencp: This is the basic idea. Think about someone working on Kafka + Samza. It sucks to have to duplicate configs (like what cluster type to use) across multiple projects. (Although this also makes one of the drawbacks of the way we currently configure the Vagrant clusters pretty clear too -- it's not possible to share them between projects.)

ducktape/command_line/main.py
ewencp: Fixed

ducktape/command_line/main.py
ewencp: Fixed

ducktape/command_line/main.py
ewencp: Fixed

ducktape/command_line/main.py
ewencp: Despite the recent type checking PEP, explicitly checking the type is discouraged, and shouldn't really be necessary here. They just need to have the right set of methods for us to invoke.

ewencp:
I reused METADATA_DIR not because of it's naming, but because you appropriated the entire .ducktape/ directory for metadata alone :) That variable name certainly isn't set in stone.

I tend to want to use dot files because they hide away stuff that (unfortunately) usually needs to be at the root level of the repository. Maybe it's less of an issue for projects that use ducktape, but there's a tendency to have a mess of root-level files now because everyone uses them for configs - Vagrantfile, travis.yml, language-specific packaging info (e.g. setup.py), etc. It really clutters the root directory.

Using a dot file (even if it ended up in a different location) is also an indicator that you generally don't want to be making changes to the file -- it's hidden for a reason.

Anyway, I'm open to alternatives, but I'd also argue that we should *stop* writing different things to the root directory. If we have a variety of stuff we need to write, we should a) collect it all together in one location and b) make it easy to setup a good .gitignore for it.

ewencp:
@granders Updated patch to share .ducktape. On the user-specific config, I plan on ducktape taking over the world, so I'm just planning ahead for its pervasive use :)

But seriously, the idea here was for people who want to run on, e.g., their own standalone cluster regularly across Kafka, Samza, muckrake tests, etc., this would be an easy way to force the cluster to always use a custom type. This makes selecting a default like Vagrant a lot less contentious since there's an easy, global out.

ducktape/command_line/main.py
```diff
@@ -35,9 +39,25 @@ def parse_args():
...
+        project_configs = list(itertools.chain(*[line.split() for line in open(project_config_file).readlines()]))
+
+    user_configs = []
+    user_config_file = os.path.expanduser(ConsoleConfig.USER_CONFIG_FILE)
+    if os.path.exists(user_config_file):
+        user_configs = list(itertools.chain(*[line.split() for line in open(user_config_file).readlines()]))
+
+    # The order of concatenation ensures overrides are handled properly -- for action=store arguments, the later options
+    # will override earlier instances
+    args = parser.parse_args(project_configs + user_configs + sys.argv[1:])
```
granders: It looks like maybe this should be (flip the order of user and project configs):
user_configs + project_configs + sys.argv[1:]

ducktape/command_line/config.py
```diff
@@ -16,8 +16,20 @@
...
+
     # Store various bookkeeping data here
-    METADATA_DIR = ".ducktape"
+    METADATA_DIR = os.path.join(DUCKTAPE_DIR, "metadata")
+
+    # Default path, relative to current project directory, to the project's ducktape config file
+    PROJECT_CONFIG_FILE = os.path.join(DUCKTAPE_DIR, "config")
+
+    # Default path to the user-specific config file
+    USER_CONFIG_FILE = '~/.ducktape/config'
```
granders: minor, but you could stick with os.path.join() here?

granders:
@ewencp 
LGTM with the two tweaks I mentioned.

ducktape/command_line/main.py
ewencp: I considered that, but the motivating use case was the global config for cluster type because you want to use the same type across projects. For that to work you need to be able to override project-level defaults with the user config, i.e. if we set a default via a config file in Kafka and in Samza, I'd want to override that for all projects simultaneously. I think a lot of projects struggle with this ordering issue, and then some end up adding project-specific sections to the user-level config (e.g. delimited by `[project]` headers) or end up with something like the Vagrantfile vs. Vagrantfile.local configs I setup for Kafka.

I'm open to suggestions for how to support both use cases well.

granders: Oops, I read this as disagreeing with the comment but I was mistaken. 

This is definitely unintuitive to me however. I guess my mental model is greater specificity == higher priority. I would expect to set up my user config for common cross-project configs and override configs as needed at the project level. 

For example, if user config overrides project config and I want some config to be *different* between projects I have to: a) check whether there is a user config file b) delete relevant config there c) add that deleted config back to all the other projects d) update the one config on the one project that needs to be different.

Maybe it's moot in that it boils down to preference as to which types of config changes you prefer being less convenient. 

That said, however overrides work, the resulting "composite config" should be immediately visible (maybe print this in main?). This helps avoid cases where you think you're running with some configuration but it turns out to be different due to misunderstanding of override behavior, an unexpected config file in an unexpected place, or say a forgotten --no-teardown that was intended to be deleted but wasn't.

ewencp: Yeah, I take your point. The reason I tend to view user configs more specific than project configs is because project configs end up applying to all users of the project (because the file is shared in the repo). That was why I was pointing out Vagrantfile vs. Vagrantfile.local -- that was the best solution I could come up with to standardize truly project-wide settings (that apply to all users) but provide the override in a convenient location on a per-checkout basis.

granders: Ok I understand now why you are treating user config as more specific.

What do you think of printing the composite config when ducktape is run?

ewencp: How about with --debug? I'm skeptical of adding that output to every run since it adds noise.

granders: Sounds good - Logging it with the session logger at debug level will do the trick.

PR #60, opened 2015-06-11
Move muckrake

### LINKED ISSUE #48: mechanism for data collection
current tests just pipe collected performance (etc) data to logs.

ducktape should pipe it (json is probably fine) to its own file so it's easily discoverable when regression checks are added

### LINKED ISSUE #57: Copying logs causes exception if a service was never started
For example, due to a failure of a different service during a test, this test didn't start the ConsoleConsumer service. Therefore it never allocated any nodes and triggered this exception:

```
Traceback (most recent call last):
  File "/Users/ewencp/confluent/ducktape.git/ducktape/tests/runner.py", line 143, in teardown_single_test
    self.current_test.copy_service_logs()
  File "/Users/ewencp/confluent/ducktape.git/ducktape/tests/test.py", line 70, in copy_service_logs
    for node in service.nodes:
AttributeError: 'ConsoleConsumer' object has no attribute 'nodes'
```

granders:
Parallel branch to move_muckrake in the kafka fork.

Notable changes
- Moved background_thread.py to ducktape
- Added SingleTestReporter
- Added per-test data output file (fixes #48)
- Fixed #57 by initializing self.nodes = [] in Service


ducktape/tests/reporter.py
```diff
@@ -18,45 +18,109 @@
 import pkg_resources
 
 
-class TestReporter(object):
+SEPARATOR_LENGTH = 100
```
ewencp: You might just want something like [this](https://gist.github.com/jtriley/1108174) instead of using a fixed width. Although maybe that only makes sense for the Stdout versions.

ducktape/tests/reporter.py
```diff
@@ -18,45 +18,109 @@
...
+
+        return "\n".join(header_lines)
+
+    def result_string(self):
+        """Stringify a single result."""
+        result = self.result
+        result_lines = []
+
+        if result.data is not None:
+            result_lines.append(json.dumps(result.data))
```
ewencp: We should probably be catching failures to serialize and doing something useful with them (unless they get caught ok here and the normal exception cleanup code handles this case?). Would be nice to warn the user why they're getting this in case they include something non-JSON-serializable in their output.

ducktape/tests/reporter.py
```diff
@@ -18,45 +18,109 @@
...
+
+class SingleTestStdoutReporter(SingleTestReporter):
+    def report(self):
+        print self.report_string()
+
+
+class SingleTestFileReporter(SingleTestReporter):
+
+    def report(self):
+        report_file = os.path.join(self.result.test_context.results_dir, "report.txt")
```
ewencp: I think report.txt and data.json are new, should they be documented in the test output directory structure in the README?

ewencp:
@granders These were all minor comments, feel free to merge, deferring anything to issues if that makes sense since the current version functions as expected.

granders:
@ewencp Ready for another pass

Main differences: 
- Updated aws section in readme
- --version flag

ducktape/command_line/config.py
```diff
@@ -43,3 +43,7 @@ class ConsoleConfig(object):
     # Log this to indicate a test is misbehaving to help end user find which test is at fault
     BAD_TEST_MESSAGE = "BAD_TEST"
 
+    # ducktape version
+    DUCKTAPE_VERSION = "0.2.0"
```
ewencp: Kind of weird for this to be under the command line part since it's just the version of the entire package.

The normal place to put this is in the packages __init__.py. See, e.g., https://github.com/kennethreitz/requests/blob/master/requests/__init__.py#L44 for an example (which also includes other metadata). They also go through pains to avoid duplicating it so there's no chance of things getting out of sync: https://github.com/kennethreitz/requests/blob/master/setup.py#L31 Might want to follow a similar approach.

ducktape/services/service.py
```diff
@@ -100,11 +102,29 @@ def start(self):
...
+            except:
+                pass
+
+            try:
+                self.force_clean_node(node)
+            except:
+                pass
+
+            try:
+                self.stop_node(node)
```
ewencp: Hmm, I haven't looked at this code for awhile, but how does this sequence of calls make sense? It doesn't seem like stop_node makes sense if we have something called _kill_running_processes. And force_clean_node seems to delete stuff out of /mnt, but happens before stop_node, but then couldn't something in stop_node log to a file in /mnt?

A couple of other problems:

1. Can we just drop _kill_running_processes and delete that method? It's just killing all java processes, which isn't a good solution in ducktape? If we want to provide something like that for java processes it should be in kafka/muckrake. Plus, it's not even really safe if there are any other java processes that might be legitimately running -- it was just an ugly hack when I could assume this was all in a VM.

2. force_clean_node's implementation is assuming the layout that we're using for kafka/muckrake. We can't really provide that as a default implementation in ducktape. Again, we could move this out to some common Service base class/mixin in kafka/muckrake. Alternatively, both force_clean_node (which handles files) and _kill_running_processes seem like things you might want to specify at the cluster level, so maybe there's a better hook to be provided for them in ducktape. Or yet another alternative would be removing them entirely -- they actually should only be necessary during development since robust Service implementations won't leave any trace if they clean up properly.

ducktape/tests/reporter.py
```diff
@@ -12,98 +12,131 @@
...
+    def result_string(self):
+        """Stringify single result"""
+        result_lines = [
+            "test_id:    %s" % self.result.test_context.test_id,
+            "status:     %s" % pass_fail(self.result.success),
+            "run time:   %s" % format_time(self.result.run_time),
+        ]
+
+        if self.result.data is not None:
+            result_lines.append(json.dumps(self.result.data))
```
ewencp: Are we at all worried about making this part of the standard output? THe few we have now are nice short summaries. What if we return much more data (e.g. per-second throughput numbers for a long running throughput test) to do performance regression tests?

ducktape/tests/reporter.py
```diff
@@ -12,98 +12,131 @@
...
 
-    def result_string(self, result):
-        """Stringify a single result."""
-
-        result_lines = [
-            self.pass_fail(result.success) + ":     " + result.test_name,
-            "run time: %s" % self.format_time(result.run_time)
-            ]
-
-        if not result.success:
```
ewencp: As far as I can tell, there's no longer an equivalent to this code which would log the summary info when there's an error. So I think we now have no logs containing the actual error.

setup.py
```diff
@@ -2,6 +2,7 @@
 from setuptools.command.test import test as TestCommand
 import sys
 
+from ducktape.utils.util import ducktape_version
```
ewencp: Hmm, are you sure this actually works properly? Do you know what the rules are for how PYTHONPATH is setup when you use setup.py directly, via pip, etc? I'm not sure, so I'm not sure we're guaranteed it will pick up `ducktape.utils.util` from the directory relative to this file or if it's possible it can grab, for example, an installed version instead. (See previous note about how requests handles this -- not sure that approach is ideal either, but they probably know how to handle this better than I do.)

ducktape/tests/reporter.py
granders: Yes I was just noticing I no longer got the failure explanation...
Fixed

ducktape/tests/reporter.py
granders: At this stage of the game without any automatic checks on the content of this data, I think it's better for the data collected to be immediately visible without extra effort.

Otherwise it is easy for "false positives" to slide by undiscovered as with tests using ConsumerPerformanceService

ducktape/services/service.py
granders: For now I'd prefer to remove these completely since they don't make a lot of sense in the more general setting, and could be somewhat dangerous (reckless removal of files, reckless destruction of java processes) if used in the wrong context. 

Miscellaneous thoughts regarding cleanup:
Default "nuclear cleanup" behavior should probably take into account some state associated with cluster type. 

For file cleanup, maybe the cluster specifies a root directory where any logs/persistent state goes by convention. 

For process cleanup, we need a more uniform approach to process handling. Maybe services register one or more pids when they start (e.g. a mapping { key => pid } where key is something like node + extra info (extra info is needed for services such as Hadoop that start multiple processes per node)). Where these pids are persisted would depend on the cluster type.

setup.py
granders: No I'm not sure. Switched the the mechanism used in requests where the setup.py file actually greps for the version

granders:
@ewencp 
- Moved version to __init__.py
- Removed nuclear cleanup/kill methods from Service.py

ducktape/services/service.py
ewencp: On "nuclear cleanup", agreed, you need some info about the cluster type. I'm a bit worried about relying on that option though -- anything sufficiently complex is going to turn into a mess. For example, in system tests for clients which use a wide variety of languages, is there really going to be a reasonable nuclear option? Or if there is, is it going to require a ton of steps to cover all the possible process types?

For file cleanup, I like that idea though we need to figure out where global-ish configs like that should come from.

For process cleanup, agreed a more uniform approach to processes would help. One option would be for the pid files to be stored in a standard location so that its easy to just scan through and kill -9 them if they still exist. Of course that's another global-ish config that needs to be specified on a per-project basis.

ewencp:
LGTM. There's one more issue with the removal of `allow_fail` on `stop_node`, but that's an issue in the corresponding Kafka patch.

PR #65, opened 2015-06-29
Parametrize

granders:
@ewencp 
Ready for initial review.

The marking logic isn't thread safe. At this point it doesn't need to be since it is exclusively used during the loading process.

A couple observations
- Current approach to marking breaks if a non-ducktape decorator is above a ducktape decorator.
- I don't know the best way to signal which methods in decorate.py should be public.
- I'm still not convinced that `@matrix` above `@parametrize` should "multiply"
- I updated the loader to construct each entire TestContext object. This lets me display --collect-only results in a nicer way, and simplified setting up the right directory structure for results.


granders:
I pushed a muckrake branch (won't merge that branch) to allow experimentation:

    $ cd muckrake
    $ git fetch && git checkout parametrize
    $ ducktape muckrake/tests/test_parametrized.py

ducktape/decorate.py
```diff
@@ -0,0 +1,229 @@
+# Copyright 2015 Confluent Inc.
```
ewencp: The module name sounds like it'll just end up being a dumping ground for any sort of decorator. Rename to parametrize?

ducktape/decorate.py
```diff
@@ -0,0 +1,229 @@
...
+
+
+import functools
+
+PARAMETRIZED = "PARAMETRIZED"
+INJECTED = "INJECTED"
+MATRIX = "MATRIX"
+MARKS = {PARAMETRIZED, INJECTED, MATRIX}
+
+_marked_functions = {}
```
ewencp: You could also use the `__dict__` field on the unbound methods to store the information about what's been marked, which would avoid this global data.

ducktape/decorate.py
```diff
@@ -0,0 +1,229 @@
...
+        indices[j] = i % length_of_j
+        i /= length_of_j
+        if i == 0:
+            break
+
+    return indices
+
+
+def matrix(**kwargs):
+    """Function decorator used to parametrize its with a matrix of values.
```
ewencp: typo? "parametrize its with"

ducktape/decorate.py
```diff
@@ -0,0 +1,229 @@
...
+        if _expandable(f):
+            kwargs_list.extend(f.kwargs_list)
+
+        wrapped = TestGenerator(f, kwargs_list)
+        mark_as(wrapped, PARAMETRIZED)
+        return wrapped
+    return parametrizer
+
+
+def _inject(**kwargs):
```
ewencp: Pretty sure this is just `functools.partial` (which is actually a bit more general since it handles positional arguments as well, although that's probably not important here).

ducktape/tests/loader.py
```diff
@@ -12,7 +12,8 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-from ducktape.tests.test import Test
+from ducktape.tests.test import Test, TestContext
+from ducktape.decorate import _expandable
```
ewencp: Why the underscore if this is used by this module?

ducktape/tests/loader.py
```diff
@@ -179,17 +168,84 @@ def import_modules(self, file_list):
...
+        for f_name in dir(t_info.cls):
+            f = getattr(t_info.cls, f_name)
+            if self.is_test_function(f):
+                test_methods.append(f)
+
+        test_info_list = []
+        for f in test_methods:
+            t = TestInfo(module=t_info.module, cls=t_info.cls, function=f)
+
+            if _expandable(f):
```
ewencp: Can we just check whether it's an iterable rather than having this `_expandable` method? e.g., couldn't we just try to use it in `extend()` and if it fails append it? (Or rather, do this with `f`, only constructing TestInfos a bit later). This seems like a more general way to handle this and would work with any decorator rather than requiring all of them to use the "marking" approach that relies on global state in that one module.

ducktape/tests/test.py
```diff
@@ -17,6 +17,7 @@
 from ducktape.command_line.config import ConsoleConfig
 from ducktape.services.service_registry import ServiceRegistry
 from ducktape.template import TemplateRenderer
+from ducktape.decorate import injected
```
ewencp: Unused import?

ducktape/tests/test.py
```diff
@@ -143,14 +160,18 @@ def results_dir(self):
             d = os.path.join(d, self.cls.__name__)
         if self.function is not None:
             d = os.path.join(d, self.function.__name__)
+
+        if self.injected_args is not None:
+            # The function has arguments injected; append these to the id
+            parameters = ".".join(["%s=%s" % (k, self.injected_args[k]) for k in self.injected_args])
```
ewencp: Not sure there's a good solution for this, but this may need some filtering if it's going to be used as part of a path. At a bare minimum, `os.sep` needs filtering/escaping or it'll create additional subdirectories. There are probably other characters that'll be odd to include.

Not sure if this is a realistic problem thought -- in practice I assume most will be simple number or string settings. On the other hand, settings like zkconnect with a chroot (which you might use for testing use of chroot/multitenancy) would include path separators...

ducktape/tests/test.py
```diff
@@ -164,6 +185,10 @@ def test_name(self):
                            self.cls.__name__ if self.cls is not None else None,
                            self.function.__name__ if self.function is not None else None]
 
+        if self.injected_args is not None:
+            # The function has arguments injected; append these to the id
+            name_components.extend(["%s=%s" % (k, self.injected_args[k]) for k in self.injected_args])
```
ewencp: This appears twice, just turn it into a method/property?

ducktape/decorate.py
```diff
@@ -0,0 +1,229 @@
...
+        else:
+            all_kwargs_list = kwargs_list
+
+        wrapped = TestGenerator(f, all_kwargs_list)
+        mark_as(wrapped, MATRIX)
+        return wrapped
+    return parametrizer
+
+
+def parametrize(**kwargs):
```
ewencp: Apparently I like the British spelling, parameterize. For my personal sanity, I recommend adding an alias, `parameterize = parametrize`.

ewencp:
On private functions, use one underscore as a prefix and a `from ducktape.decorate import *` won't import it. There isn't strict enforcement (you can still get at it via `import ducktape.decorate`), but the prefix indicates your intention. Classes have a similar approach -- one underscore is protected, two is private and triggers some name mangling to help discourage incorrect use of the private field.

I left a bunch of notes, but the implementation looks good to me.

ducktape/tests/loader.py
granders: Renamed _expandable to parametrized (parametrized returns True if @matrix or @parametrize is present)

I envisioned the "marking approach" as more widely applicable - potentially could be used with decorators like @ignore, @timeout(timeout=blah), etc. 

The trouble with iterable is it turns out to be *too* general - we end up with things that aren't test functions if we expand every iterable attribute.

ducktape/tests/test.py
granders: removed

ducktape/decorate.py
granders: renamed, moved to package "mark"

ducktape/decorate.py
granders: done

ducktape/decorate.py
granders: done

ducktape/decorate.py
granders: Changed _inject to use functools.partial - though I kept _inject so I could attach original_func.__name__ etc to the wrapper function using functools.wrap

(if you just use functools.partial you lose this metadata)

ducktape/tests/loader.py
granders: renamed to `parametrized`

ducktape/tests/loader.py
ewencp: Ah, I see. Just to be clear, the issue with other non-method fields, like if I have something like this?

```
class MyTest(Test):
    field1 = [1,2,3]
    field2 = "foobar"
```

ducktape/decorate.py
ewencp: I guess we're following py.test's lead here :) Not really sure "mark" addresses my initial concern, which was that every decorator you could imagine would end up in here, e.g. @ignore, @timeout, etc. as you mentioned below. Maybe there aren't actually that many though.

ducktape/tests/test.py
granders: Added simple filter for this

ducktape/tests/test.py
granders: okeedoke

ducktape/decorate.py
granders: Sigh, ok...

ducktape/tests/loader.py
granders: Yep that's the idea.

granders:
@ewencp Ready for another look.

ducktape/tests/test.py
```diff
@@ -160,11 +203,12 @@ def test_name(self):
...
-        name_components = [self.module,
-                           self.cls.__name__ if self.cls is not None else None,
-                           self.function.__name__ if self.function is not None else None]
+        name_components = [self.module_name,
+                           self.cls_name,
+                           self.function_name,
+                           self.injected_args_name]
 
-        return ".".join(filter(lambda x: x is not None, name_components))
+        return ".".join(filter(lambda x: len(x) > 0 and x is not None,  name_components))
```
ewencp: The order of the conditionals seems wrong, need to check x is not None first or len(None) will throw a TypeError.

ewencp:
@granders Found one easy to fix bug, the rest LGTM. Do we need to update the README and/or is it time to setup real docs? As we add more features like this we're going to build up quite a backlog of docs to write...

granders:
@ewencp Yes I think it is about time to set up real docs...

Issue #62, opened 2015-06-25
min_cluster_size error can be a bit nicer


gwenshap:
Currently we have:
There aren't enough available nodes to satisfy the resource
request. Your test has almost certainly incorrectly implemented its
min_cluster_size() method.

Will be nice to include:
* Expected cluster size
* Where is it actually configured (in case I want to fix my incorrect setting)
* Maybe even justification ("We need 3 brokers, 1 zk, 1 runner")
* Actual cluster size
* How did you get the actual size (in case there is a bug and you are miscounting)
* Output of Vagrant Status, so we will see who is right
* "Your test has almost certainly incorrectly implemented its
min_cluster_size() method" should be changed to "Either your test incorrectly implemented its
min_cluster_size() method or you did not start enough Vagrant workers for the test to execute."

Also, kittens and unicorns :)

granders:
closed via #67

PR #67, opened 2015-07-07
Issue 62 min cluster

granders:
@gwenshap @ewencp 
Addresses #62 

Tried to clarify error messages when the cluster is too small or doesn't have enough free nodes by outputing info on: cluster size, number of nodes requested, number of available nodes. Output also includes info on how many nodes have been allocated to each different service so far.

The message about min_cluster_size is misleading - it's actually ok for min_cluster_size to underestimate. min_cluster_size is at best a lower bound because new services can be 'registered' at any time in a test.

I don't have any Vagrant - specific output in the error messages since we may not be using a Vagrant cluster, and there's way for us to know where the cluster size is actually being configured in general.

### DIFF
--- a/ducktape/cluster/remoteaccount.py
+++ b/ducktape/cluster/remoteaccount.py
@@ -30,13 +30,12 @@ class RemoteAccount(HttpMixin):
         self.externally_routable_ip = None
         self.logger = logger
 
-    def __repr__(self):
-        r = ["hostname: %s" % self.hostname,
-             "user: %s" % ("" if self.user is None else self.user),
-             "ssh_hostname: %s" % ("" if self.ssh_hostname is None else self.ssh_hostname),
-             "ssh_args: %s" % ("" if self.ssh_args is None else self.ssh_args)]
-
-        return "<RemoteAccount: " + ", ".join(r) + ">"
+    def __str__(self):
+        r = ""
+        if self.user:
+            r += self.user + "@"
+        r += self.hostname
+        return r
 
     @property
     def local(self):
@@ -194,10 +193,4 @@ class RemoteAccount(HttpMixin):
                 return
             raise e
 
-    def __str__(self):
-        r = ""
-        if self.user:
-            r += self.user + "@"
-        r += self.hostname
-        return r
 
--- a/ducktape/cluster/remoteaccount.py
+++ b/ducktape/cluster/remoteaccount.py
@@ -28,11 +28,15 @@ class RemoteAccount(HttpMixin):
         self.ssh_args = ssh_args
         self.ssh_hostname = ssh_hostname
         self.externally_routable_ip = None
-
         self.logger = logger
 
-    def set_logger(self, logger):
-        self.logger = logger
+    def __repr__(self):
+        r = ["hostname: %s" % self.hostname,
+             "" if self.user is None else self.user,
+             "" if self.ssh_args is None else self.ssh_args,
+             "" if self.ssh_hostname is None else self.ssh_hostname]
+
+        return "<RemoteAccount: " + ", ".join(r) + ">"
 
     @property
     def local(self):
--- a/ducktape/cluster/remoteaccount.py
+++ b/ducktape/cluster/remoteaccount.py
@@ -32,9 +32,9 @@ class RemoteAccount(HttpMixin):
 
     def __repr__(self):
         r = ["hostname: %s" % self.hostname,
-             "" if self.user is None else self.user,
-             "" if self.ssh_args is None else self.ssh_args,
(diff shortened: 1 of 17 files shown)


ducktape/cluster/remoteaccount.py
```diff
@@ -28,11 +28,15 @@ def __init__(self, hostname, user=None, ssh_args=None, ssh_hostname=None, logger
...
         self.externally_routable_ip = None
-
         self.logger = logger
 
-    def set_logger(self, logger):
-        self.logger = logger
+    def __repr__(self):
+        r = ["hostname: %s" % self.hostname,
+             "user: %s" % ("" if self.user is None else self.user),
+             "ssh_hostname: %s" % ("" if self.ssh_hostname is None else self.ssh_hostname),
```
ewencp: I assume you're using `__repr__` because containers use it for elements in their `__str__`? Not sure we want to use the default for containers here. This is going to get pretty noisy if you include all of this (especially ssh_args, which has a *lot* of flags under Vagrant). Wouldn't just printing out a str representation with a couple of pieces of info be enough?

ducktape/services/service.py
```diff
@@ -78,7 +88,14 @@ def allocate_nodes(self):
...
 
         self.logger.debug("Requesting nodes from the cluster.")
-        self.nodes = self.cluster.request(self.num_nodes)
+
+        try:
+            self.nodes = self.cluster.request(self.num_nodes)
+        except RuntimeError as e:
+            if hasattr(self.context, "services"):
+                msg = e.message + " Currently registered services: " + str(self.context.services)
+                self.logger.debug("Currently registered services: " + str(self.context.services))
```
ewencp: redundant, just reuse the string listing the registered services

ducktape/tests/loader.py
```diff
@@ -55,10 +55,13 @@ class TestLoader(object):
 
     def __init__(self, session_context):
         self.session_context = session_context
-        self.logger = session_context.logger
         self.test_file_pattern = DEFAULT_TEST_FILE_PATTERN
         self.test_function_pattern = DEFAULT_TEST_FUNCTION_PATTERN
 
+    @property
```
ewencp: Why are all these fields being converted into property methods? Did something change that made them dynamic such that we need to look them up every time?

setup.py
```diff
@@ -42,6 +42,7 @@ def run_tests(self):
       url="http://github.com/confluentinc/ducktape",
       packages=find_packages(),
       package_data={'ducktape': ['templates/report/*']},
-      tests_require=['pytest'],
+      install_requires=['jinja2', 'requests'],
+      tests_require=['pytest', 'jinja2', 'requests'],
```
ewencp: Pretty sure you don't have to specify all the install_requires in the tests_require -- they are combined for test. If you want to verify, you could test this pretty easily in a VM.

It'd also be nice to keep these types of changes isolated to their own changesets, even if they are trivially small.

ewencp:
@granders Unit tests? There are already tests for the cluster classes that should have at least minor updates. It'd be good to get coverage of how the test runner handles these cases given constrained resources, e.g. a mock cluster with only a couple of nodes should fail in the expected (graceful) way.

setup.py
granders: confirmed this in a VM

ducktape/tests/loader.py
granders: I realized it was potentially unsafe to assume those fields never changed.

For example, the main command-line logic instantiates the session context with a 'None' cluster and then instantiates the cluster and populates this field only if tests will be run.

granders:
@ewencp Ready for another look. Updated unit tests, added one for test runner

tests/runner/check_serial_runner.py
```diff
@@ -0,0 +1,56 @@
...
+        mock_cluster = MagicMock()
+        mock_cluster.__len__.return_value = 1
+        session_context = tests.ducktape_mock.session_context(mock_cluster)
+
+        test_context = TestContext(session_context, module=None, cls=TestThingy, function=TestThingy.test_pi)
+        runner = SerialTestRunner(session_context, [test_context])
+        runner.log = Mock()
+
+        # Even though the cluster is too small, the test runner should this handle gracefully without raising an error
+        runner.run_all_tests()
```
ewencp: `run_all_tests` returns all the results. Could we just check that -- both to verify that `run_all_tests` actually ran the test and caught the error, and to avoid having to do the more complicated version below?

ewencp:
@granders LGTM. I had a minor comment on the unit test that might make it simpler, but feel free to merge.

granders:
@ewencp good call. Thanks for the review

PR #71, opened 2015-07-13
Cleaned up a few things in loader debug logging.

granders:
Reduced log noise while trying to import a file - this makes clear which file actually failed to import.

@ewencp - pretty tiny change here, should be quick.

### DIFF
--- a/ducktape/tests/loader.py
+++ b/ducktape/tests/loader.py
@@ -133,9 +133,10 @@ class TestLoader(object):
             if len(test_info_list) == 0:
                 raise LoaderException("Didn't find any tests for symble %s." % symbol)
 
-        self.logger.debug("Discovered these tests: " + str(test_info_list))
-        return [TestContext(self.session_context, t.module_name, t.cls, t.function, t.injected_args)
-                for t in test_info_list]
+        test_list = [TestContext(self.session_context, t.module_name, t.cls, t.function, t.injected_args)
+                     for t in test_info_list]
+        self.logger.debug("Discovered these tests: " + str(test_list))
+        return test_list
 
     def import_modules(self, file_list):
         """Attempt to import modules in the file list.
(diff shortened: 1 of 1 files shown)


ducktape/tests/loader.py
```diff
@@ -154,21 +155,23 @@ def import_modules(self, file_list):
...
                 except Exception as e:
-                    self.logger.debug("Could not import " + module_name + ": " + e.message)
-                    self.logger.debug(
-                        "   But it's ok: at most one import should work per module, so import failures are expected.")
                     continue
                 finally:
                     path_pieces = path_pieces[1:]
 
+            if not success:
+                self.logger.debug("Unable to import %s" % f)
```
ewencp: Only difference between this and the previous version is that this one isn't as clear that multiple imports are attempted with additional components. This log might be misleading since it says it's unable to import the file, but one of the intermediate pieces could work as well, right?

ewencp:
@granders Very minor comment you might want to address, but LGTM.

ducktape/tests/loader.py
granders: Hm, actually this change was almost entirely the reason for this tiny PR.

For any given file, there are a bunch of import attempts until one works, or until all attempts fail.
We really just want to know whether an import succeeded overall (at least on import path worked), or if it failed overall (all import paths failed for that file).

Previously the logging for this was quite noisy, since each file would have many failed import attempts, even though the overall import of the given file was successful, and I found it hard to see when a file truly failed to import.

ewencp: Sorry, I should have been clearer. Cleaning up the logs so it doesn't spit out about 80 debug messages for a normal, pretty small set of tests is good. I think that change makes sense.

I was just pointing out that this terse message may not give enough info. But now that I think about it more, the only way it can be confusing is if the set of files we're filtering to isn't restrictive enough, but since we filter down to `.py` files, I think this is fine.

PR #73, opened 2015-08-06
Issue 72 symlink

### LINKED ISSUE #72: ducktape fails if results/latest points to nothing
results/latest is a symlink

If the directory to which `results/latest` points is removed (but the link is not), the next ducktape run will fail (e.g. http://jenkins.confluent.io/job/kafka_system_tests/18/console)

```
    if os.path.exists(latest_test_dir):  # Problem: in this case, os.path.exists -> False
        os.unlink(latest_test_dir)
    os.symlink(results_dir, latest_test_dir)  # Now this fails because latest_test_dir was not unlinked
```

granders:
Fixes #72, added a few corresponding unit tests

@ewencp  Mind taking a quick look?


ewencp:
@granders LGTM

Issue #37, opened 2015-04-16
configurable test output location


granders:
User should be able to specify where the output of one run of ducktape gets writting (e.g. ./results or <path_to_all_test_results>)

A sensible default would go in command_line.config

ewencp:
Fixed by #75.

PR #75, opened 2015-08-12
Configurable results dir

granders:
@ewencp 

- added command-line option for configuring a custom directory for results. Addresses #37
- previously calling `ducktape --version` had the side effect of creating a metadata pwd/.ducktape. Removed this side-effect


ewencp:
LGTM

PR #77, opened 2015-08-13
Include class fields in template rendering context so constants defined on the class can be used easily in templates.

ewencp:


ewencp:
@granders Ran into this while writing a test -- my Copycat test class needs to specify a filename to use and it is needed as a constant in the test itself and in the config file generated for the process. The most natural place for this was as a constant in the test class, but the `self.__dict__` passed in as context does not include class-level fields. This is a simple fix to add the class level `__dict__` as well.

granders:
LGTM

PR #78, opened 2015-08-13
Make ssh utilities return exit status when failure is allowed so the caller can check whether the command was successful.

ewencp:
@granders Quick and easy review. No tests because we don't have a good way to unit test anything involving ssh currently....

### DIFF
--- a/ducktape/cluster/remoteaccount.py
+++ b/ducktape/cluster/remoteaccount.py
@@ -180,17 +180,21 @@ class RemoteAccount(HttpMixin):
     def _ssh_quiet(self, cmd, allow_fail=False):
         """Runs the command on the remote host using SSH and blocks until the remote command finishes.
 
-        If it succeeds, there is no output; if it fails the output is printed and the CalledProcessError is re-raised.
+        If it succeeds, there is no output; if it fails the output is printed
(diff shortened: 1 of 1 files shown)


ducktape/cluster/remoteaccount.py
```diff
@@ -180,17 +180,21 @@ def create_file(self, path, contents):
     def _ssh_quiet(self, cmd, allow_fail=False):
         """Runs the command on the remote host using SSH and blocks until the remote command finishes.
 
-        If it succeeds, there is no output; if it fails the output is printed and the CalledProcessError is re-raised.
+        If it succeeds, there is no output; if it fails the output is printed
```
granders: There appear to be 4 methods which all call this which will now return a status instead of None.

Should the comments on those methods be updated to reflect this behavior?

granders:
@ewencp One small question, otherwise LGTM

ducktape/cluster/remoteaccount.py
ewencp: I updated the docs for `ssh()`. For the rest, it's not relevant since they don't use `allow_fail=True` and we weren't making any guarantees about what they previously returned.

PR #79, opened 2015-08-13
Update wait until

granders:
@ewencp 
wait_until now raises an Exception with the desired error message instead of returning False if it times out


ewencp:
LGTM

PR #80, opened 2015-08-13
When searching for tests, log import failures that aren't ImportErrors as errors since it most likely indicates the code has a syntax error and might cause tests to be incorrectly skipped.

ewencp:
@granders Another really simple one. I chose to filter by ImportError even though it's not a perfect heuristic. Ignoring ImportErrors is necessary because of the way we construct the module names to import (by using all file system path components, joining with dots, and then removing one element at a time until we find a way to load the file). What we lose by doing it this way is if the file is supposed to be a test but tries to import something non-existent, which also obviously causes an ImportError.

### DIFF
--- a/ducktape/tests/loader.py
+++ b/ducktape/tests/loader.py
@@ -154,7 +154,7 @@ class TestLoader(object):
                 raise Exception("Expected absolute path ending in '.py' but got " + f)
 
             # Try all possible module imports for given file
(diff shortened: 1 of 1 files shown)


granders:
Ok aside from adding a comment on the somewhat unintuitive filtering of ImportError, LGTM

PR #82, opened 2015-08-14
Display test description and data in main html report

### LINKED ISSUE #66: Test report/output should include test description
Ideally test descriptions should be easily discoverable in the test reports.

ducktape could without much trouble provide a mechanism for grabbing test descriptions and adding to the reports. Simplest most natural implementation I can think of would be to grab docstrings from test methods.

### LINKED ISSUE #70: update html report to include test data
Some tests now return data (e.g. various performance tests)

This data is now available in Result objects, and should be visible on generated html reports

granders:
@ewencp 
This is a super minimal implementation, but better than nothing I think.

fixes #70 
fixes #66

### DIFF
--- a/ducktape/templates/report/report.html
+++ b/ducktape/templates/report/report.html
@@ -65,11 +65,11 @@
           var className = this.props.test.test_result + "Class"
           return (
             <tr className={className}>
-              <td colSpan='5' align='center'>{this.props.test.test_name}</td>
-              <td colSpan='5' align='center'>{this.props.test.description}</td>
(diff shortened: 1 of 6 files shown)


ewencp:
@granders This basically looks good to me. One tweak: the  names print out like this: `<module=kafkatest.tests.replication_test, cls=ReplicationTest, function=test_clean_bounce, injected_args=None>` it'd be nice if they were formatted more nicely.

We'll eventually need to be able to hide the data since it can be kind of large and makes it harder to scan the output, but we can just defer that to a later patch that also cleans up other styling stuff on the UI.

PR #83, opened 2015-08-16
issue 81

### LINKED ISSUE #81: Difficult to discover bugs in cleanup code
When developing a test, I started noticing weird behavior and eventually tracked it down to the fact that my service had not been cleaning up properly. There were still instances of it running when I ran the test again and it caused the test to break (they were auto-registering Kafka topics when the test wasn't expecting them to).

It took me awhile to discover this because even though there was a log indicating that the stop method was failing, it was buried in the info/debug logs and at the WARNING level:

```
[WARNING - 2015-08-14 10:20:39,176 - service_registry - stop_all - lineno:37]: Error stopping service <CopycatStandaloneService: num_nodes: 1, allocated: True, nodes: ['worker3']>: pids() takes exactly 1 argument (2 given)
```

Any reason this is only at the WARNING level? I think the obvious thing to do is log it at the error level, make sure it gets logged to the console (not just in the log files that I'd have to look at separately).

granders:
@ewencp Tiny change

fixes #81 

I didn't change the log level of those cleanup calls since I suspect some are not true errors. Instead, WARNING and above are logged now to stdout by default (and DEBUG and above with --debug flag)


ewencp:
I think the basic idea makes sense, but I think a common pattern is going to lead to a lot of noise. For example, I use `wait_until` with an SSH command that I expect to fail until the condition I'm waiting for is satisfied. This leads to a bunch of warning output:

```
[WARNING - 2015-08-17 10:26:06,155 - remoteaccount - _ssh_quiet - lineno:199]: Error running remote command: ssh vagrant@worker2 -o 'HostName 127.0.0.1' -o 'Port 2200' -o 'UserKnownHostsFile /dev/null' -o 'StrictHostKeyChecking no' -o 'PasswordAuthentication no' -o 'IdentityFile /Users/ewencp/kafka.git/.vagrant/machines/worker2/virtualbox/private_key' -o 'IdentitiesOnly yes' -o 'LogLevel FATAL'  'grep '\''Kafka Server.*started'\'' /mnt/kafka.log'
```

Or maybe the problem here is more that having a warning for every SSH command that has a non-zero exit status is the problem?

More generally, have you run kafka tests and muckrake tests with this patch to see what the output looks like even when everything passes just to sanity check it's not too noisy?

granders:
@ewencp 
I've run a subset with this, and the noise level is not too bad from my vantage point.

However, `wait_until` pattern will become much more prevalent, and will (significantly) crank up the noise. I actually think that the logging in `_ssh_quiet` should be downgraded to INFO or below since running it in tight loops with multiple failures is now expected.

ewencp:
Agreed. I think it'll be too common to run a variety of commands that might fail just to check the current state of the system, so it doesn't make sense at its current level. Can you just make that change as part of this patch? Then it LGTM as long as the noise from other tests is also fine -- that was the only noise I saw in the small set of tests I checked.

PR #84, opened 2015-08-19
Various modifications for Kafka client tests

edenhill:


ducktape/command_line/main.py
```diff
@@ -147,6 +147,10 @@ def main():
             print "    " + str(test)
         sys.exit(0)
 
+    if len(tests) == 0:
```
granders: Maybe we should see this message even if --collect-only is flagged? (I.e. move this above the `if args.collect_only` block)

Also, "No test classes found" -> "No tests found"

ducktape/tests/reporter.py
```diff
@@ -117,6 +117,30 @@ def header_string(self):
 
         return "\n".join(header_lines)
 
+    def result_string(self, result):
```
granders: Move these changes to SingleResultReporter.result_string

ducktape/tests/runner.py
```diff
@@ -82,9 +82,9 @@ def run_all_tests(self):
                 self.log(logging.INFO, "PASS")
 
             except BaseException as e:
-                self.log(logging.INFO, "FAIL")
+                self.log(logging.ERROR, "FAIL")
                 result.success = False
-                result.summary += e.message + "\n" + traceback.format_exc(limit=16)
+                result.summary += str(e.message) + "\n" + str(traceback.format_exc(limit=16))
```
granders: ah good, I got bit by this recently

ducktape/tests/reporter.py
```diff
@@ -117,6 +117,30 @@ def header_string(self):
...
+
+        if result.data is not None:
+            if type(result.data) in [str, unicode]:
+                # Result is a string blob
+                result_lines.append(result.data)
+            else:
+                # Result is a dict
+                result_lines.append(json.dumps(result.data))
+
+        result_lines.append("-----------------------------------------------------------------------------------------")
```
granders: `append("-" * self.width)`

ducktape/command_line/config.py
```diff
@@ -38,7 +38,7 @@ class ConsoleConfig(object):
     RESULTS_ROOT_DIRECTORY = "./results"
 
     SESSION_LOG_FORMATTER = '[%(levelname)s:%(asctime)s]: %(message)s'
-    TEST_LOG_FORMATTER = '[%(levelname)-5s - %(asctime)s - %(module)s - %(funcName)s - lineno:%(lineno)s]: %(message)s'
+    TEST_LOG_FORMATTER = '[%(levelname)-5s %(asctime)s - %(module)s:%(lineno)s %(funcName)s]: %(message)s\033[0m'
```
granders: control character at the end makes it bold?

ducktape/tests/test.py
```diff
@@ -250,6 +250,15 @@ def configure_logger(self):
         info_fh.setFormatter(formatter)
         debug_fh.setFormatter(formatter)
 
+        if self.session_context.debug:
+            # Colorize output if debugging is enabled
+            # FIXME: Might want to check for terminal capability too.
+            logging.addLevelName(logging.CRITICAL, "\033[41;33mCRIT ")
```
granders: Why truncate printout of "CRITICAL"?

edenhill: Output alignment and screen estate frugality, otherwise the timestamp position will slip around depending on logging level which is annoying.

ducktape/command_line/config.py
edenhill: That control character removes any previous control codes (i.e., colors in this case).
The colors control codes are added with logging level name.

ducktape/command_line/main.py
edenhill: Ok

ducktape/utils/wait.py
```diff
@@ -0,0 +1,83 @@
+# Copyright 2015 Confluent Inc.
```
granders: My vote is keep this and ditch `wait_until`

As for time saved stuff... I guess it's useful if you're trying to convince people not to use wait.sleep, but ultimately do we really want to keep it around?

ducktape/utils/wait.py
```diff
@@ -0,0 +1,83 @@
...
+    global time_saved_sum
+    return time_saved_sum
+
+
+def sleep (seconds):
+    """ Simple wrapper for time.sleep() but indicates that the programmer
+        made a conscious decisions not to use wait.until() """
+    return time.sleep(seconds)
+
+def until0 (timeout, expect_result, test_func, *args):
```
granders: This is potentially useful externally and thus deserves a snazzier name - perhaps `until_result`?
Also, why not include `**kwargs` in each of the three existing `until*` methods?

ducktape/command_line/config.py
ewencp: Is this actually safe to just include here, is it? I ask because I've definitely seen cases where escape codes bleed through when used indiscriminately, e.g. they fill automated build logs with a lot of noise because the escape codes are not filtered out. Is there any guidance in the logging package about how to handle this, or does it already strip out escape codes when you're not connected to a terminal? Alternatively, I think we have some detection of whether we're in a terminal and the size of the terminal somewhere; maybe we can use that to decide whether it's safe to add this escape code or not?

ducktape/tests/action.py
```diff
@@ -0,0 +1,230 @@
...
+# Provides pass/fail/aux-results reporting for specific test actions without
+# necessarily affecting the entire test outcome.
+#
+
+import traceback
+import inspect
+import time
+
+class TestAction (object):
+    # Accumulated test actions in various forms.
```
ewencp: This could use some docstring. I'm not sure how it's supposed to be used....

ducktape/tests/action.py
```diff
@@ -0,0 +1,230 @@
...
+                        inherited to all sub-actions.
+        """
+        
+        super(TestAction, self).__init__()
+        self.name       = name
+        self.instance   = instance
+        self.desc       = desc
+        self.logger     = logger
+        self.start_time = time.time()
+        self.duration   = 999.999
```
ewencp: odd choice for this default value. would `None` or an impossible value as a sentinel (e.g. `-1`) be better? I think this is basically a "has not run yet" constant.

ducktape/tests/action.py
```diff
@@ -0,0 +1,230 @@
...
+        # Resolve parent
+        if type(parent) == str:
```
ewencp: If you're going to check like this, `isinstance(parent, basestring)` is better here (and if we care about Python 3 you may need `six.string_types`, though so far we haven't worried about that).

But also, I'm also not sure about the logic here. If they specify a parent as a string and it doesn't exist in `actions`, is just removing the parent information what we want to do? Shouldn't that be considered an error? Also, is the lookup in `actions`, which seems to be a list of `TestActions` right? Or should we be looking it up in `actions_by_fullname`?

ducktape/tests/action.py
```diff
@@ -0,0 +1,230 @@
...
+            # Use parent's logger
+            if not self.logger and self.parent.logger:
+                self.logger = self.parent.logger
+            # Prefix our name with parent's name
+            prefix = self.parent.fullname + '.'
+
+
+        # If there is no description but the calling function has a doc string, use that,
+        # unless the parent action has the same one (a bit repetative).
+        # FIXME: Doesnt seem to work for class functions
```
ewencp: I assume that using the calling stack frame has some motivation, but I'm unclear why so right now I'd suggest just dropping this bit of code. Is there any reason not to use the parent as a fallback and nothing if there is no parent or the parent has no doc (and perhaps continue upward as necessary)?

ducktape/tests/action.py
```diff
@@ -0,0 +1,230 @@
...
+        #    if not self.parent or doc != self.parent.desc:
+        #        self.desc = doc
+
+        if instance != None:
+            self.fullname = '%s%s@%s' % (prefix, name, instance)
+        else:
+            self.fullname = prefix + name
+
+        if not self.logger:
+            self.logger = self._dummy_logger
```
ewencp: Does this work? Don't we usually make `self.logger` a `logging.Logger` object, which seems incompatible with it being a bound method like `self._dummy_logger`?

ducktape/tests/action.py
```diff
@@ -0,0 +1,230 @@
...
+
+        res_list = [res]
+        if len(self.subs) > 0:
+            for sub in self.subs:
+                res_list.extend(sub.get_result())
+        return res_list
+
+    def report (self, reporter):
+        """ *ActionReporter() feeder
+            FIXME: This should be consolidated with get_results() above """
```
ewencp: Agreed, I feel like the control is inverted here. This should just be a data object and the reporter should be responsible for traversing it's standard layout unless we think there will be multiple subtypes of `TestAction`.

ducktape/tests/reporter.py
```diff
@@ -166,9 +190,17 @@ def format_result(self, result):
...
             "description": result.description,
             "run_time": format_time(result.run_time),
-            "data": "" if result.data is None else json.dumps(result.data, sort_keys=True, indent=2, separators=(',', ': ')),
-            "test_log": self.test_results_dir(result)
+            "data": "",
+            "test_log": self.test_results_dir(result),
+            "actions": list()
         }
+
+        if type(result.data) == dict and 'actions' in result.data:
```
ewencp: Not sure I have a better approach, but this seems brittle...

ducktape/tests/test.py
ewencp: Yeah, this is actually really nice, but I think a check on whether we're in a terminal is a must. These are littering my logs with escape codes. Unfortunately I don't think there's an integrated way to handle this, but an approach [like this](http://stackoverflow.com/questions/384076/how-can-i-color-python-logging-output) might work, although I'd rather not have to do string replacement like that. Something that detects the terminal and only applies these changes (and the other one that resets the formatting) to the loggers being sent to a terminal would be much better.

ducktape/utils/wait.py
ewencp: Normally in Python you'd just use a `lambda` and the whole question about `*args` and `**kwargs` would be moot.

On naming, agree that `until0` is a very C-like name, though I'm not sure we want to treat this as public interface -- the two wrappers already cover all the cases, passing in the expected result in explicitly rather than using the easier-to-read wrappers should probably be discouraged. `_until` would give the indication that it's private and avoid `*` imports.

ducktape/utils/wait.py
```diff
@@ -0,0 +1,83 @@
...
+    abs_tmout = start_time + timeout
+
+    interval = timeout / 50.0
+
+    while time.time() < abs_tmout:
+        r = False
+        try:
+            r = test_func(*args)
+        except:
+            traceback.print_exc()
```
ewencp: This goes to stdout -- shouldn't it be going to some logger?

I think we've struggled a bit with this in other utilities (e.g. we have to set a logger on `RemoteAccount` and what we decide to log and at what level has been pretty ad hoc and has only kind of worked so far).

Can we choose one of these options?:
1. Treat exceptions as an accepted failure mode and don't log anything. In other words, callers should expect that *any* exception thrown by the test_func will simply be treated as not meeting the condition. (Note that this gets a bit weird for `until_not`, which could exit due to an unrelated exception rather than bec).
2. Don't handle the exception at all. The caller should be implementing something that actually returns True or False, and if it throws an exception we consider that an error and say we can't be sure of what it means (by letting it propagate up the stack).

ewencp:
@edenhill Would love to see some unit tests in there as well. Something quick and basic for `TestAction` should cover most of its functionality (and it should definitely be easy to unit test). The `until` functions also lend themselves to easy unit tests (albeit annoyingly requiring non-trivial runtime).

We don't have it now, so we could defer until later too, but if we don't have a good way to unit test the reporter classes, we might want to at least have a set of dummy tests that we can run ducktape against to generate a report that we can just sanity check manually (e.g. a basic test, one that returns results, on with `TestAction`s, etc).

ewencp:
Reopening since I think the reference from #88 was a mistake and should have pointed at #85

PR #86, opened 2015-08-23
Add a utility for monitoring a log until message appears.

ewencp:


ewencp:
@granders Moved this from https://github.com/apache/kafka/pull/150/ into ducktape. I'll reuse it there if we get this in a released version (i.e. 0.3.1).

granders:
@ewencp This is great
I was just thinking however about what would be good way to test this. 

One idea is to create something similar to the `kafkatest/sanity_checks`, and add a simple correctness check. But in order to check correctness here, I think you'd need to perhaps return the contents of the line located with the ssh command.

Having the contents of that line could be potentially useful for users as well?

LGTM as long you add a tracking issue at least.

granders:
Oops, one more thing:
in `ducktape/__init__.py`, update the version to 0.3.1

ewencp:
@granders Updated. We've gone full inception and added ducktape system tests for ducktape.

I kept the Vagrantfile simpler here because I don't think we need AWS testing for this -- these "system" tests should never grow beyond a couple of nodes to test the SSH functionality, anything used to induce partitions, etc.

Also bumped the version number. I had figured that would be done after merging this as part of the process of actually generating the release.

granders:
@ewencp I've been dreaming of ducktape inception since 0.1.0, now my dream has come true :)

Vagrantfile
```diff
@@ -0,0 +1,58 @@
...
+    override.hostmanager.ignore_private_ip = false
+
+    # Brokers started with the standard script currently set Xms and Xmx to 1G,
+    # plus we need some extra head room.
+    vb.customize ["modifyvm", :id, "--memory", ram_megabytes.to_s]
+  end
+
+  ## Cluster definition
+  (1..num_workers).each { |i|
+    name = "worker" + i.to_s
```
granders: Maybe `ducktape-worker + i.to_s` per conversation?

ducktape/errors.py
```diff
@@ -0,0 +1,16 @@
+# Copyright 2015 Confluent Inc.
```
granders: Couple questions:
- Since you've added an errors file, would LoaderError (loader.py) belong here?
- I've read some sources which recommend if we roll our own errors, they should subclass a common base class, e.g. something like `DucktapeError` What do you think?

Vagrantfile
ewencp: Changed to `"ducktape" + i.to_s`. `ducktape-worker` is far too long and inconvenient for me to type :)

By the way, in order to get the version bump in the right place in the git history, I had to rebase to reorder the commits and then force push -- that's why I usually try to keep the version change as a separate commit (and one that doesn't actually require a PR).

ducktape/errors.py
ewencp: 1. Yes, but since we're just adding these, lets defer those to separate patches. There are bunch of places we raise generic exceptions where we shouldn't. There are also other exceptions I'd like to change (e.g., not expose subprocess.CalledProcessError directly, instead converting it to our own type; that's an implementation detail and could potentially change if, for example, we used paramiko).
2. Yes, that's a good idea. Have updated the patch to include DucktapeError.

granders:
Sounds good, LGTM

PR #87, opened 2015-08-25
template.py - loader also searches packages for template files

granders:
Necessary for graduating `kafkatest` to an installable dependency.

@ewencp Let me know if this approach makes sense to you. 

Also, I haven't had any good ideas on unit-testing the "load from package" functionality. My only idea here was to add a simple unit-testable helper method for some basic sanity checks.


ewencp:
@granders LGTM. It's probably tough to unit test loading from a package. Does jinja have unit tests for PackageLoader?

Alternatively, if we start adding more system tests to validate functionality, this would be easy to test as well.

granders:
@ewencp It does have very simple tests here (https://github.com/mitsuhiko/jinja2/blob/master/tests/test_loader.py), although it's not super clear to me how this works

ewencp:
@granders I didn't dig in to check, but they are have pytest annotations on them. They probably exclude them from normal tests and only run them under specialized environments.

PR #88, opened 2015-08-26
Added timeout to BackgroundThreadService.wait method.

granders:
This was a convenient patch in which to get rid of `self._workers = None` 
fixes #84


granders:
@ewencp 
Hopefully a quick review. Added another ducktape ducktape test as well.

systests/services/test_background_thread_service.py
```diff
@@ -0,0 +1,76 @@
...
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+from ducktape.services.background_thread import BackgroundThreadService
+from ducktape.tests.test import Test
+from ducktape.errors import TimeoutError
+import time
+
+
+class BackgroundThreadTestService(BackgroundThreadService):
```
ewencp: Does this really need to be a system test? Why not just make it a unit test?

ewencp:
@granders I'd move the test into a unit test, but otherwise looks good to me.

granders:
@ewencp Good call, moving to unit test was easier than I expected

PR #92, opened 2015-08-27
Fixed RemoteAccount.wait_for_http_service and added unit tests. 

granders:
@ewencp 
This was broken by previous changes to behavior of wait_until.

### DIFF
--- a/ducktape/cluster/remoteaccount.py
+++ b/ducktape/cluster/remoteaccount.py
@@ -46,15 +46,14 @@ class RemoteAccount(HttpMixin):
         """Wait until this service node is available/awake."""
         url = "http://%s:%s%s" % (self.externally_routable_ip, str(port), path)
 
-        awake = wait_until(lambda: self._can_ping_url(url, headers), timeout_sec=timeout, backoff_sec=.25)
-        if not awake:
-            raise Exception("Timed out trying to contact service on %s. " % url +
-                            "Either the service failed to start, or there is a problem with the url.")
+        err_msg = "Timed out trying to contact service on %s. " % url + \
+                            "Either the service failed to start, or there is a problem with the url."
+        wait_until(lambda: self._can_ping_url(url, headers), timeout_sec=timeout, backoff_sec=.25, err_msg=err_msg)
 
     def _can_ping_url(self, url, headers):
         """See if we can successfully issue a GET request to the given url."""
         try:
-            self.http_request(url, "GET", "", headers)
+            self.http_request(url, "GET", "", headers, timeout=.1)
             return True
         except:
             return False
@@ -100,7 +99,6 @@ class RemoteAccount(HttpMixin):
             raise subprocess.CalledProcessError(proc.returncode, ssh_cmd)
         self.logger.debug("Ran cmd " + ssh_cmd + " -> " + str(stdoutdata))
         return stdoutdata
-
 
     def alive(self, pid):
         """Return True if and only if process with given pid is alive."""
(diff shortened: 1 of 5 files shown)


ducktape/cluster/remoteaccount.py
```diff
@@ -46,15 +46,14 @@ def wait_for_http_service(self, port, headers, timeout=20, path='/'):
...
-                            "Either the service failed to start, or there is a problem with the url.")
+        err_msg = "Timed out trying to contact service on %s. " % url + \
+                            "Either the service failed to start, or there is a problem with the url."
+        wait_until(lambda: self._can_ping_url(url, headers), timeout_sec=timeout, backoff_sec=.25, err_msg=err_msg)
 
     def _can_ping_url(self, url, headers):
         """See if we can successfully issue a GET request to the given url."""
         try:
-            self.http_request(url, "GET", "", headers)
+            self.http_request(url, "GET", "", headers, timeout=.1)
```
ewencp: Will this be reliable? Not sure about hard coded constants that small -- minor hiccups could pause something for 100ms. Is there a reason for this to be this small?

ducktape/utils/http_utils.py
```diff
@@ -25,4 +26,16 @@ def http_request(self, url, method, data="", headers=None):
...
+            # seems to raise errors when set to a number. Using an opener works however.
+            opener = urllib2.build_opener()
+            if timeout is None:
+                response = opener.open(req)
+            else:
+                response = opener.open(req, timeout=timeout)
+
+            return response
+
+        except Exception as e:
```
ewencp: Doesn't seem like this try/except is doing anything. Did you mean to translate it to a ducktape exception or should we just remove the block?

tests/cluster/check_remoteaccount.py
```diff
@@ -0,0 +1,87 @@
...
+        self.handler = SimpleHTTPServer.SimpleHTTPRequestHandler
+        self.httpd = SocketServer.TCPServer(("", self.port), self.handler)
+        self.close_signal = threading.Event()
+        self.server_started = False
+
+    def start(self, delay_sec=0.0):
+        """Run the server after specified delay"""
+        def run():
+            end_delay = time.time() + delay_sec
+            while time.time() < end_delay and not self.close_signal.is_set():
```
ewencp: Seems like this is checking things that it shouldn't need to. I thought the point of `Event` is that it doesn't have spurious wakeups so you should just be able to the body of this loop without the loop itself?

tests/test_utils.py
```diff
@@ -0,0 +1,29 @@
...
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+import socket
+
+
+def find_available_port(min_port=8000, max_port=9000):
```
ewencp: This actually isn't a reliable way of getting a port -- once you close a socket you cannot be sure of its availability. This was the source of a bunch of issues in the Kafka unit tests.

It may not matter much for the couple of tests we have here, just beware that this can be a problem and can lead to confusing test failures.

ewencp:
@granders Had a couple of comments, but LGTM. Feel free to merge after addressing the comments.

ducktape/utils/http_utils.py
granders: oops good point. updated

tests/cluster/check_remoteaccount.py
granders: interesting, I wasn't aware of this...

tests/test_utils.py
granders: Added a comment to this effect

ducktape/cluster/remoteaccount.py
granders: This is intended for use in a retry/backoff loop, so I guess this would only be a problem if an available connection consistently timed out.

250ms?

PR #94, opened 2015-08-28
Don't report SSH process errors to INFO since they are a normal part of many tests and generate a lot of noise in the info log.

ewencp:


ewencp:
@granders Trivial, but wanted to run this by you -- I'm kind of sick of seeing SSH commands in info logs, which makes it hard to scan that log to get an idea of what went wrong. I think this is getting a lot worse now because we have more SSH commands running that are expected to fail for awhile (due to `wait_until` and `monitor_log`).

granders:
@ewencp 
Makes sense, LGTM

PR #98, opened 2015-09-10
Cluster slot update

granders:
@ewencp 
This update slightly changes the behavior of node allocation.

Previously: when a Service was instantiated, it would have a `nodes` attribute consisting of an empty list which would be populated with ClusterSlot objects during node allocation.

Now: when a Service is instantiated, it has a `nodes` attribute consisting of `num_nodes` ClusterSlot objects, but these objects do not yet point to specific machines. So ClusterSlot objects become more like "logical nodes", and RemoteAccounts are more like the "physical nodes" to which the logical nodes eventually point.

Motivation: up until now, we've been ok with multi-node Services consist of identical nodes. With the need for upgrade tests and compatibility tests, this change makes it much more natural to express per-node attributes such as version, node-specific configuration etc, particularly since it can be useful to express those attributes before the physical nodes have been allocated and the services started.


ducktape/cluster/json.py
```diff
@@ -55,13 +54,12 @@ def request(self, nslots):
         result = []
         for i in range(nslots):
             node = self.available_nodes.popleft()
-            result.append(ClusterSlot(self, node, slot_id=self.id_source))
-            self.in_use_nodes.add(self.id_source)
-            self.id_source += 1
+            result.append(node)
```
ewencp: This changes the type of output for this method from list of ClusterSlots to list of RemoteAccounts, but there's not a corresponding change to LocalhostCluster.

Also, the ClusterSlot import is no longer used and can be removed.

ewencp:
@granders I'm not sure I understand the motivation here. How does having the ClusterSlot but not a remote account help you? Is the idea that you want to attach other attributes to the ClusterSlot before nodes are allocated for the test?

The current patch is a bit awkward because the whole point of the ClusterSlot class is that Cluster ensures it never hands out any RemoteAccount without it being wrapped in a ClusterSlot so that you can't end up with a RemoteAccount that isn't properly tied back to a Cluster object to free it to.

PR #99, opened 2015-09-11
Moved node allocation out of service.start and into Service.__init__

granders:
@ewencp 
Moved node allocation out of start, and into instantiation


ewencp:
LGTM. In addition to unit tests, did a basic sanity check with a couple of Kafka tests.

PR #101, opened 2015-09-21
test 2

granders:


granders:
take 2: cleaned up the test output by adding "bots team" to collaborators on ducktape repo

Issue #103, opened 2015-09-22
v0.3.3 can leak cluster nodes


granders:
An error in Service.free can cause cluster nodes to not be fully freed.

Cause: v.0.3.3 introduced logic which actually removes nodes from a list while iterating through the list.

PR #104, opened 2015-09-22
Fix Service.free logic

granders:
@ewencp 
Fairly small PR here. v0.3.3 had a dumb error resulting in the cluster gradually running out of nodes because nodes were being improperly freed by service classes with multiple nodes.

Fixes https://github.com/confluentinc/ducktape/issues/103


ducktape/services/service.py
```diff
@@ -92,13 +92,14 @@ def allocate_nodes(self):
         if self.allocated:
             raise Exception("Requesting nodes for a service that has already been allocated nodes.")
 
-        self.logger.debug("Requesting nodes from the cluster.")
+        self.logger.debug("Requesting %d nodes from the cluster." % self.num_nodes)
 
         try:
             self.nodes = self.cluster.request(self.num_nodes)
         except RuntimeError as e:
+            msg = ""
```
ewencp: Shouldn't the `str(e.message)` be outside the if statement?

ducktape/services/service.py
```diff
@@ -181,10 +184,11 @@ def free(self):
             node.account.logger = None
             self.free_node(node)
 
+        self.nodes = []
+
     def free_node(self, node):
```
ewencp: Can we get rid of `free_node`? It's not safe to call on its own -- it looks like `free` is supposed to be the only safe way to clean up nodes now. At a minimum, seems like this should be `_free_node`.

ducktape/tests/reporter.py
```diff
@@ -63,7 +63,7 @@ def result_string(self):
             result_lines.append("    " + self.result.summary)
 
         if self.result.data is not None:
-            result_lines.append(json.dumps(self.result.data))
+            result_lines.append(pprint.pformat(self.result.data))
```
ewencp: These changes seem unrelated to the rest of the patch. Why change to pprint? In the parts below where it writes to `.json` files, is the output guaranteed to be JSON compatible? It seems like the places where we cared about pretty printing we were already properly indenting the JSON.

ewencp:
Just so I'm sure I understand the issue correctly, the problem was that there was a call to `list.remove` while iterating over the `list` of nodes?

Left a few comments, but looks good. Feel free to merge after addressing comments (and deciding whether to include those `pprint` changes).

ducktape/services/service.py
granders: yep, done

ducktape/services/service.py
granders: done

ducktape/tests/reporter.py
granders: Yeah, this was a bit iffy.
I wanted to support the use case where functions are used to parameterize tests, e.g. `@parameterize(x=my_function)`, and `json.dumps` doesn't know how to stringify functions.

This is also tricky if we also want to be able to specify which parameters to use in a test at the command-line.

rolling this back

granders:
@ewencp 
Yes, that right, the problem was calling `list.remove` while iterating over the list.

PR #105, opened 2015-09-22
Postpone updating latest symlink until after tests run

granders:
@ewencp Trivial fix - latest symlink is only updated after tests run and reports are created.

### DIFF
--- a/ducktape/command_line/main.py
+++ b/ducktape/command_line/main.py
@@ -108,7 +108,9 @@ def setup_results_directory(results_root, new_results_dir):
             "A file or directory at %s already exists. Exiting without overwriting." % new_results_dir)
     mkdir_p(new_results_dir)
 
-    # Create or update symlink "latest" which points to the new test results directory
+
+def update_latest_symlink(results_root, new_results_dir):
+    """Create or update symlink "latest" which points to the new test results directory"""
     latest_test_dir = os.path.join(results_root, "latest")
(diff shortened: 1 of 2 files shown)


granders:
... and that's why we have a PR builder

ducktape/command_line/main.py
```diff
@@ -184,5 +186,6 @@ def main():
     reporter = HTMLSummaryReporter(test_results)
     reporter.report()
 
+    update_latest_symlink(args.results_root, results_dir)
```
ewencp: Is there any chance this could be lost to an exception? I obviously prefer the new behavior, but want to make sure it works well in case of failures too. Since it's deferred until the very end now, unexpected errors could cause the update to not occur.

Are we confident enough of the rest of the code in here that is not in a `try`/`catch` block that we don't need to put this in a `finally` clause? Or do we just think that if something does throw an exception, the issue is obvious and we should ignore the `latest` directory anyway?

ewencp:
One minor question, but I much prefer this behavior. Please merge ASAP :)

ducktape/command_line/main.py
granders: My thinking here was that any error that propagates to `main` (probably causing reports not to be generated) or which takes place in `main`  is catastrophic enough that it doesn't make sense to ensure the symlink points to the new report.

Issue #106, opened 2015-10-06
ssh_capture


granders:
@lindong28 observed that if you run
```
lines = account.ssh_capture(cmd, ...)
```
The remote command will not be run until until we iterate through the returned iterator because of the way python generators do lazy evaluation.

We would like to, at minimum, be able to 
```
lines = account.ssh_capture(cmd, ...)
do_something_else_while_cmd_starts_up_remotely()
for line in lines:
   process(line)
```

see https://github.com/apache/kafka/pull/275 for more discussion

PR #107, opened 2015-10-06
Updated ssh_capture

granders:
Remote command is now run even if the iterator is never accessed

Fixes https://github.com/confluentinc/ducktape/issues/106


granders:
This is a minimally invasive fix which won't be noticeable except in the newer use-case mentioned in https://github.com/apache/kafka/pull/275

ducktape/cluster/remoteaccount.py
```diff
@@ -81,14 +81,18 @@ def ssh_capture(self, cmd, allow_fail=False, callback=None):
...
+            for line in iter(proc.stdout.readline, ''):
+                if callback is None:
+                    yield line
+                else:
+                    yield callback(line)
+            proc.communicate()
+            if proc.returncode != 0 and not allow_fail:
+                raise subprocess.CalledProcessError(proc.returncode, ssh_cmd)
+
+        return output_generator()
```
ewencp: Ah, nice solution. I was thinking you were going to have to make an iterator class.

ewencp:
LGTM. It'll be a tiny release, but want to bump the version number so the Kafka patch can take advantage of this?

Issue #69, opened 2015-07-13
run parametrized tests with specific parameters


granders:
Now that we've added `@parametrize` and `@matrix`, we can concisely define a large group of related tests.

e.g.
```
@matrix(x=[1, 2, 3])
@matrix(y=['a', 'b', 'c']
def some_test(x, y):
    # ...
```
But when running with ducktape, if we run `some_test` specifically, it will now run 9 different tests. It would be very useful for poor test developers to be able to run `some_test` with specific parameters.

For example, if we wanted to run `some_test` with `x = 1` and `y = 'a'`, we might do something like:
`ducktape tests/the_test.py::TheTest.some_test --args {'x': 1, 'y': 'a'}`

granders:
As part of this patch, it may be useful for parametrized tests to output the injected args to make it easy to copy the arguments and target and run any given parametrization of the a given.

PR #108, opened 2015-10-07
Added command-line option allowing manual specification of injected args

granders:
@ewencp 
I've been meaning to get to this for a while, and it turned out to be reasonably simple.

It's quite a pain not to be able to specify a particular parametrization, and it seems more external developers are using parametrization, so I thought I'd take a quick stab at it.

Fixes https://github.com/confluentinc/ducktape/issues/69


ducktape/command_line/main.py
```diff
@@ -56,6 +57,7 @@ def parse_args():
                              "This is primarily useful for test developers who want to interact with running " +
                              "services after a test has run. ")
     parser.add_argument("--version", action="store_true", help="display version")
+    parser.add_argument("--inject", action="store", help="Inject these arguments into the specified test. Use JSON to format.")
```
ewencp: I can think of two possible ways to handle the general problem of running parameterized tests. One is what you're doing here, by specifying the arguments directly. I assume the reason for using JSON is that it's the only way to preserve the type information necessary to pass the arguments correctly.

The second approach is to have the test loader generate all the parameterized test names and then just look them up by name. For example `kafkatest.tests.copycat_test.CopycatStandaloneFileTest.test_file_source_and_sink.converter=org.apache.kafka.copycat.json.JsonConverter.schemas=True` is the full test name in one of the Kafka tests. Since we already have the test loader using filenames, the actual name passed on the command line would be something more like `kafkatest/tests/copycat_test.py::CopycatStandaloneFileTest.test_file_source_and_sink.converter=org.apache.kafka.copycat.json.JsonConverter.schemas=True`.

Any particular reason for going with the former approach? The latter is what I initially imagined, mainly because I figured it would be easy to copy/paste the test name from log files. I guess one reason the arguments approach is nice is that you can run variants of the test that aren't already listed?

ewencp: Also, can we change the command line argument name? I know you use `inject` in the implementation of parameterization, but I don't think it's very intuitive for users. `--parameters` maybe, which directly mirrors the naming of the annotation?

tests/loader/check_loader.py
```diff
@@ -96,7 +96,20 @@ def check_test_loader_with_class(self):
...
 
+        each test method annotated with @parametrize or @matrix should only expand to a single discovered test,
+        and the injected args should be those passed in from command-line.
+        """
+        injected = {"x": 1, "y": -1}
+        loader = TestLoader(self.SESSION_CONTEXT, injected_args=injected)
 
+        file = os.path.join(discover_dir(), "test_decorated.py")
+        tests = loader.discover([file])
+        assert len(tests) == 4
```
ewencp: I found this behavior surprising and confusing. Should we really allow these injected parameters to apply to multiple tests? In this test, `test_decorated.py` happens to have tests that all take the same parameters, but the output wouldn't make much sense if the tests took different parameters (or none at all).

In fact, it won't even run properly if some tests don't have matching parameters unless test methods always include `**kwargs` in their argument list.

Should we restrict this option to only work when you specify a specific test method (not even just a single test class, but all the way to the specific method to ensure you're targeting the right test)?

ewencp:
@granders Had some questions and tweaks, but basically looks good. How about also updating the README? In particular, we already have a bunch of examples of how to run specific tests, so maybe we can just add one more that includes the new flag.

ducktape/command_line/main.py
granders: Sure, in fact I'm not a huge fan of `inject`. Maybe `--parameters` or `--args`

tests/loader/check_loader.py
granders: Tests that take different parameters or none will result cause exceptions, and I think this is the right behavior.

Hmm I suppose you could raise an error if the loader finds more than one test, but I don't really see why it should have that restriction? It may be an unusual use-case to target more than one test this way, but I don't see why it should be an error condition.

ducktape/command_line/main.py
granders: I like the idea of being able to copypaste a one-line command from the log to target a particular test, but that could be implemented with either approach.

I actually have a strong preference for the former approach because it is better suited for exploratory testing. It can even be used on methods without parametrization: for example,
```
def test_something(self, x=-1, y=2):
   ...test logic...
```
```
ducktape test_something --args "{'x': 100, 'y': 15}"
```

ewencp: Yeah, I considered `--args` or `--arguments`. `--args` has the obvious benefit of brevity. `--parameters` has the benefit of consistency. I'd be happy with either one.

ewencp: Yeah, not requiring `@parametrize` or `@matrix` could be handy. I suspect it's not *that* big a deal to edit a method like that given that you're probably mucking with the code if you're running tests like that, but it's nice nonetheless.

I think my strongest negative reaction to the current approach is that even in the case where I just want to run an existing parameterization, I need to construct a carefully quoted JSON parameter.

That said, the two approaches aren't actually exclusive. We can easily make this version work, allowing you to pretty flexibly override test settings regardless of how the test is parameterized, and then later add test loader support for the other approach where you fully qualify the test name including parameters.

ducktape/command_line/main.py
```diff
@@ -130,6 +132,16 @@ def main():
         print ducktape_version()
         sys.exit(0)
 
+    injected_args = None
+    if args.inject:
+        try:
+            injected = args.inject
+            injected = injected.replace("'", "\"")
```
ewencp: I forgot to note before that I found this escaping confusing. Your example above clarifies why you did it though. You gave this as an example:

`ducktape test_something --args "{'x': 100, 'y': 15}"`

If you write it as 

`ducktape test_something --args '{"x": 100, "y": 15}'`

instead, then you don't need this escaping. Even if you didn't want to use the single quotes (which would, e.g., defeat variable substitution), you could write:

`ducktape test_something --args "{\"x\": 100, \"y\": 15}"`

I think we should avoid doing this escaping ourselves -- if we do it, then people trying to do valid things might get unexpected results. For example

`ducktape test_something --args "{\"param\": \"string value with aposrophe: it's\" }"`

will have the `'` character replaced incorrectly.

granders: ah, good point

granders:
@ewencp Updated README, addressed your comments

README.md
```diff
@@ -16,26 +16,25 @@ TBD
 Install
 -------
 
-    git clone https://github.com/confluentinc/ducktape.git
-    cd ducktape
-    python setup.py install
+    pip install ducktape
```
ewencp: Nice! Unrelated to this patch, but a welcome change :)

ewencp:
LGTM

Issue #109, opened 2015-10-13
Provide has_next() in the object returned by ssh_capture()


lindong28:
It can be useful to provide `has_next()` in the object returned by ssh_capture(), e.g. when we want to wait for a process to start. An example look like the following:

```
console_consumer_cmd = "..."
output = node.account.ssh_capture(console_consumer_cmd)
output.has_next() // wait until console_consumer has started
```

PR #110, opened 2015-10-14
Provide has_next() in the object returned by ssh_capture()

lindong28:
Provide has_next() in the object returned by ssh_capture() https://github.com/confluentinc/ducktape/issues/109


ewencp:
add to whitelist

granders:
@lindong28 
Note that `has_next` blocks until the next line of input is available. I suspect this will be a lot more useful if it's non-blocking, particularly on `ssh_capture`

Then we can do things like:
`wait_until(my_captured_ssh.has_next, timeout_sec=10, err_msg="Expected output to be available from remote process")`

One approach is to use the `select` module: https://docs.python.org/2/library/select.html 

I think you can do something like `select.select([proc.stdout], [], [], 0)` to check whether output is available on `proc.stdout` right now without blocking.

lindong28:
@granders I have intentionally made this call blocking to support the usecase that motivated this feature -- wait for first line from consumer before considering it as started. But as you mentioned, it is useful to provide a timeout here so we don't blocking infinitely if something goes wrong.

Sure, I will update the patch to support timeout. But do you think it will be more useful to support timeout directly in `has_next()`? For example, `has_next(timeout_sec=None)` will block. In your example usecase, the usage will look like:

```
started = my_capture_ssh.has_next(timeout_sec=10)
if started == false:
  print "Expected output to be available from remote process"
  sys.exit(-1)
```

granders:
@lindong28 good point, timeout in `has_next` makes sense. 

The next question is how should we signal "doesn't have next within timeout" - return False, or raise an exception? I suspect raising an exception is the most common use case here, so my initial vote is for raising an exception to signal timeout.

Also, regarding naming - the purpose of this mechanism is really to "block for a while until next is available or timeout", which is different than what I think of when I see `has_next` - is there a more intuitive name than `has_next`? 

I wonder if it would be useful to provide similar timeout option in `ssh_capture`? The same argument about blocking forever in `has_next` applies to blocking on each implicit `next` call when iterating through `ssh_capture()`. An optional timeout would behave similarly, but block for a maximum of `timeout_sec` or raise an exception before the next line of output is available.

lindong28:
@granders I agree with you that `has_next(timeout)` should raise an exception when it times out. My reason is to distinguish between "it doesn't have output" and "it is still waiting for output before timeout".

Regarding the naming, I think this call will also unblock and return False when it knows there is no output. For example, in the use case where we start a remote consumer and block waiting for consumer to start, `has_next()` will not be enough because the consumer may actually crash without providing any output. Therefore it probably makes sense to call it `has_next()`, right?

I am not sure if I understand your reason why "ssh_capture()" should also have timeout. The semantics of `ssh_capture()` is to execute a remote command remotely and wait for output. If you have timeout, are you going to timeout because "it has been more than 5 seconds before it gets a new line" or "cmd hasn't finished after 5 seconds"? Do you have specific use case for adding timeout in `ssh_capture`, suppose we already have `has_next(timeout)` from iterable object returned by `ssh_capture()`?

lindong28:
@granders Can you review updated patch? I have enabled timeout in `has_next()`. In answer to your previous question, I prefer to keep the name `has_next()` because we also rely on its output value.

ducktape/cluster/remoteaccount.py
```diff
@@ -236,6 +236,35 @@ def monitor_log(self, log):
...
+            return next(self.iter_obj)
+        next_obj = self.cached
+        self.cached = self.sentinel
+        return next_obj
+
+    def has_next(self, timeout_sec=None):
+        assert timeout_sec == None or self.descriptor != None, "should have descriptor to enforce timeout"
+
+        if self.cached is self.sentinel:
+            if timeout_sec == None or select.select([self.descriptor], [], [], timeout_sec)[0]:
```
granders: This won't work if you don't `import select` at the top of the file :)

lindong28: Ah.. thanks! I tested it on a standalone python file and copy past my changes to remoteaccount.py. But I forgot to add the import. Fixed now.

granders:
@lindong28 
Since you're using `select` to detect whether output is available from the stream within a timeout, it seems like the `cached` and `sentinel` fields are unnecessary

Also, we should add some unit tests - probably in the `tests/cluster/check_remote_account.py` (see ducktape readme for instructions on running unit tests with pytest).

ducktape/cluster/remoteaccount.py
granders: np :) also, unit tests will help with this kind of thing :)

lindong28:
@granders I think we can remove `cached` and `sentinel` if `descriptor` is always available. But I intentionally made `descriptor` an optional parameter so that developer may use `iter_wrapper` to wrap any iterable object without having `descriptor`. What do you think?

Sure thing, I will add unit test for this.

granders:
@lindong28 Ok now I understand. Hm, this is a handy helper function within `remoteaccount`, but given the way iterators are used in python, I'm not sure I see a use case for wrapping normal iterators?

lindong28:
@granders Yeah I don't have specific usecase for this. My motivation for this is that there is no reason we need descriptor to do has_next(). But I am not strong on this. Please let me know if you like the code to be simplified by removing `sentinel` and I will do that:)

granders:
@lindong28 Cool, yeah let's simplify it for now. 
We can generalize it if a use case does come up.

lindong28:
@granders I just realized that `sentinel` is necessary to implement `has_next()`. `proc.stdout` will be ready for reading even when it is end of file. You can test it by running the following example:

```
import select
import subprocess

class iter_wrapper(object):
    """
    Helper class that wraps around an iterable object to provide has_next() in addition to next()
    """
    def __init__(self, iter_obj, descriptor):
        self.iter_obj = iter_obj
        self.descriptor = descriptor

    def __iter__(self):
        return self

    def next(self):
        return next(self.iter_obj)

    def has_next(self, timeout_sec=None):
        print select.select([self.descriptor], [], [], timeout_sec)
        return select.select([self.descriptor], [], [], timeout_sec)[0]


def ssh_capture(cmd, allow_fail=False, callback=None):
    '''Runs the command via SSH and captures the output, yielding lines of the output.'''
    ssh_cmd = cmd
    proc = subprocess.Popen(ssh_cmd, shell=True, stdout=subprocess.PIPE, stderr=subprocess.STDOUT)

    def output_generator():
        for line in iter(proc.stdout.readline, ''):
            if callback is None:
                yield line
            else:
                yield callback(line)
        proc.communicate()
        if proc.returncode != 0 and not allow_fail:
            raise subprocess.CalledProcessError(proc.returncode, ssh_cmd)

    return iter_wrapper(output_generator(), proc.stdout)


output = ssh_capture("tail example.txt")

for i in range(20):
    if output.has_next(timeout_sec=2):
      print output.next()
    else:
      print "no ouput"
      break

```

lindong28:
@granders Could you let me know how to run existing test, e.g. check_remoetaccount.py? I didn't find it in README.

granders:
@lindong28 
See the unit test section in the readme here: https://github.com/confluentinc/ducktape
In summary, run `py.test` or `python setup.py test` in the ducktape directory to run ducktape unit tests

lindong28:
My bad.. It is actually in the README.

Not sure if I did anything wrong. By running "python setup.py test" for 313abbc8fcdb3e82d54563038f607b018a83d256 I got the following result:
`2 failed, 46 passed, 1 skipped, 30 error in 3.45 seconds`

I will write the test.

granders:
@lindong28 Have you succeeded in running the unit tests? If not, can you provide a larger snippet from the test output? Hard to tell much from just the summary statistics

lindong28:
@granders No.. I consistently get the following error. I was going to look into this but didn't get a chance.

```
_______________________________________________________________ ERROR collecting tests/tests/check_session.py ________________________________________________________________
import file mismatch:
imported module 'tests.tests.check_session' has this __file__ attribute:
  /home/dolin/research/ducktape/build/lib/tests/tests/check_session.py
which is not the same as the test file we want to collect:
  /home/dolin/research/ducktape/tests/tests/check_session.py
HINT: remove __pycache__ / .pyc files and/or use a unique basename for your test file modules
_________________________________________________________________ ERROR collecting tests/tests/check_test.py _________________________________________________________________
import file mismatch:
imported module 'tests.tests.check_test' has this __file__ attribute:
  /home/dolin/research/ducktape/build/lib/tests/tests/check_test.py
which is not the same as the test file we want to collect:
  /home/dolin/research/ducktape/tests/tests/check_test.py
HINT: remove __pycache__ / .pyc files and/or use a unique basename for your test file modules
_________________________________________________________________ ERROR collecting tests/utils/check_util.py _________________________________________________________________
import file mismatch:
imported module 'tests.utils.check_util' has this __file__ attribute:
  /home/dolin/research/ducktape/build/lib/tests/utils/check_util.py
which is not the same as the test file we want to collect:
  /home/dolin/research/ducktape/tests/utils/check_util.py
HINT: remove __pycache__ / .pyc files and/or use a unique basename for your test file modules
================================================================================== FAILURES ==================================================================================
_____________________________________________________________ CheckTemplateRenderingService.check_file_template ______________________________________________________________

self = <tests.templates.service.check_render.CheckTemplateRenderingService object at 0x7fba24426d50>

    def check_file_template(self):
>       self.new_instance().render_file_template()

build/lib/tests/templates/service/check_render.py:39: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 
build/lib/tests/templates/service/check_render.py:73: in render_file_template
    assert "Sample world" == self.render("sample")
build/bdist.linux-x86_64/egg/ducktape/template.py:90: in render
    template = self.template_env.get_template(path)
/export/apps/python/2.7.10/lib/python2.7/site-packages/Jinja2-2.8-py2.7.egg/jinja2/environment.py:812: in get_template
    return self._load_template(name, self.make_globals(globals))
/export/apps/python/2.7.10/lib/python2.7/site-packages/Jinja2-2.8-py2.7.egg/jinja2/environment.py:774: in _load_template
    cache_key = self.loader.get_source(self, name)[1]
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 

self = <jinja2.loaders.ChoiceLoader object at 0x7fba2443b1d0>, environment = <jinja2.environment.Environment object at 0x7fba2443b190>, template = 'sample'

    def get_source(self, environment, template):
        for loader in self.loaders:
            try:
                return loader.get_source(environment, template)
            except TemplateNotFound:
                pass
>       raise TemplateNotFound(template)
E       TemplateNotFound: sample

/export/apps/python/2.7.10/lib/python2.7/site-packages/Jinja2-2.8-py2.7.egg/jinja2/loaders.py:399: TemplateNotFound
_______________________________________________________________ CheckTemplateRenderingTest.check_file_template _______________________________________________________________

self = <tests.templates.test.check_render.CheckTemplateRenderingTest object at 0x7fba243d13d0>

    def check_file_template(self):
        test = self.setup()
        test.name = "world"
>       assert "Sample world" == test.render("sample")

build/lib/tests/templates/test/check_render.py:45: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 
build/bdist.linux-x86_64/egg/ducktape/template.py:90: in render
    template = self.template_env.get_template(path)
/export/apps/python/2.7.10/lib/python2.7/site-packages/Jinja2-2.8-py2.7.egg/jinja2/environment.py:812: in get_template
    return self._load_template(name, self.make_globals(globals))
/export/apps/python/2.7.10/lib/python2.7/site-packages/Jinja2-2.8-py2.7.egg/jinja2/environment.py:774: in _load_template
    cache_key = self.loader.get_source(self, name)[1]
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 

self = <jinja2.loaders.ChoiceLoader object at 0x7fba243d1e10>, environment = <jinja2.environment.Environment object at 0x7fba243d1dd0>, template = 'sample'

    def get_source(self, environment, template):
        for loader in self.loaders:
            try:
                return loader.get_source(environment, template)
            except TemplateNotFound:
                pass
>       raise TemplateNotFound(template)
E       TemplateNotFound: sample

/export/apps/python/2.7.10/lib/python2.7/site-packages/Jinja2-2.8-py2.7.egg/jinja2/loaders.py:399: TemplateNotFound
========================================================== 2 failed, 46 passed, 1 skipped, 30 error in 3.41 seconds ==========================================================

```

lindong28:
@granders Are you able to reproduce the error by doing "python setup.py test" on Ubuntu or Fedora? I see the same error when testing Ducktape on both platforms.

ewencp:
@lindong28 It looks like you have some stale files in your `build` directory. Not sure why those are causing problems. A `python setup.py clean` might clear it up, but it's generally safe to just `rm -r build` since that's just a temporary directory anyway.

I generally don't even have this directory since I rarely run the `build`/`install` commands for `setup.py`. I almost always just use `python setup.py develop` against my development checkout of ducktape.

lindong28:
@ewencp Thanks for your information. After removing the build folder "python setup.py test" succeeds on Fedora. But it still fails on Ubuntu with the follow error. I provide the error below for your information. But I should be unblocked to develop test on Fedora.

```
===================================================== test session starts =====================================================
platform linux2 -- Python 2.7.3 -- py-1.4.30 -- pytest-2.7.2
rootdir: /home/lindong/research/ducktape, inifile: setup.cfg
collected 55 items 

.eggs/pbr-1.8.1-py2.7.egg/pbr/tests/testpackage/test-requirements.txt .
tests/cluster/check_json.py .....
tests/cluster/check_localhost.py ..
tests/cluster/check_remoteaccount.py ..
tests/cluster/check_vagrant.py .
tests/command_line/check_main.py ..
tests/loader/check_loader.py ......
tests/mark/check_ignore.py ......
tests/mark/check_parametrize.py ........
tests/runner/check_serial_runner.py ..
tests/services/check_background_thread_service.py ..
tests/services/check_service.py .
tests/templates/service/check_render.py ....F
tests/templates/test/check_render.py .F.
tests/tests/check_session.py .
tests/tests/check_test.py ......
tests/utils/check_util.py ..

========================================================== FAILURES ===========================================================
______________________________________ CheckTemplateRenderingService.check_file_template ______________________________________

self = <tests.templates.service.check_render.CheckTemplateRenderingService object at 0x1deea50>

    def check_file_template(self):
>       self.new_instance().render_file_template()

tests/templates/service/check_render.py:39: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _
tests/templates/service/check_render.py:73: in render_file_template
    assert "Sample world" == self.render("sample")
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _

self = <TemplateRenderingService: num_nodes: 1, nodes: []>, path = 'sample', kwargs = {}
class_dir = '/home/lindong/research/ducktape/tests/templates/service', module_name = 'tests.templates.service.check_render'
package = 'tests', package_search_path = 'templates/service/templates'
loaders = [<jinja2.loaders.FileSystemLoader object at 0x1deee50>, <jinja2.loaders.PackageLoader object at 0x1deef10>], msg = ''

    def render(self, path, **kwargs):
        """
            Render a template loaded from a file.
            template files referenced in file f should be in a sibling directory of f called "templates".
    
            :param path: path, relative to the search paths, to the template file
            :param kwargs:
            :return:
            """
        if not hasattr(self, 'template_loader'):
            class_dir = os.path.dirname(inspect.getfile(self.__class__))
    
            module_name = self.__class__.__module__
            package, package_search_path = self._package_search_path(module_name)
    
            loaders = []
            msg = ""
            if os.path.isdir(class_dir):
                # FileSystemLoader overrides PackageLoader if the path containing this directory
                # is a valid directory. FileSystemLoader throws an error from which ChoiceLoader
                # doesn't recover if the directory is invalid
                loaders.append(FileSystemLoader(os.path.join(class_dir, 'templates')))
            else:
                msg += "Will not search in %s for template files since it is not a valid directory. " % class_dir
    
            if package_is_installed(package):
                loaders.append(PackageLoader(package, package_search_path))
            else:
                msg += "Will not search in package %s for template files because it cannot be imported."
    
            if len(loaders) == 0:
                # Expect at least one of FileSystemLoader and PackageLoader to be present
                raise EnvironmentError(msg)
    
            self.template_loader = ChoiceLoader(loaders)
>           self.template_env = Environment(loader=self.template_loader, trim_blocks=True, lstrip_blocks=True)
E           TypeError: __init__() got an unexpected keyword argument 'lstrip_blocks'

ducktape/template.py:88: TypeError
_______________________________________ CheckTemplateRenderingTest.check_file_template ________________________________________

self = <tests.templates.test.check_render.CheckTemplateRenderingTest object at 0x1df0390>

    def check_file_template(self):
        test = self.setup()
        test.name = "world"
>       assert "Sample world" == test.render("sample")

tests/templates/test/check_render.py:45: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _

self = <tests.templates.test.check_render.TemplateRenderingTest object at 0x1df0250>, path = 'sample', kwargs = {}
class_dir = '/home/lindong/research/ducktape/tests/templates/test', module_name = 'tests.templates.test.check_render'
package = 'tests', package_search_path = 'templates/test/templates'
loaders = [<jinja2.loaders.FileSystemLoader object at 0x1df02d0>, <jinja2.loaders.PackageLoader object at 0x1df0490>], msg = ''

    def render(self, path, **kwargs):
        """
            Render a template loaded from a file.
            template files referenced in file f should be in a sibling directory of f called "templates".
    
            :param path: path, relative to the search paths, to the template file
            :param kwargs:
            :return:
            """
        if not hasattr(self, 'template_loader'):
            class_dir = os.path.dirname(inspect.getfile(self.__class__))
    
            module_name = self.__class__.__module__
            package, package_search_path = self._package_search_path(module_name)
    
            loaders = []
            msg = ""
            if os.path.isdir(class_dir):
                # FileSystemLoader overrides PackageLoader if the path containing this directory
                # is a valid directory. FileSystemLoader throws an error from which ChoiceLoader
                # doesn't recover if the directory is invalid
                loaders.append(FileSystemLoader(os.path.join(class_dir, 'templates')))
            else:
                msg += "Will not search in %s for template files since it is not a valid directory. " % class_dir
    
            if package_is_installed(package):
                loaders.append(PackageLoader(package, package_search_path))
            else:
                msg += "Will not search in package %s for template files because it cannot be imported."
    
            if len(loaders) == 0:
                # Expect at least one of FileSystemLoader and PackageLoader to be present
                raise EnvironmentError(msg)
    
            self.template_loader = ChoiceLoader(loaders)
>           self.template_env = Environment(loader=self.template_loader, trim_blocks=True, lstrip_blocks=True)
E           TypeError: __init__() got an unexpected keyword argument 'lstrip_blocks'

ducktape/template.py:88: TypeError
============================================= 2 failed, 53 passed in 3.02 seconds =============================================

```

granders:
@lindong28 I'm speculating here, but some of the failures you've been seeing may have to do with the fact that we've included `__init__.py` files in our test directories, which I think means our unit tests get packaged and installed along with everything else. 

I wonder if you can also try to `pip uninstall ducktape` before running the tests.

lindong28:
@granders Sure. I just tried "python setup.py test" after doing "sudo pip uninstall". There are still the same error messages.

lindong28:
@granders I have added unit test for iter_wrapper. Can you have a look?

lindong28:
The test failed because `ssh local ... ` reports error `Host key verification failed`. Do you know why this would fail on the test machine?

I updated the patch so that if `host == localhost and user == None and ssh_args == None` Ducktape will not do `ssh local` to execute a command locally.

tests/cluster/check_remoteaccount.py
```diff
@@ -56,6 +55,39 @@ def stop(self):
             raise Exception("SimpleServer failed to stop quickly")
 
 
+
+class CheckIterWrapper(object):
+    def setup(self):
+        self.line_num = 6
+        self.account = MockAccount()
+        self.account.ssh_capture("mkdir -p /tmp").has_next()
```
granders: use `self.account.ssh` if you are not capturing output

tests/cluster/check_remoteaccount.py
```diff
@@ -56,6 +55,39 @@ def stop(self):
...
+        self.temp_file = "/tmp/ducktape-test-" + str(random.randint(0, 100000))
+        for i in range(self.line_num):
+            self.account.ssh_capture("echo " + str(i) + " >> " + self.temp_file).has_next()
```
granders: `self.account.ssh`

tests/cluster/check_remoteaccount.py
```diff
@@ -56,6 +55,39 @@ def stop(self):
...
+        for i in range(self.line_num):
+            assert output.has_next() == True
+            assert output.next().strip() == str(i)
+        start = time.time()
+        assert output.has_next(timeout_sec=5) == False
+        stop = time.time()
+        assert (stop - start >= 5) and (stop - start) < 6, "has_next() should return right after 5 seconds"
+
+    def teardown(self):
+        self.account.ssh_capture("rm -f " + self.temp_file)
```
granders: `self.account.ssh`

tests/cluster/check_remoteaccount.py
```diff
@@ -56,6 +55,39 @@ def stop(self):
...
+
+    def check_iter_wrapper(self):
+        output = self.account.ssh_capture("tail " + self.temp_file)
+        for i in range(self.line_num):
+            assert output.has_next() == True
+            assert output.next().strip() == str(i)
+        start = time.time()
+        assert output.has_next() == False
+        stop = time.time()
+        assert stop - start < 1, "has_next() should return immediately"
```
granders: `time.time()` is in units of seconds, so maybe `stop - start < .001`?

tests/cluster/check_remoteaccount.py
```diff
@@ -56,6 +55,39 @@ def stop(self):
...
+        self.account = MockAccount()
+        self.account.ssh_capture("mkdir -p /tmp").has_next()
+        self.temp_file = "/tmp/ducktape-test-" + str(random.randint(0, 100000))
+        for i in range(self.line_num):
+            self.account.ssh_capture("echo " + str(i) + " >> " + self.temp_file).has_next()
+
+    def check_iter_wrapper(self):
+        output = self.account.ssh_capture("tail " + self.temp_file)
+        for i in range(self.line_num):
+            assert output.has_next() == True
```
granders: Not a big deal, but I personally prefer `assert output.has_next()`

tests/cluster/check_remoteaccount.py
```diff
@@ -56,6 +55,39 @@ def stop(self):
...
+            assert output.next().strip() == str(i)
+        start = time.time()
+        assert output.has_next() == False
+        stop = time.time()
+        assert stop - start < 1, "has_next() should return immediately"
+
+    def check_iter_wrapper_timeout(self):
+        output = self.account.ssh_capture("tail -F " + self.temp_file)
+        for i in range(self.line_num):
+            assert output.has_next() == True
```
granders: Since you're checking timeout in this test, maybe use something like
`assert output.has_next(timeout_sec=.1)`

tests/cluster/check_remoteaccount.py
```diff
@@ -56,6 +55,39 @@ def stop(self):
...
+            assert output.next().strip() == str(i)
+        start = time.time()
+        assert output.has_next(timeout_sec=5) == False
+        stop = time.time()
+        assert (stop - start >= 5) and (stop - start) < 6, "has_next() should return right after 5 seconds"
```
granders: Similar to above - you could probably safely tighten the bound of this check (closer to 5)

ducktape/cluster/remoteaccount.py
```diff
@@ -59,6 +60,8 @@ def _can_ping_url(self, url, headers):
             return False
 
     def ssh_command(self, cmd):
+        if not self.user and self.hostname == 'localhost' and not self.ssh_args:
```
granders: Can use `if self.local: ...`

lindong28:
@granders Thanks for your detailed review! I have updated the patch as you suggested.

granders:
Thanks @lindong28, couple last comments and this should be good.

I neglected this before: although both work, it's better to change `x == None` everywhere to `x is None`
see, e.g. http://jaredgrubb.blogspot.com/2009/04/python-is-none-vs-none.html

Also, can you document in the `iter_wrapper` how `timeout_sec` is treated (block for a maximum of `timeout_sec` to see if output is available), and also the fact that `has_next()` when called with no arguments will block until output is available?

lindong28:
@granders Thanks for the review! Please check the updated patch.

granders:
Thanks @lindong28, LGTM

PR #111, opened 2015-10-15
WIP: Detect regression

granders:
@ewencp 
This is a semi-functional prototype - interested in initial high level feedback
Definitely some cruft here (and a couple failing unit tests) but some of the features here are:

- Persistent storage for test result metadata
- `@regression` decorator, and corresponding `DefaultRegressionTest`
- Mostly decoupled reporting from a particular test run. One can generate a report (almost) from a particular session id purely by grabbing historical data.
- Crude visibility into historical data in html report ()


ducktape/command_line/main.py
```diff
@@ -141,6 +157,10 @@ def main():
             print "parameters are not valid json: " + str(e.message)
             sys.exit(1)
 
+    if args.report_only:
+        generate_reports(args.session_id, args.results_root)
```
ewencp: `--session-id` isn't required, you might need to check that it was set / is valid.

Also, it doesn't look like this matches the argument list. I think you're missing the `result_store`, which isn't generated until later in this function.

ducktape/command_line/main.py
```diff
@@ -184,19 +204,14 @@ def main():
 
     # Run the tests
     runner = SerialTestRunner(session_context, tests)
-    test_results = runner.run_all_tests()
+    runner.run_all_tests()  # populates a thing with test results
```
ewencp: Haha, not sure this is the most useful comment...

ducktape/command_line/main.py
```diff
@@ -184,19 +204,14 @@ def main():
...
-    reporter = HTMLSummaryReporter(test_results)
-    reporter.report()
 
+    result_store = FileSystemResultStore(args.results_root)
+    generate_reports(session_context.session_id, result_store, results_dir)
     update_latest_symlink(args.results_root, results_dir)
-    if not test_results.get_aggregate_success():
-        sys.exit(1)
+
+    # if not test_results.get_aggregate_success():
```
ewencp: Is this just because this is WIP and commented out for testing? Fine if it is, but just want to make sure we don't accidentally leave this commented out.

ducktape/mark/_parametrize.py
```diff
@@ -22,6 +22,11 @@
 
 def mark_as(fun, mark):
     """Attach a tag indicating that fun has been marked with the given mark"""
+
+    if isinstance(fun, TestGenerator):
```
ewencp: I'm not clear why we have this special case. There's already a check below for if the function has a `marks` attribute. Why the special case here? Is this because you want the `marks` `@property` to include the wrapped object's marks as well?

`isinstance` has its uses, but it's pretty unpythonic. I'm wondering if there's a better solution here. Normally I'd suggest using the `@marks.setter` annotation to add a setter in `TestGenerator`, but that won't work here since it's a collection that you want to add to instead of replace.

ducktape/mark/_regression.py
```diff
@@ -0,0 +1,41 @@
...
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+from ducktape.mark._parametrize import mark_as
+
+
+def has_regression(fun):
```
ewencp: Name here is misleading/confusing. `has_regression_mark`?

ducktape/mark/_regression.py
```diff
@@ -0,0 +1,41 @@
...
+    if hasattr(fun, "marks"):
+        for mark in fun.marks:
+            if isinstance(mark, RegressionMark):
+                return True
+    return False
+
+
+class Mark(object):
```
ewencp: Why does this class exist? Why doesn't `RegressionMark` inherit directly from `object`?

ducktape/tests/loader.py
```diff
@@ -32,11 +34,22 @@ class LoaderException(Exception):
 
 class TestInfo(object):
     """Helper class used to wrap discovered test information"""
-    def __init__(self, module=None, cls=None, function=None, injected_args=None):
+
+    @staticmethod
+    def from_test_info(t_info):
```
ewencp: These seems more intuitive as a non-static `clone` method.

ducktape/tests/loader.py
```diff
@@ -32,11 +34,22 @@ class LoaderException(Exception):
...
+        """Create a copy"""
+        return TestInfo(t_info.module, t_info.cls, t_info.function, t_info.injected_args, t_info.regression_variable_selector)
+
+    def __init__(self, module=None, cls=None, function=None, injected_args=None, regression_variable_selector=None):
```
ewencp: Just a thought that I haven't looked thoroughly into the implications of, but if we're going to end up with "fake" tests like these ones for regressions, it might make sense to make some of them subclasses of `TestInfo` instead of continuing to add parameters to the constructor here with a bunch of conditionals in the class. This works ok for now as long as we only have a few ways to customize it (injected args, regression_variable_selector), but it gets messy when scaled up.

ducktape/tests/loader.py
```diff
@@ -138,11 +152,20 @@ def discover(self, test_discovery_symbols):
...
                 raise LoaderException("Didn't find any tests for symbol %s." % symbol)
 
+        # Assemble the test context objects
+        test_context_list = []
+        for t in test_info_list:
+            if t.regression_variable_selector is None:
+                test_context_list.append(
+                    TestContext(self.session_context, t.module_name, t.cls, t.function, t.injected_args))
+            else:
+                target_test_context = TestContext(self.session_context, t.module_name, t.cls, t.function, t.injected_args)
```
ewencp: The `TestContext` being constructed in both branches of this conditional are the same and can be lifted outside it. The only difference is whether you wrap it in the `DefaultRegressionTest.create_test_context`.

ducktape/tests/loader.py
```diff
@@ -177,15 +200,12 @@ def import_modules(self, file_list):
...
-                    # error. This does mean we fail to log a subset of import
-                    # errors that we should: if the module is valid but itself
-                    # triggers an ImportError (e.g. typo in an import line),
-                    # that error will be swallowed.
-                    if not isinstance(e, ImportError):
+                    # TypeError, etc)
+                    #
+                    # Log all errors except those caused by modules which don't exist
+
+                    module_does_not_exist = isinstance(e, ImportError) and re.search("No module named", str(e))
```
ewencp: omg, this is a wonderful change! Tracking down `ImportError`s due to bugs in code rather than missing modules was pretty painful before.

ducktape/tests/loader.py
```diff
@@ -230,10 +250,24 @@ def expand_class(self, t_info):
...
+        test_info_list.extend(self.expand_regressions(test_info_list))
         return test_info_list
-
-    def expand_function(self, t_info):
+    
+    def expand_regressions(self, t_info_list):
+        regression_list = []
+        
+        for t_info in t_info_list:
+            if has_regression(t_info.function):
```
ewencp: Hmm, given the way this is used, it might be cleaner to provide a `regressions(t_info.function)` method that is a generator of `RegressionMark` annotations. As it is now, you scan through the list to figure out if there is a regression (`has_regression`), then scan through again doing the same check that `has_regression` does to filter down to only the `RegressionMark`s.

ducktape/tests/loader.py
```diff
@@ -230,10 +250,24 @@ def expand_class(self, t_info):
...
+                for mark in t_info.function.marks:
+                    if isinstance(mark, RegressionMark):
+                        t_copy = TestInfo.from_test_info(t_info)
+                        t_copy.regression_variable_selector = mark.variable_selector
```
ewencp: The clone method could also just accept additional keyword arguments to provide overrides. That would turn this into a one-liner like

```
regression_list.append(t_info.clone(regression_variable_selector=mark.variable_selector))
```

ducktape/tests/regression_test.py
```diff
@@ -0,0 +1,64 @@
...
+import statistics
+
+
+class RegressionTest(Test):
+    pass
+
+
+class DefaultRegressionTest(RegressionTest):
+    """
+    Only run if some test with matching TestKey gets run
```
ewencp: This could use more detail about what the default regression test actually does, i.e. when will it fail, when will it succeed?

ducktape/tests/regression_test.py
```diff
@@ -0,0 +1,64 @@
...
+        injected_args = {"target_test_id": target_test_context.test_id, "variable_selector": variable_selector}
+        test_function =_inject(**injected_args)(cls.test)
+
+        return TestContext(session_context, module=DefaultRegressionTest.__module__, cls=cls, function=test_function,
+                           injected_args=injected_args)
+
+    @staticmethod
+    def fetch_variable(variable_selector, datum):
+        value = datum
+        for selector in variable_selector:
```
ewencp: Maybe examples are all I needed, but it took me until here to figure out that `variable_selector` is a list of field names to be applied hierarchically...

ducktape/tests/regression_test.py
```diff
@@ -0,0 +1,64 @@
...
+            value = value[selector]
+            if value is None:
```
ewencp: Hmm, not sure if this escape hatch is a nice convenience or likely to hide coding mistakes...

ducktape/tests/regression_test.py
```diff
@@ -0,0 +1,64 @@
...
+    def __init__(self, test_context, *args, **kwargs):
+        super(DefaultRegressionTest, self).__init__(test_context, *args, **kwargs)
+
+    def test(self, target_test_id, variable_selector):
+        # get data for up to 20? previous runs
+        result_store = self.test_context.session_context.result_store
+        test_data = result_store.test_data(target_test_id)
+
+        test_data = [d for d in test_data if d["status"].lower() == "pass" and d is not None]
+        test_data = [DefaultRegressionTest.fetch_variable(variable_selector, datum) for datum in test_data]
```
ewencp: Multiple passes seem unnecessary, just combine the first part of the second list comprehension with the remainder of the first.

ducktape/tests/regression_test.py
```diff
@@ -0,0 +1,64 @@
...
+
+        num_data_points = min(20, len(test_data))
```
ewencp: Looks like the comment above might be addressing this, but filtering at the result_store when loading the data is better. This approach clearly isn't going to handle a larger amount of history well. I realize we need to be able to grab more than 20 items due to the filtering to executions that passed, but we shouldn't need all of the history. Might be fine for now though, or even if the result_store just capped the total amount of data that will ever be returned.

On a related note, it'd be nice if this was configurable, but I think we can avoid that in an initial patch. It can clearly be added in the future and this is just the default anyway.

ducktape/tests/regression_test.py
```diff
@@ -0,0 +1,64 @@
...
+        test_data = test_data[-num_data_points:]
+        return test_data
```
ewencp: Is there a real assertion somewhere? What triggers a failure for a DefaultRegressionTest?

ducktape/tests/reporter.py
```diff
@@ -45,25 +46,25 @@ def format_time(t):
...
-            "status:     %s" % pass_fail(self.result.success),
-            "run time:   %s" % format_time(self.result.run_time),
+            "test_id:    %s" % self.datum["test_id"],
+            "status:     %s" % self.datum["status"].upper(),
+            "run time:   %s" % format_time(self.datum["run_time_sec"]),
         ]
 
-        if not self.result.success:
-           # Add summary if the test failed
+        if self.datum["status"].lower() != "pass" and self.datum["error_msg"] is not None:
```
ewencp: Not a big deal, but it looks like we're breaking with previous result data since some of the standard output fields differ? We don't have any way to reload that data so it doesn't matter I guess?

ducktape/tests/reporter.py
```diff
@@ -79,22 +80,26 @@ def report(self):
...
+        self.results_dir = results_dir
+
     def report(self):
         self.width = DEFAULT_SEPARATOR_WIDTH
-        report_file = os.path.join(self.result.test_context.results_dir, "report.txt")
+        report_file = os.path.join(self.results_dir, "report.txt")
         with open(report_file, "w") as fp:
             fp.write(self.report_string())
 
-        # write collected data
```
ewencp: I don't particularly like the extra clutter, but is there any chance someone will miss having the output JSON readily available in the output directory? Is it easy enough to get to otherwise, either via the `ResultStore` or in some log, that it won't be missed?

ducktape/tests/reporter.py
```diff
@@ -142,75 +152,77 @@ def report(self):
...
+    def __init__(self, session_id, result_store, results_dir):
+        super(HTMLSummaryReporter, self).__init__(session_id, result_store)
+        self.results_dir = results_dir
 
-    def format_test_name(self, result):
-        lines = ["Module: " + result.test_context.module_name,
-                 "Class:  " + result.test_context.cls_name,
-                 "Method: " + result.test_context.function_name]
+    def format_test_name(self, datum):
+        lines = ["Module: " + datum["module_name"],
```
ewencp: Haven't seen where this all gets included in the test "datum" yet, but it seems like there are quite a few things at the top level. Should we group all of this in a sub-object with a name that is very unlikely to conflict with anything reported by the tests into the result object?

ducktape/tests/reporter.py
```diff
@@ -142,75 +152,77 @@ def report(self):
...
 
         Path is relative to the base results_dir. Relative path behaves better if the results directory is copied,
         moved etc.
         """
-        base_dir = os.path.abspath(result.session_context.results_dir)
-        base_dir = os.path.join(base_dir, "")  # Ensure trailing directory indicator
-
-        test_results_dir = os.path.abspath(result.test_context.results_dir)
-        return test_results_dir[len(base_dir):]  # truncate the "absolute" portion
+        # base_dir = os.path.abspath(self.session_results_dir)
```
ewencp: More commented stuff that we should either remove or uncomment.

ducktape/tests/result_store.py
```diff
@@ -0,0 +1,175 @@
...
+        "test_id": result.test_context.test_id,
+        "module_name": result.test_context.module_name,
+        "cls_name": result.test_context.cls_name,
+        "function_name": result.test_context.function_name,
+        "injected_args": result.test_context.injected_args,
+        "run_time_sec": result.run_time,
+        "services": result.test_context.services,
+        "status": "pass" if result.success else "fail",
+        "error_msg": result.summary,
+        "data": result.data
```
ewencp: Ah, I think I see now. Instead of moving all the test run info to a sub-object, you've kept all the test-specific result data in a sub-object. That works too :)

ducktape/tests/result_store.py
```diff
@@ -0,0 +1,175 @@
...
+    }
+
+
+def create_session_datum(session_datum):
+    return session_datum
```
ewencp: This seems not very useful?

ducktape/tests/result_store.py
```diff
@@ -0,0 +1,175 @@
...
+                                datum["injected_args"])
+
+    def __init__(self, module_name, cls_name, function_name, injected_args={}):
+        self.module_name = module_name
+        self.cls_name = cls_name
+        self.function_name = function_name
+        self.injected_args = injected_args
+
+    def __repr__(self):
+        return "~~".join(
```
ewencp: This seems like a weird implementation of `__repr__`. Is this actually valid? Normally you want `__repr__` to be a) exhaustive and b) if possible, have `eval(repr(foo)) == foo`. Is there a reason we've implemented `__repr__` instead of, e.g. `__str__` or `__unicode__`?

ducktape/tests/result_store.py
```diff
@@ -0,0 +1,175 @@
...
+    DB_NAME = "test_data.db"
+
+    def __init__(self, root_results_dir):
+        """Assume key is a string representing a path relative to root result directory"""
+        self.root_results_dir = root_results_dir
+        self.db_file = os.path.join(self.root_results_dir, "test_data.db")
+        self._bootstrap_store()
+
+    def _bootstrap_store(self):
+        with sqlite3.connect(self.db_file) as connection:
```
ewencp: Wow, a real DB! I though for sure this was going to be a bunch of pickled objects.

ducktape/tests/result_store.py
```diff
@@ -0,0 +1,175 @@
...
+
+            cur.execute("""create table if not exists session_data (
+              id integer primary key autoincrement,
+              timestamp integer not null,
+              session_id text not null,
+              data text not null
+            );""")
+
+
+    def _serialize(self, datum):
```
ewencp: `_serialize` and `_deserialize` could be made `@staticmethod`s

ducktape/tests/result_store.py
```diff
@@ -0,0 +1,175 @@
...
+        return json.dumps(datum, separators=(',', ':'))
+
+    def _deserialize(self, datum_bytes):
+        return json.loads(datum_bytes)
+
+    def put(self, session_id, test_id, datum):
+        """"""
+        with sqlite3.connect(self.db_file) as connection:
```
ewencp: Would it be better to get the connection once and then just do:

```
with self.connection as connection:
```

to take advantage of the context manager transaction rollback? In fact, `self.connection` could just cache the connection so the `sqlite3.connect` is implicit with the first access.

ducktape/tests/runner.py
```diff
@@ -58,9 +61,15 @@ def __init__(self, *args, **kwargs):
         self.current_test = None
         self.current_test_context = None
 
+        # Move regression tests to the end of the queue
+        regression_tests = [t for t in self.tests if isinstance(t.cls, RegressionTest)]
```
ewencp: Long term, I would like to avoid customized support in the runner like this. That said, I think since this is completely internal right now, we can clean this up later. We'll eventually want to be able to express test dependencies, and when we can, the test loader can setup the dependency so that the test runner will handle it generically, rather than this customized approach for `RegressionTest` we're using now. Might want to file an issue so we don't lose track of this.

ducktape/tests/test.py
```diff
@@ -135,6 +136,22 @@ def _escape_pathname(s):
 
 class TestContext(Logger):
     """Wrapper class for state variables needed to properly run a single 'test unit'."""
+    _TEST_ID_DELIMITER = "~~"
```
ewencp: Now I'm recognizing this from the earlier `__repr__` question. Why the magic delimter?

tests/mark/check_regression_decorator.py
```diff
@@ -0,0 +1,185 @@
+# Copyright 2015 Confluent Inc.
```
ewencp: Is this just a copy of the parametrization tests? Doesn't seem to add any checks for the `@regression` decorator.

ewencp:
Something else to think about: how does this interact with something like a PR builder? Do we need to be able to just toggle off regression tests entirely for a PR builder, where things are expected to fail, encounter experimental changes, etc? Or is that dangerous since you won't see the regression until post-commit on the normal builder? I'm not sure of the answer here, but it could be relevant to the kafka_system_tests_branch_builder soon after regression system tests start being added to Kafka...

In terms of how we plan for short term regression tests needs vs the ideal long term design, I think this does a good job of hiding a lot of the details from the user. In particular, since we're only using a default regression test right now, there's very little impact on the public API. We could change a lot of the implementation, as well as add features, and not affect existing tests at all. So I think that's a good position to be in. I left a bunch of comments, but they are all details -- the general patch looks good.

Issue #112, opened 2015-10-15
add ignore or ok_to_fail annotation


granders:
When tests are failing repeatedly, we can temporarily mark with `@ok_to_fail` or `@ignore` to avoid excessive noise for those paying attention to the test runs.

`@ignore` temporarily disables a test or group of tests
Ignored tests are discovered, but not run. The test report(s) should include a summary of ignored tests.

nehanarkhede:
Nice to see this :)

granders:
:)

granders:
fixed via https://github.com/confluentinc/ducktape/pull/115

PR #115, opened 2015-10-19
Ignore decorator

granders:
@ewencp 
Added `@ignore` decorator, addresses https://github.com/confluentinc/ducktape/issues/112
- Used without parameters (`@ignore`), all parametrizations of the marked function will be ignored.
- Used with parameters (`@ignore(x=1, y=2)`), only the specific parametrization of the marked function will be ignored.
- Ignored test functions are acknowledged in the test report, but not run by the test runner.

Other notes:
- Re-implemented the parametrization decorators in a way that is simpler, and that I think makes it much easier to add new marks
- The only user facing consequence is that a `@matrix` decorator no longer longer "multiplies" the parametrize or matrix decorators below it, but I've been skeptical of this for a while, and have seen near 0 usage of matrix in this way.
- Moved the work of expanding a marked function into `TestContext` objects into the `mark` module, and in the process ditched `TestInfo` since it is largely identical to `TestContext`.
- This necessitated updating `TestContext` so that the logger is lazily configured - i.e., not until its first use. Before, the act of instantiating a `TestContext` object had the side effect of creating directories... :/
- Removed the British spelling for `parametrize`. It just seems too confusing to make it possible to have multiple spellings in a code base.


ducktape/mark/_mark.py
```diff
@@ -0,0 +1,359 @@
...
+        """Attach a tag indicating that fun has been marked with the given mark
+
+        Marking fun updates it with two attributes:
+
+        - marks:      a list of mark objects applied to the function. These may be strings or objects subclassing Mark
+                      we use a list because in some cases, it is useful to preserve ordering.
+        - mark_names: a set of names of marks applied to the function
+        """
+        # Update fun.marks
+        if hasattr(fun, "marks"):
```
ewencp: Took me a minute to figure out what was going on here because we're mixing the `hasattr`/`getattr`/`setattr` with the `__dict__` style access. Any reason for this?

ducktape/mark/_mark.py
```diff
@@ -0,0 +1,359 @@
...
+
+        If parameters are supplied to the @ignore decorator, only ignore the parametrization with matching parameter(s)
+
+        @ignore(x=2, y=3)
+        @parametrize(x=1, y=0)  # This test will run as usual
+        @parametrize(x=2, y=3)  # This test will be ignored
+        def the_test(...):
+            ...
+    """
+    if len(args) == 1 and len(kwargs) == 0:
```
ewencp: Isn't the wrapped function always going to be the first arg, and must be there anyway? So maybe this can be made clearer if the method is defined as

```
def ignore(f, *args, **kwargs)
```

Additionally, do we want the `*args` part, or should we only have `**kwargs`? It looks like the positional arguments are ignored below (the `ignorer` nested function doesn't use them), but by including `*args`, we allow extra positional arguments to be provided.

ducktape/mark/_mark.py
```diff
@@ -0,0 +1,359 @@
...
+
+        expanded = []
+        parametrize_marks = [m for m in f.marks if m.name == PARAMETRIZED.name]
+        matrix_marks = [m for m in f.marks if m.name == MATRIX.name]
+        ignore_marks = [m for m in f.marks if m.name == IGNORE.name]
+
+        ignore_list = [m.injected_args for m in ignore_marks]
+        ignore_all = None in ignore_list
+
+        if test_parameters is not None:
```
ewencp: This seems like it's going to get very messy over time as we have more annotations. We've basically centralized the entire process of expanding the set of tests into one class/method that has to understand everything that all annotations do.

Is there a reason we can't keep each of these blocks in their respective `Mark` classes and only have one smaller method here that manages the high-level process of generating the initial singular `TestContext` for the unmarked test function, then dispatching a `list` of `TestContext`'s to each `Mark`, which generates an output `list` of `TestContext`'s that it derives from the input list?

It seems like this should compose well, cleanly separate each annotation, and discourages dependencies between annotations that cannot be expressed simply via a `TestContext` -- basically, if different annotations want to interact, they have to agree to a clear contract for how to exchange information about what they are doing via `TestContext` attributes alone (which, of course, are extensible by adding attributes, as is done here with, e.g., `ignore`).

In theory this also provides a cleaner mechanism for extending the system with project-specific annotations, though clearly you also need the right hooks into other parts of the system for that to work (e.g. you couldn't support a user-defined `ignore` annotation without also providing a way to hook into the runner/reporter to actual implement the part that checks the `ignore` flag and takes the right action).

ducktape/mark/_mark.py
```diff
@@ -0,0 +1,359 @@
...
+            context = TestContext(self.session_context, self.module, self.cls, self.function)
+            context.ignore = ignore_all
+            return [context]
+
+        # expand parametrize marks, adding "ignore" metadata where necessary
+        for m in parametrize_marks:
+            injected_fun = _inject(**m.injected_args)(self.function)
+
+            context = TestContext(self.session_context, self.module, self.cls, injected_fun, m.injected_args)
+            context.ignore = None in ignore_list or m.injected_args in ignore_list
```
ewencp: I think these pieces overriding the `ignore` flag are a great example of how this is going to get really messy if we don't set these annotations up to compose well from the get go.

I'd suggest changing all of these to make use of a `TestContext` constructor that accepts `**kwargs` to be able to copy the entire original dict, with updates applied to keys that you know need to be overridden (the function and args in this case). This ensures all existing properties will be propagated, which makes a lot of cases work out naturally when composed (e.g., even if you put `@ignore()` first, the flag will be propagated to all `TestContext`s even if they are further expanded; if you included arguments, the flag + arguments can be copied to all `TestContext`s such that the ignore flag is still only applied to `TestContext`s with matching keyword args).

ducktape/tests/runner.py
```diff
@@ -73,10 +73,23 @@ def run_all_tests(self):
 
             # Create single testable unit and corresponding test result object
             self.current_test_context = test_context
+            result = TestResult(self.current_test_context)
+
+            if self.current_test_context.ignore:
+                # Skip running this test, but keep track of the fact that we ignored it
```
ewencp: Interesting, I was taking `ignore` to mean that you should run the test, but because we know it might fail you don't want it to mark the entire test run as having failed. This approach could work too, but it also means that any transient test failures that may be hard to reproduce can't get additional logging added, which could be valuable since it allows you to keep collecting info from normal test executions without having to pull up your own test infrastructure.

Not sure which is more useful.

ducktape/mark/_mark.py
granders: For reasons I don't fully understand yet, setting attributes on an instancemethod is finicky and will cause errors if one tries to use `setattr(...)`, but updating the `__dict__` works

ducktape/mark/_mark.py
granders: I think this is right, but see if this makes sense to you. Maybe there's a better way to do this?

The odd structure here is due to the tricky usage both with and without parameters:
```
# In this case we have essentially ignore(f)
@ignore
def f():
  ...

# In this case, we have ignore(*args, **kwargs)(f)
@ignore(x=1, y=2)
def f(x, y):
  ...
```

ducktape/mark/_mark.py
granders: added copy constructor with this behavior

ducktape/mark/_mark.py
granders: Yeah... I see your point.
I made an along these lines, although it required a bit more complexity than only propagating lists of `TestContext` objects.

ducktape/tests/runner.py
granders: I matched the `@Ignore` behavior in JUnit I think. 

Actually running the tests seemed unintuitive to me, but I guess we can see how we like this behavior in practice.

With the metadata in place, modifying the runner behavior is relatively simple.

granders:
@ewencp 
I made the mark behavior more modular, although it required keeping track of a bit more state than just the list of context objects. Let me know if you have better ideas here...

ducktape/mark/_mark.py
ewencp: You're seeing the effect of the hooks that Python provides for intercepting `getattr` and `setattr` by overriding special methods on the object: https://docs.python.org/2/reference/datamodel.html#object.__getattr__

This is partially why I'm asking -- the two paths you're using have differences, so you could end up with inconsistencies. I think the way you're using it here ends up fine since `__getattr__` is only called if the attribute wasn't found via normal lookups, which will include checking the `__dict__`, but the inconsistency is odd since you can also read from `__dict__`.

ducktape/mark/_mark.py
ewencp: I see, that makes sense.

ducktape/tests/runner.py
ewencp: Yeah, the reason I mentioned this is because we actually just ran into a case sort of like this today. The test is failing and we have a proposed fix which we're pretty confident of, but it wouldn't hurt to ignore the results for a few days to make sure it really is fixed.

In this case, the bug was tracked down without too much trouble, but for a transient issue that only occurs occasionally, it's useful to be able to add debugging code or settings specifically in the test (or, e.g., more verbose logging) and still run the tests.

I think they both potentially make sense. With unit tests, you usually end up with fewer of these transient errors (at least in most projects...), so that mode isn't as important. With system tests, you often need to collect enough evidence to figure out what's going wrong. We can leave this as is and come up with another decorator or a flag for this one if we want to support both approaches.

ducktape/mark/_mark.py
```diff
@@ -0,0 +1,374 @@
...
+    """Ignore a specific parametrization of test."""
+    def __init__(self, **kwargs):
+        # Ignore tests with injected_args matching self.injected_args
+        self.injected_args = kwargs
+
+    @property
+    def name(self):
+        return "IGNORE"
+
+    def apply(self, seed_context, context_list, maps):
```
ewencp: Why jump through all these hoops with this `apply` method with `maps` parameter? Why wouldn't something like

```
def apply(self, contexts):
    return ctx.clone(ignore=(self.injected_args is None or self.injected_args == ctx.injected_args)) for ctx in contexts
```

work? I was thinking that every `Mark` would take a list/generator of `TestContext`s and output an updated lists. `@ignore` returns the same number, just tagged as necessary; `@parametrize` just needs to add one new entry to the end; `@matrix` does the same, etc.

If I'm reading this correctly, you've done something similar, but where you collect all the mapping functions first, then apply them all in the base class's `apply`, which makes it a lot harder to understand what's going on.

ducktape/templates/report/report.html
```diff
@@ -62,7 +70,7 @@
       
       var TestRow = React.createClass({
         render: function() {
-          var className = this.props.test.test_result + "Class"
+          var className = this.props.test.test_result
```
ewencp: Nit: The `Detail` column doesn't point to anything valid when a test has been ignored.

ducktape/tests/test.py
```diff
@@ -18,6 +18,7 @@
 from ducktape.services.service_registry import ServiceRegistry
 from ducktape.template import TemplateRenderer
 
+import copy
```
ewencp: Stray import? Looks like you didn't end up using this module at all.

ducktape/tests/test.py
```diff
@@ -143,6 +144,7 @@ def __init__(self, session_context, module=None, cls=None, function=None, inject
         self.cls = cls
         self.function = function
         self.injected_args = injected_args
+        self.ignore = ignore  # Don't run this test if ignore is True
```
ewencp: I'd keep this class as minimal as possible and just allow annotations to add attributes as they need to. This does require a bit more care when using the attributes, but decouples the details of the implementation of the annotations from the core TestContext object.

The same could probably also hold for `injected_args`. I don't think it *needs* to always be set on this object. The major difference between the two, however, which would make removing `injected_args` a more significant change, is that `injected_args` is used to generate a bunch of derived info like the output directory and full test name.

ducktape/tests/test.py
```diff
@@ -150,16 +152,15 @@ def __init__(self, session_context, module=None, cls=None, function=None, inject
...
-
-        self._logger_configured = False
-        self.configure_logger()
-
     def __repr__(self):
         return "<module=%s, cls=%s, function=%s, injected_args=%s>" % \
                (self.module, self.cls_name, self.function_name, str(self.injected_args))
 
+    def copy(self, **kwargs):
+        ctx_copy = TestContext(self.session_context, self.module, self.cls, self.function, self.injected_args, self.ignore)
```
ewencp: You should be able to avoid always having to keep this up to date with the constructor args by just using the fact that you can get the `__dict__`. You could either change all arguments in the constructor to be keyword arguments and then do something like:

`ctx_copy = TestContext(**self.__dict__)`

In fact, I think that works even if you leave `session_context` as is. Regardless, my previous comment about not including `ignore` on every `TestContext` objects would help here by keeping the number of arguments on this line fixed. As it stands, as we add more annotations this list is just going to get ridiculously unwieldy.

tests/runner/check_serial_runner.py
```diff
@@ -51,8 +52,9 @@ def check_simple_run(self):
 
         results = runner.run_all_tests()
         assert len(results) == 1
-        assert results.num_failed() == 0
-        assert results.num_passed() == 1
+        assert results.num_failed == 0
+        assert results.num_passed == 1
+        assert results.num_ignored == 0
```
ewencp: Do you want a test that checks that the serial test runner handles ignored tests properly (i.e. validates the test isn't invoked) and gives the right count?

ducktape/mark/_mark.py
granders: Yeah, I agree, passing and returning `maps` is more confusing.

I wanted the effects of an `@ignore` decorator to be the same regardless of where it is applied relative to other `mark` decorators that may be present.

The two choices I had to achieve this were a) treat ignore decorators differently in the "expander" class by applying them last, or b) something akin to this `map` approach where I know any new context will be transformed regardless of the order of application.

ducktape/tests/test.py
granders: deleted

ducktape/mark/_mark.py
ewencp: I think I see what you're getting at now. Isn't the issue then that you're trying to resolve the `ignore` flag too early? Another way of structuring this is to *always* annotate the `TestContext` object with the relevant info, i.e. 

```
def apply(self, contexts):
    return ctx.clone(ignore=True, self.ignore_args=self.injected_args)
```

And then defer evaluation of whether to *actually* ignore the test to the test runner. That said, I actually don't think requiring the ordering to be a certain way is that big a deal. This method could always detect conditions which should be considered errors (e.g., arguments are specified for the `@ignore` but none of the `TestContext`s passed in match).

granders:
@ewencp Ready for another look.

Note - I opted to keep `ignore` field in `TestContext` for now. If we see an explosion of marks, we can think about packaging that metadata more cleanly.

ewencp:
This LGTM. We discussed some possible improvements to further isolate the implementations of different test method decorators. It'd be nice to further clean up the implementation, but since it's all internal and clean enough as is, feel free to merge.

PR #121, opened 2015-10-23
Fix testcontext copy

granders:
@ewencp Hopefully a quick review
Fixes a bug which cropped up in 0.3.7

Use of the `TestContext` "copy constructor" when expanding parametrized tests resulted in all test context objects sharing a reference to the same `ServiceRegistry` object.

As a result, the number of registered services grew from test to test, causing `min_cluster_size` to balloon, causing the test runner to think the cluster was too small to run the test and all tests to fail from then on.


tests/tests/check_test_context.py
```diff
@@ -0,0 +1,54 @@
...
+
+        This problem cropped up in particular with parametrized tests.
+        """
+        expander = MarkedFunctionExpander(session_context=session_context(), cls=DummyTest, function=DummyTest.test_me)
+        ctx_list = expander.expand()
+
+        for ctx in ctx_list:
+            # Constructing an instance of the test class causes a service to be registered with the test context
+            ctx.cls(ctx)
+            assert len(ctx.services) == 1
```
ewencp: This works fine. You could also directly validate that none of the `TestContext`'s share objects that they shouldn't -- something like `assert len(set(id(x.services) for x in ctx_lists)) == len(ctx_lists)`

ducktape/tests/test.py
```diff
@@ -151,16 +150,19 @@ def __init__(self, **kwargs):
...
         # dict for toggling service log collection on/off
-        self.log_collect = kwargs.get("log_collect", {})
+        self.log_collect = {}
 
     def __repr__(self):
         return "<module=%s, cls=%s, function=%s, injected_args=%s>" % \
                (self.module, self.cls_name, self.function_name, str(self.injected_args))
 
-    def copy(self, **kwargs):
+    def copy_constructor(self, **kwargs):
```
ewencp: Why did the name change? Not that big a deal, but (despite my writing), I am a fan of concise names and actually putting "constructor" in the names of constructors is unusual (and also, this is not a constructor). It also required changes a bunch of places that muddy the git history :)

ewencp:
@granders Good catch, that could be a nasty bug. Left a couple of comments, but LGTM.

ducktape/tests/test.py
granders: Yes this is a good point. My concern with `copy` is that it suggests it actually copies the full original object. That's not quite true since the newly constructed object does not include a copy of the other object's `ServiceRegistry` object for example.

 `copy_constructor` is maybe not that much better of a name in this regard...

granders:
Nice, I wasn't familiar with python's `id` function before

Issue #122, opened 2015-10-27
log test exceptions to test_log.info in addition to session_log.info


granders:
@ewencp noticed that exceptions and the stacktrace from exceptions thrown in tests are reported in the session log, but not in the relevant test log.

The fix is to update `runner.py` to log the same stacktrace info to `self.current_test.logger`

granders:
Fixed via  #123

PR #123, opened 2015-10-28
Minor logging enhancement

granders:
@ewencp Addresses https://github.com/confluentinc/ducktape/issues/122

This cleans up the printout during test run a bit, and propagates test error to test_log.* and session_log.*

Also, removes the pesky "default" which kept appearing in jenkins console...


ducktape/tests/runner.py
```diff
@@ -117,10 +117,10 @@ def run_all_tests(self):
                 result.stop_time = time.time()
                 self.results.append(result)
 
+                self.log(logging.INFO, "Summary: %s" % str(result.summary))
+                self.log(logging.INFO, "Data: %s" % str(result.data))
                 test_reporter = SingleResultFileReporter(result)
                 test_reporter.report()
-                test_reporter = SingleResultStdoutReporter(result)
```
ewencp: At a minimum, there's an extraneous import now. But also, if we're not even creating this anymore, do we need to keep the class around at all?

ewencp:
@granders LGTM. I'd also clean out the bit of code I pointed out if you can.

ewencp:
Actually, I think I take that back. Having *some* delineation between tests was useful. Worst case, maybe just add a blank line after the `Data` line?

granders:
@ewencp Added delimiter

ewencp:
@granders much better, thanks.

Issue #95, opened 2015-09-04
Gather log files etc even if --no-teardown flag is used


granders:
Existing behavior omits the step where logs are collected from remote machines if `--no-teardown` flag is supplied.

In my experience however, this behavior is somewhat annoying in particular since you'd only really use `---no-teardown` while debugging

PR #124, opened 2015-10-28
Collect logs even if no_teardown is specified

granders:
@ewencp Another small PR. Addresses https://github.com/confluentinc/ducktape/issues/95

It is useful to have collected the logs to the driver machine even if we don't tear down services.

### DIFF
--- a/ducktape/tests/runner.py
+++ b/ducktape/tests/runner.py
@@ -110,15 +110,7 @@ class SerialTestRunner(TestRunner):
                 self.stop_testing = self.session_context.exit_first or isinstance(e, KeyboardInterrupt)
 
             finally:
-                if not self.session_context.no_teardown:
-                    self.log(logging.INFO, "tearing down")
-                    self.teardown_single_test()
-                else:
-                    # Collect logs even if no_teardown is specified
-                    try:
-                        self.current_test.copy_service_logs()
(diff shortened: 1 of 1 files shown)


ducktape/tests/runner.py
```diff
@@ -113,6 +113,12 @@ def run_all_tests(self):
                 if not self.session_context.no_teardown:
                     self.log(logging.INFO, "tearing down")
                     self.teardown_single_test()
+                else:
```
ewencp: Would it make more sense to just add a flag to `teardown_single_test` so we keep the code unified and don't duplicate parts of it? I'm actually noticing that we also don't free nodes (which is fine here since reusing them doesn't make much sense if you wanted to preserve the running services) and that with this copied code, we also don't set the `stop_testing` flag if something goes wrong.

Neither is critical now since both should be fine with the no teardown flag since they only affect things when you run multiple tests, but I'm wondering if we're just going to keep having to think about the two code paths separately and remember to fill things in for both.

ewencp:
@granders Left a comment about organization, but this looks fine. Feel free to merge regardless of what you decide on my comment.

ducktape/tests/runner.py
granders: Good point, that is better

Issue #114, opened 2015-10-16
Cache Vagrant ssh info


granders:
Fetching vagrant ssh info every test run is slow: 10 - 30 seconds for me locally, and one user has reported up to a minute!

The cluster information shouldn't change very often, so we could cache this information in a file.

So what to do if the cache is out of date?
One solution is to update the Vagrant cluster implementation:
- cache ssh info (or rather the json containing this info) to a file
- if the file is present, use it
- if the file is present, warn the user that we're using cached ssh info, and if they have updated the cluster, they should remove the file to allow Vagrant cluster to refresh the ssh info

lindong28:
Instead of having user explicit removing the file from disk, it is probably more intuitive to provide command line option for user. How about the following scheme:

1) by default Ducktape will always fetch Vagrant ssh info and cached the retrieved info in a file
2) if specified --use-cached-info is specified AND cached info is not available, then same as 1)
3) if specified --use-cached-info is specified AND cached info is available, then use cached info

granders:
Yeah, a flag would make manual file deletion unnecessary, which is better. One question I have about such a flag is whether it's generic enough to apply to potentially different cluster types. Maybe it is?

I'd opt for the default behavior to use the cache if it's present, since this will be the most frequent use-case. In other words:

Default: use cache if present; else, lookup `ssh-config` and repopulate cache
`--refresh-cluster` flag: lookup `ssh-config` and repopulate cache

lindong28:
@granders Currently I assume this is only useful for VagrantCluster. I don't think we need to apply this flag to other cluster types like localhost unless there is reason to do so. What do you think?

While this is indeed the most frequent use-case, I am sill in favor of not using cache by default, because correctness is more important than performance. If default behavior is to use cache, and user has re-initialized the virtual machine, then their future ducktape test will fail by default. It is usually easy to say "this flag will improve your performance" than saying "your job may fail without this flag". Just my 2c.

PR #126, opened 2015-11-09
issue-114; Cache Vagrant ssh info

lindong28:


lindong28:
@granders This patch addresses https://github.com/confluentinc/ducktape/issues/114. Can you review this patch?

ducktape/cluster/json.py
```diff
@@ -23,11 +23,8 @@ class JsonCluster(Cluster):
     An implementation of Cluster that uses static settings specified in a cluster file.
     """
 
-    def __init__(self, cluster_json=None):
+    def __init__(self, cluster_json, *args, **kwargs):
         super(JsonCluster, self).__init__()
-        if cluster_json is None:
```
ewencp: This branch was intended to support manually defined clusters, i.e. if you have a set of nodes and just want to maintain a list of them manually. It's important to support clusters like this since not everyone's testbeds will be managed by Vagrant.

ducktape/cluster/vagrant.py
```diff
@@ -76,8 +88,21 @@ def __init__(self):
...
+            externally_routable_ip[str(node_account)] = node_account.externally_routable_ip
+
+        # cache cluster info by writing info to a local file in JSON format.
+        cluster_json["is_aws"] = is_aws
+        cluster_json["externally_routable_ip"] = externally_routable_ip
+
+        # Do not create a new file and cache cluster info in unit test
+        try:
+            json.dump(cluster_json, open(ConsoleConfig.CLUSTER_INFO_FILE, "w"))
+        except:
```
ewencp: Can we catch a specific exception type? Catching all exceptions like this is error prone because if something changes and you start throwing some *other* exception, you won't be able to see the issue.

ewencp:
@lindong28 Left a couple of comments, but this looks nice -- the long startup pauses waiting for vagrant to look up info are quite annoying and will be nice to get rid of them!

ducktape/cluster/vagrant.py
lindong28: Certainly. I updated the patch to catch IOError here.

ducktape/cluster/vagrant.py
```diff
@@ -24,10 +28,18 @@ class VagrantCluster(JsonCluster):
     advertised, this assumes that the Vagrant VM's name is a routeable hostname on all the hosts.
     """
 
-    def __init__(self):
+    def __init__(self, *args, **kwargs):
         hostname, ssh_hostname, username, flags = None, None, None, ""
         nodes = []
 
+        if kwargs.get("use_cluster_info", False):
```
granders: Is there a reason this isn't in `JsonCluster`? It seems like it's cached json so it's equally applicable.

Also, by moving loading from file exclusively into Vagrant cluster, I think we now can't instantiate a non-Vagrant json cluster from file

ducktape/cluster/json.py
lindong28: @ewencp I removed it because I thought this branch is never used in the code. I just realized it is used, when the user specifies `--cluster=ducktape.cluster.json.JsonCluster`. 

I find that these two options, `--cluster=ducktape.cluster.json.JsonCluster` and `--use-cluster-info`, overlap with each other. Both provide the functionality to allow user to load the cluster info from a json file with hard-coded path on disk, which seems a little confusing. And this may also be useful to allow user to specify path to json file, instead of creating/editing the file with the hard-coded path.

Would it be better to provide a single option, say `--use-cluster-info=/path/to/jsonfile`, which will load the cluster info from that json file?

ducktape/cluster/vagrant.py
lindong28: @granders It is in `VagrantCluster`, because conceptually this cached file is written a running Ducktape with `VagrantCluster`. It probably makes sense to read/write the file in the same class. 

Another reason is that, if this is in the `JsonCluster`, user needs to specify both `--cluster=ducktape.cluster.json.JsonCluster` and `--use-cluster-info`, which is inconvenient.

Actually I share with you the opinion that this functionality here is very similar to the one in `JsonCluster`. I have provided one solution to consolidate the two, in the reply to @ewencp 's comment. Can you have a look?

lindong28:
@ewencp @granders Can you have a look at this patch? Thanks.

lindong28:
Can anyone review this patch?

granders:
@lindong28 Thanks for the ping, will take a look

lindong28:
Great. Thanks @granders

ducktape/cluster/json.py
granders: @lindong28 
We definitely still need to be able to bring up a `JsonCluster` from a file, but I see your point about this overlap. It would be good to sort out what we want the user-facing behavior to be here.

Below is an idea to unify the usage somewhat, what do you think?

example usage:
JsonCluster:
`ducktape --cluster ductape.cluster.json.JsonCluster --cluster-file cluster.json ...`
Behavior:
- load in cluster information from file as before, but user specifies location of file.
- (optionally allow `cluster.json` to continue to be the default location?)

VagrantCluster (the default)
`ducktape --cluster-file cluster.json ...`
Behavior:
- If cluster-file is not specified, gather json data every time as we do today by calling `vagrant ssh-info`
- If cluster-file is specified, but file is not present, call `vagrant ssh-info` and cache this information to cluster-file for future use
- If cluster-file is specified and file is present, load directly from file

lindong28:
@granders Thanks for your comment! I agree with your suggestion and have updated patch to implement the suggested usage. Please let me know if it looks good now.

ducktape/cluster/json.py
```diff
@@ -23,11 +23,14 @@ class JsonCluster(Cluster):
...
     """
 
-    def __init__(self, cluster_json=None):
+    def __init__(self, cluster_json=None, *args, **kwargs):
         super(JsonCluster, self).__init__()
         if cluster_json is None:
-            cluster_json_path = os.path.abspath(os.path.join(os.getcwd(), "cluster.json"))
-            cluster_json = json.load(open(cluster_json_path))
+            # This is a directly instantiation of JsonCluster rather than from a subclass
+            cluster_file = kwargs.get("cluster_file")
```
granders: minor - this looks equivalent to
`cluster_file = kwargs.get("cluster_file", ConsoleConfig.CLUSTER_FILE)`

ducktape/cluster/vagrant.py
```diff
@@ -12,9 +12,13 @@
...
 
-from .cluster import Cluster, ClusterSlot
-from .json import JsonCluster
+from __future__ import absolute_import
 
+from .json import JsonCluster
+from .cluster import Cluster, ClusterSlot
+from ducktape.command_line.config import ConsoleConfig
+import os
+import json
```
granders: Hm I see why you had to do `from __future__ import absolute_import` here...

I wonder if we should rename our `cluster.json` module to `cluster.json_cluster` to avoid  this kind of collision?

ducktape/cluster/vagrant.py
```diff
@@ -24,10 +28,53 @@ class VagrantCluster(JsonCluster):
     advertised, this assumes that the Vagrant VM's name is a routeable hostname on all the hosts.
     """
 
-    def __init__(self):
-        hostname, ssh_hostname, username, flags = None, None, None, ""
-        nodes = []
+    def __init__(self, *args, **kwargs):
```
granders: Class docstring should be updated to reflect new behavior

ducktape/cluster/vagrant.py
```diff
@@ -24,10 +28,53 @@ class VagrantCluster(JsonCluster):
...
+        is_read_from_file = False
 
+        cluster_file = kwargs.get("cluster_file")
+        if cluster_file is not None:
+            try:
+                cluster_json = json.load(open(os.path.abspath(cluster_file)))
+                is_read_from_file = True
+            except IOError:
+                # It is OK if file is not found. Call vagrant ssh-info to read the cluster info.
+                pass
```
granders: minor - you can put `is_read_from_file = False` in the except block; the `is_read_from_file` will still scope to the `__init__` method

ducktape/cluster/vagrant.py
ewencp: I ran into the same issue when prototyping some new code. I don't think we should rename since the absolute import is required in the future (3.0+) anyway. In the few places we need it, we can just use the `from __future__` import.

ducktape/cluster/vagrant.py
```diff
@@ -24,10 +28,53 @@ class VagrantCluster(JsonCluster):
...
+                node_account.externally_routable_ip = self._externally_routable_ip(is_aws, node_account)
+                externally_routable_ip[str(node_account)] = node_account.externally_routable_ip
+
+        # If cluster_file is specified, cache cluster to the file
+        if cluster_file is not None:
+            cluster_json["externally_routable_ip"] = externally_routable_ip
+            try:
+                json.dump(cluster_json, open(cluster_file, "w"))
+            except IOError:
+                # Do not create a new file and cache cluster info in unit test
```
granders: Though this may not warrant killing the test run, there should at least be a warning printed to indicate that the write to the `cluster_file` path failed

ducktape/command_line/main.py
```diff
@@ -176,7 +178,7 @@ def main():
         (cluster_mod_name, cluster_class_name) = args.cluster.rsplit('.', 1)
         cluster_mod = importlib.import_module(cluster_mod_name)
         cluster_class = getattr(cluster_mod, cluster_class_name)
-        session_context.cluster = cluster_class()
+        session_context.cluster = cluster_class(cluster_file = args.cluster_file)
```
granders: python convention - no space around the `=` in the key/val pair in kwargs of a function call

ducktape/cluster/vagrant.py
lindong28: I agree with @ewencp.

ducktape/cluster/json.py
lindong28: By its current implementation, cluster_file is always provided in kwargs. If its value is not specified in command line, kwargs will have ("cluster_file", None) in it. Then `kwargs.get("cluster_file", ConsoleConfig.CLUSTER_FILE)` will return `None`, which doesn't work for us.

We can replace `session_context.cluster = cluster_class(cluster_file = args.cluster_file)` will the following code in main.py. But it simply moves the extra code from one place to another... what do you think?

```
if (args.cluster_file is None):
  session_context.cluster = cluster_class()
else:
  session_context.cluster = cluster_class(cluster_file = args.cluster_file)
```

ducktape/cluster/vagrant.py
lindong28: Sure. Will do.

ducktape/cluster/vagrant.py
lindong28: I agree we should print a message to make that explicit.

With regard to your first point, are you saying that I should move `is_read_from_file = False` from its current position to the except block? What would be the value of `is_read_from_file` if cluster_file is None?

ducktape/cluster/vagrant.py
lindong28: Sure. Will do it.

ducktape/command_line/main.py
lindong28: Sure, will fix it.

ducktape/cluster/json.py
```diff
@@ -12,9 +12,9 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
+from ducktape.command_line.config import ConsoleConfig
```
granders: I think you need `from __future__ import absolute_import` here as well. 

When I tested with `JsonCluster` on my local machine, it failed to load the json file because it was picking up the wrong `json` module. However using absolute_import fixed this

lindong28: Thanks for catching this. I will fix it.

ducktape/cluster/vagrant.py
lindong28: BTW, I intentionally left "unit test" in the comment. Because this try/catch is specifically added to allow unit test to work. Other than unit test, I don't find a good reason that we will have IOError here, unless there is something very wrong (e.g. read only file system). It is OK to let test fail in that case, right?

ducktape/cluster/vagrant.py
```diff
@@ -24,10 +28,53 @@ class VagrantCluster(JsonCluster):
...
+        if not is_read_from_file:
+            cluster_json = {
+                "nodes": self._get_nodes_from_vagrant()
+            }
+
+        super(VagrantCluster, self).__init__(cluster_json)
+
+        # Load externally_routable_ip for each account
+        if is_read_from_file:
+            externally_routable_ip = cluster_json["externally_routable_ip"]
```
granders: This bit was confusing to me at first, but I think I understand now what's going on. 

This actually makes the schema of the Vagrant cluster json different from that of a corresponding JsonCluster, and I think this is not quite what we want. See if this makes sense to you:

If I cache cluster info and examine the `RemotAccount` objects created with a call like this

`ducktape test/path --cluster-file cluster.json`

I would expect those `RemoteAccount` objects to be identical to `RemoteAccount` objects created if I then instantiated the cluster this way:

`ducktape test/path --cluster ducktape.cluster.json.JsonCluster --cluster-info cluster.json`

This isn't true because in the first case, the `RemoteAccount` objects will have a non-None `externally_routable_ip` field, but in the second case, `externally_routable_ip` field will be None.

---

In other words, as long as `cluster.json` exists, it's conceptually simpler if these two commands are completely interchangeable:
`ducktape test/path --cluster-file cluster.json`
`ducktape test/path --cluster ducktape.cluster.json.JsonCluster --cluster-info cluster.json`

---

I think the json written to file should look like this: 

```
{
  "nodes": [{"ssh_hostname": "127.0.0.1",
                   "externally_routable_ip": "192.68.asfdasdfasd"
                   "hostname": "ducktape1", 
                   "user": "vagrant", 
                   "ssh_args": "-o 'HostName 127.0.0.1' -o 'Port 2222' -o 'UserKnownHostsFile /dev/null' -o 'StrictHostKeyChecking no' -o 'PasswordAuthentication no' -o 'IdentityFile /Users/geoffreyanderson/Google_Drive/Confluent_code/ducktape-dong/.vagrant/machines/ducktape1/virtualbox/private_key' -o 'IdentitiesOnly yes' -o 'LogLevel FATAL' "
}]
}
```

and `JsonCluster` can instantiate its list of `RemoteAccount` objects with the additional `externally_routable_ip` field.

Does this seem reasonable?

ducktape/cluster/vagrant.py
```diff
@@ -24,10 +28,53 @@ class VagrantCluster(JsonCluster):
...
+            externally_routable_ip = {}
+            for node_account in self.available_nodes:
+                node_account.externally_routable_ip = self._externally_routable_ip(is_aws, node_account)
+                externally_routable_ip[str(node_account)] = node_account.externally_routable_ip
+
+        # If cluster_file is specified, cache cluster to the file
+        if cluster_file is not None:
+            cluster_json["externally_routable_ip"] = externally_routable_ip
+            try:
+                json.dump(cluster_json, open(cluster_file, "w"))
```
granders: It wouldn't hurt to prettify this so a user can open the json file and read it more easily

ducktape/command_line/main.py
```diff
@@ -47,6 +47,8 @@ def parse_args():
                         help="path to project-specific configuration file.")
     parser.add_argument("--cluster", action="store", default=ConsoleConfig.CLUSTER_TYPE,
                         help="cluster class to use to allocate nodes for tests.")
+    parser.add_argument("--cluster-file", action="store", default=None,
+                        help="path to a json file which provides information needed to initialize a json cluster")
```
granders: Mention the caching behavior when used with Vagrant cluster?

granders:
@lindong28 Thanks for the updates so far, this is looking really useful. My biggest comment is about the schema of the json stored with a `VagrantCluster` vs `JsonCluster`.

One more bit as well: we should probably add a few unit tests to validate the caching behavior of the class
- caches to file when `--cluster-file` option is present, but the actual cluster file doesn't exist
- loads from cache when `--cluster-file` option is present, and when actual cluster file exists

You'll probably need to do some monkeypatching like in `check_one_host_parsing` in `check_vagrant.py`

ducktape/cluster/json.py
granders: Nevermind, this makes sense now

ducktape/cluster/vagrant.py
granders: sounds good

ducktape/cluster/vagrant.py
granders: Sorry - my initial comment about `is_read_from_file` was actually incorrect, so ignore please :)

ducktape/cluster/vagrant.py
granders: Hm interesting... do you mean the `check_vagrant` unit test, or another one? I wouldn't expect the existing tests to even trigger this path, so that sounds suspicious...

I agree with your second comment - it's probably reasonable to let the exception bubble up if there's a problem writing to the file

ducktape/cluster/vagrant.py
lindong28: @granders Good point. This is very reasonable. 

However, since you mentioned this, I find it a bit weird that `externally_routable_ip` is assigned in VagrantCluster but not in JsonCluster. Does this suggest that the existing implementation of JsonCluster doesn't actually work because it doesn't assign value to `RemoteAccount.externally_routable_ip`?

ducktape/cluster/vagrant.py
lindong28: Sure. Will do it.

ducktape/command_line/main.py
lindong28: Sure. Will do it.

ducktape/cluster/vagrant.py
lindong28: I am very sure that I added this try/catch because one unit test failed earlier. But I couldn't reproduce this now. Maybe the behavior changed due to some other commits checked-in between.

I will remove this try/catch.

ducktape/cluster/vagrant.py
lindong28: Fixed now.

lindong28:
@granders Thanks so much for detailed review. I have addressed all your comments. I will do an integration test and add unit test later.

ducktape/cluster/vagrant.py
```diff
@@ -22,12 +26,49 @@ class VagrantCluster(JsonCluster):
...
+                "nodes": self._get_nodes_from_vagrant()
+            }
+
+        super(VagrantCluster, self).__init__(cluster_json)
+
+        is_aws = self._is_aws()
+        if not is_read_from_file:
+            # go through and find fully qualified domain name for each node
+            # this makes it possible to not require write access to /etc/hosts on the test driver machine
+            externally_routable_ip = {}
```
granders: should be able to delete this now since you're attaching the data directly to each node

ducktape/cluster/vagrant.py
```diff
@@ -22,12 +26,49 @@ class VagrantCluster(JsonCluster):
...
+            for node_account in self.available_nodes:
+                node_account.externally_routable_ip = self._externally_routable_ip(is_aws, node_account)
+                externally_routable_ip[str(node_account)] = node_account.externally_routable_ip
```
granders: can delete

ducktape/cluster/vagrant.py
```diff
@@ -22,12 +26,49 @@ class VagrantCluster(JsonCluster):
...
 
+            # If cluster_file is specified, cache cluster info in the file
+            if cluster_file is not None:
+                cluster_json["externally_routable_ip"] = externally_routable_ip
```
granders: can delete

ducktape/cluster/vagrant.py
lindong28: Can you specify which lines of code can be deleted? The commend in github refers to an entire block code changed code.

I double checked code but couldn't understand which lines to delete...

ducktape/cluster/vagrant.py
lindong28: I think this line is still useful. In the case where use specified --cluster-file to be used with VagrantCluster and the file doesn't exist yet, we need to write the cluster info to the file.

ducktape/cluster/vagrant.py
granders: It looks like `externally_routable_ip` still isn't being attached to the individual nodes in `cluster_json` - 
so the schema of the json created by `VagrangCluster` and that expected by `JsonCluster` is still slightly different

granders:
@lindong28 Thanks for the updates. The schemas still don't quite match between `VagrantCluster` and `JsonCluster`, but that's a pretty small extra tweak.

I'll keep an eye out for the tests

ducktape/cluster/vagrant.py
lindong28: Suppose the cluster info is read from file, then `JsonCluster` will be initialized with the `externally_routable_ip`.

Otherwise, it is true that `JsonCluster` will be initialized without `externally_routable_ip`. However, this seems inevitable. This is because `RemoteAccount` needs to be created in `JsonCluster`, `externally_routable_ip` for vagrant cluster should be read in `VagrantCluster` via ssh command, and the ssh command should be issued via `RemoteAccount`. I don't have a better solution here. Do you have a solution for this?

ducktape/cluster/vagrant.py
granders: Yep that's right we want this info but in slightly different format (hopefully my next comment will clarify this)

granders:
Here's the format we want from `VagrantCluster` (and the format `JsonCluster` currently expects)

```
{
  "nodes": [
    {
      "hostname": "ducktape1",
      "externally_routable_ip": "192.168.50.151",
      "ssh_args": "-o 'HostName 127.0.0.1' -o 'Port 2222' -o 'UserKnownHostsFile /dev/null' -o 'StrictHostKeyChecking no' -o 'PasswordAuthentication no' -o 'IdentityFile /Users/geoffreyanderson/Google_Drive/Confluent_code/ducktape-dong/.vagrant/machines/ducktape1/virtualbox/private_key' -o 'IdentitiesOnly yes' -o 'LogLevel FATAL' ",
      "ssh_hostname": "127.0.0.1",
      "user": "vagrant"
    }
  ]
}
```

Instead, here is the format currently output by `VagrantCluster`:
```
{
  "externally_routable_ip": {
    "vagrant@ducktape1": "192.168.50.151"
  },
  "nodes": [
    {
      "hostname": "ducktape1",
      "ssh_args": "-o 'HostName 127.0.0.1' -o 'Port 2222' -o 'UserKnownHostsFile /dev/null' -o 'StrictHostKeyChecking no' -o 'PasswordAuthentication no' -o 'IdentityFile /Users/geoffreyanderson/Google_Drive/Confluent_code/ducktape-dong/.vagrant/machines/ducktape1/virtualbox/private_key' -o 'IdentitiesOnly yes' -o 'LogLevel FATAL' ",
      "ssh_hostname": "127.0.0.1",
      "user": "vagrant"
    }
  ]
}
```

ducktape/cluster/vagrant.py
lindong28: Ah I see your point... You are saying that the externally_routable_ip should be attached to each node. My bad. Will update the patch.

granders:
@lindong28 Does this clarify my previous comments about tweaking the schema of the cached json?

lindong28:
@granders Yes, you are right. Thanks for your detailed explanation. Will update the patch to fix it.

ducktape/cluster/vagrant.py
granders: Yep that's right, sorry I wasn't quite clear before.

lindong28:
@granders In addition to fixing the problem you mentioned, I refactored the code a bit to make it more readable. I will add test case and do integration test later.

ducktape/cluster/vagrant.py
```diff
@@ -22,12 +26,47 @@ class VagrantCluster(JsonCluster):
...
+        if not is_read_from_file:
+            cluster_json = {
+                "nodes": self._get_nodes_from_vagrant()
+            }
 
+        super(VagrantCluster, self).__init__(cluster_json)
+
+        # If cluster file is specified but the cluster info is not read from it, write the cluster info into the file
+        if not is_read_from_file and cluster_file is not None:
+            nodes = [{"hostname": node_account.hostname,
```
granders: nice

ducktape/cluster/json.py
```diff
@@ -12,29 +12,45 @@
...
-            init_nodes = [RemoteAccount(ninfo["hostname"], ninfo.get("user"), ninfo.get("ssh_args"),
-                                        ssh_hostname=ninfo.get("ssh_hostname"))
-                                        for ninfo in cluster_json["nodes"]]
+            node_accounts = [RemoteAccount(ninfo["hostname"], ninfo.get("user"), ninfo.get("ssh_args"),
+                                           ssh_hostname=ninfo.get("ssh_hostname"),
+                                           externally_routable_ip=ninfo.get("externally_routable_ip"))
+                                           for ninfo in cluster_json["nodes"]]
+            for node_account in node_accounts:
+                if node_account.externally_routable_ip is None:
+                    node_account.externally_routable_ip = self._externally_routable_ip(node_account)
```
granders: Interesting idea. The only problem with moving this here and calling it every time is that we lose some of the benefits of caching in Vagrant - In `VagrantCluster`, this does an ssh command on each remote machine which incurs noticeable overhead...

lindong28: Can you elaborate what is the negative effect of doing this? I think this will have the same overhead as previous approach, right? During initialization, an ssh command will be executed on each remote machine before and after this patch.

granders: Sure, see if this makes sense - I noticed this issue because I installed your branch and tried the following locally:

`ducktape TESTPATH --cluster "ducktape.cluster.json.JsonCluster" --cluster-file cluster.json`
vs
`ducktape TESTPATH --cluster-file cluster.json`

The first form was noticeably faster to start runnings tests even with a minimal cluster. 

The difference seems to be in the call to `self._externally_routable_ip`: in the first case (`JsonCluster`), it does no work, but in the second case (`VagrantCluster`) it actually results in the ssh calls to fetch `externally_routable_ip` even when this information is already cached.

lindong28: In the case where we execute `ducktape TESTPATH --cluster-file cluster.json`, does cluster.json contains `externally_routable_ip` for each node?

If it does, according to this patch, `_externally_routable_ip()` will  not be called. Otherwise, we need to call `_externally_routable_ip()` to retrieve IP for each node anyway, before or after this patch. Does this make sense?

granders: Oops, yes you're exactly right. I looked into this a little more carefully and found the source of the discrepancy between time; see my next comment on `is_aws`

ducktape/cluster/vagrant.py
```diff
@@ -22,12 +26,47 @@ class VagrantCluster(JsonCluster):
...
+      - If cluster_file exists on the filesystem, read cluster info from the file
+      - Otherwise, retrieve cluster info via "vagrant ssh-config" from vagrant and write cluster info to cluster_file
+    - Otherwise, retrieve cluster info via "vagrant ssh-config" from vagrant
     """
 
-    def __init__(self):
-        hostname, ssh_hostname, username, flags = None, None, None, ""
-        nodes = []
+    def __init__(self, *args, **kwargs):
+        self.is_aws = self._is_aws()
```
granders: This line results in opening a subprocess that calls `vagrant status` regardless of whether cluster info is cached, which results in unnecessary overhead.

I think we can simply make `self.is_aws` a property that gets lazily evaluated. E.g.

```
    @property
    def is_aws(self):
        if self._is_aws is None:
            proc = subprocess.Popen("vagrant status", shell=True, stdout=subprocess.PIPE, stderr=subprocess.STDOUT)
            output, _ = proc.communicate()
            self._is_aws = output.find("aws") >= 0

        return self._is_aws
```

lindong28: Yeah you are right. Thanks for catching this. Patch is updated.

granders:
Thanks for the `is_aws` change - the overhead of bringing a Vagrant cluster with a cached cluster-file is now comparable to the overhead for bringing up a JsonCluster from the same file.

granders:
@lindong28 This is looking good - once we have a few unit tests we should be good to go

lindong28:
@granders Thanks much for your review and helpful feedback!

I would love to finish the unit tests and commit this patch early. It is just that I am assigned some tasks in LinkedIn that I have to finish soon... Sorry for the delay.

lindong28:
@granders Sorry for the delay.. I am overloaded by personal and LinkedIn's work.

I have added unit tests as you suggested. Please have a look. Thanks!

tests/cluster/check_vagrant.py
```diff
@@ -39,12 +41,13 @@ class CheckVagrantCluster(object):
...
 
+    cluster_file = "cluster_file_temporary.json"
+
     def check_one_host_parsing(self, monkeypatch):
         self._vagrant_ssh_data = self.two_hosts
         monkeypatch.setattr("ducktape.cluster.vagrant.VagrantCluster._vagrant_ssh_config", lambda vc: (self._vagrant_ssh_data, None))
-        monkeypatch.setattr("ducktape.cluster.vagrant.VagrantCluster._is_aws", lambda vc: False)
-        monkeypatch.setattr("ducktape.cluster.vagrant.VagrantCluster._externally_routable_ip", lambda vc, is_aws, node_account: "127.0.0.1")
-
+        monkeypatch.setattr("ducktape.cluster.vagrant.VagrantCluster.is_aws", lambda vc: False)
```
granders: Looks like monkeypatching behavior is identical in these 3 tests now - maybe break this out into a shared method?

tests/cluster/check_vagrant.py
```diff
@@ -60,3 +63,73 @@ def check_one_host_parsing(self, monkeypatch):
         assert(node2.account.user == "vagrant")
         assert(node2.account.ssh_args.strip() == "-o 'HostName 127.0.0.2' -o 'Port 2200' -o 'UserKnownHostsFile /dev/null' -o 'StrictHostKeyChecking no' -o 'PasswordAuthentication no' -o 'IdentityFile /Users/foo/ducktape.git/.vagrant/machines/worker2/virtualbox/private_key' -o 'IdentitiesOnly yes' -o 'LogLevel FATAL'")
         assert(node2.account.ssh_hostname == '127.0.0.2')
+
+    def check_cluster_file_write(self, monkeypatch):
+        self._vagrant_ssh_data = self.two_hosts
+        monkeypatch.setattr("ducktape.cluster.vagrant.VagrantCluster._vagrant_ssh_config", lambda vc: (self._vagrant_ssh_data, None))
+        monkeypatch.setattr("ducktape.cluster.vagrant.VagrantCluster.is_aws", lambda vc: False)
+        monkeypatch.setattr("ducktape.cluster.vagrant.VagrantCluster._externally_routable_ip", lambda vc, node_account: "127.0.0.1")
+        try:
```
granders: Instead of `try/except` block here, perhaps use `assert(not os.path.exists(file_path))`?

lindong28: Good point. Fixed now.

tests/cluster/check_vagrant.py
lindong28: Sure, will move these into one function.

tests/cluster/check_vagrant.py
```diff
@@ -60,3 +63,73 @@ def check_one_host_parsing(self, monkeypatch):
...
+        json.dump(cluster_json_expected, open(self.cluster_file, 'w+'), indent=2, separators=(',', ': '), sort_keys=True)
+
+        cluster = VagrantCluster(cluster_file=self.cluster_file)
+        os.remove(self.cluster_file)
+
+        assert len(cluster) == 2
+        assert(cluster.num_available_nodes() == 2)
+        node1, node2 = cluster.request(2)
+
+        assert(node1.account.hostname == "worker2")
```
granders: `assert` in python is actually not a function -
My understanding is that it actually evaluates something like `assert(a == b)` as
```
my_tuple = (a == b)
assert my_tuple
```
rather than as a function call

This can be surprising if you try to add an error statement. e.g. `assert(False, "should fail")` will actually pass because
```
my_tuple = (False, "should fail")  # tuple with 2 elements!
assert my_tuple  # tuples with multiple entries are truthy!!
```

In other words, it's safer/more correct to remove parens from the assertions. I realize the previously existing test had assertions with parens, so it'd be great to correct those as well

tests/cluster/check_vagrant.py
```diff
@@ -60,3 +63,73 @@ def check_one_host_parsing(self, monkeypatch):
...
+        assert(cluster_json_actual == cluster_json_expected)
+
+    def check_cluster_file_read(self, monkeypatch):
+        self._vagrant_ssh_data = self.two_hosts
+        monkeypatch.setattr("ducktape.cluster.vagrant.VagrantCluster._vagrant_ssh_config", lambda vc: (self._vagrant_ssh_data, None))
+        monkeypatch.setattr("ducktape.cluster.vagrant.VagrantCluster.is_aws", lambda vc: False)
+        monkeypatch.setattr("ducktape.cluster.vagrant.VagrantCluster._externally_routable_ip", lambda vc, node_account: "127.0.0.1")
+
+        nodes = []
+        nodes.append({
```
granders: Am I right in understanding that you're deliberately creating json with data that differs from the data in the mock cluster to ensure the cluster file overrides info from vagrant_ssh? It's a great idea, and I think it'd be good to call out that that is intentional in a comment.

lindong28:
@granders Thanks for the commend. Patch is updated. BTW, when you push this patch to master branch, is it possible to merge all commits into one commit, similar to what we do for kafka? I noticed commits are not merged in the past.

tests/cluster/check_vagrant.py
lindong28: Exactly. I will add comment in the code.

tests/cluster/check_vagrant.py
```diff
@@ -60,3 +65,63 @@ def check_one_host_parsing(self, monkeypatch):
         assert(node2.account.user == "vagrant")
         assert(node2.account.ssh_args.strip() == "-o 'HostName 127.0.0.2' -o 'Port 2200' -o 'UserKnownHostsFile /dev/null' -o 'StrictHostKeyChecking no' -o 'PasswordAuthentication no' -o 'IdentityFile /Users/foo/ducktape.git/.vagrant/machines/worker2/virtualbox/private_key' -o 'IdentitiesOnly yes' -o 'LogLevel FATAL'")
         assert(node2.account.ssh_hostname == '127.0.0.2')
+
+    def check_cluster_file_write(self, monkeypatch):
+        self._set_monkeypatch_attr(monkeypatch)
```
granders: A short docstring to demonstrate intent of the test would be great

tests/cluster/check_vagrant.py
```diff
@@ -60,3 +65,63 @@ def check_one_host_parsing(self, monkeypatch):
...
+                  "externally_routable_ip": node_account.externally_routable_ip}
+                  for node_account in cluster.available_nodes]
+        cluster_json_expected["nodes"] = nodes
+        cluster_json_actual = json.load(open(os.path.abspath(self.cluster_file)))
+
+        os.remove(self.cluster_file)
+        assert(cluster_json_actual == cluster_json_expected)
+
+    def check_cluster_file_read(self, monkeypatch):
+        self._set_monkeypatch_attr(monkeypatch)
```
granders: A short docstring to demonstrate intent of the test would be great

lindong28: Fixed now.

tests/cluster/check_vagrant.py
lindong28: Fixed now.

tests/cluster/check_vagrant.py
lindong28: Agree. I fixed all occurrences in check_vagrant.py. I will leave occurrences in other files for change in the future.

granders:
@lindong28 Thanks very much for the unit tests, this LGTM.

Regarding squashing commits - not a bad idea, although I think there are tradeoffs to both approaches. For the time being, we're merging PRs as-is - however, if you'd like to rebase/squash and push before I merge this PR that's totally fine by me.

lindong28:
@granders Thanks for your review as well!

I prefer not to squash commits on my side because that will cause the entire history of patch review to be discarded on github. Please merge this PR. Thanks.

PR #127, opened 2015-11-10
Make kill_process work for non-Java processes.

ewencp:


ewencp:
@granders Trivial, but potentially dangerous? It's completely historical artifact that this only works for Java and annoying to not have a non-Java version handy if you're working with other processes.

granders:
@ewencp Hm for existing uses, piping to `grep java` at worst reduced the set of hits, so removing that step should probably be fine.

We probably don't want to change the version kafkatest 0.9.0 is pinned to at this point anyway, so LGTM

ewencp:
@granders yeah, the only concern is if we have any existing uses where increasing the scope might kill important processes (e.g. if it accidentally caught `sshd`, `syslog`, etc). But I suppose the risk there is pretty low.

PR #128, opened 2015-11-11
Fix incorrect return when copying logs that causes us to skip remaining services if one service doesn't have a list of log files to collect.

ewencp:


ewencp:
@granders It took me far too long to track down a pretty silly issue here....

granders:
@ewencp Thanks, good catch. Sorry that is an annoying error. LGTM

PR #134, opened 2016-02-11
issue-133; Provide optional sudo_user in RemoteAccount for use in ssh

lindong28:
This is useful in the scenario where user has to execute cmd using another identity in the remote host. For example, it is required to do "sudo -u app ..." in order to start broker in LinkedIn's cluster.

### DIFF
--- a/ducktape/cluster/remoteaccount.py
+++ b/ducktape/cluster/remoteaccount.py
@@ -23,13 +23,14 @@ import tempfile
 from contextlib import contextmanager
 
 class RemoteAccount(HttpMixin):
-    def __init__(self, hostname, user=None, ssh_args=None, ssh_hostname=None, externally_routable_ip = None, logger=None):
+    def __init__(self, hostname, user=None, ssh_args=None, ssh_hostname=None, externally_routable_ip = None, logger=None, sudo_user=None):
         self.hostname = hostname
         self.user = user
         self.ssh_args = ssh_args
         self.ssh_hostname = ssh_hostname
         self.externally_routable_ip = externally_routable_ip
         self.logger = logger
+        self.sudo_user = sudo_user
 
     def __str__(self):
         r = ""
@@ -62,6 +63,8 @@ class RemoteAccount(HttpMixin):
     def ssh_command(self, cmd):
         if self.local:
             return cmd
+        if self.sudo_user is not None:
+            cmd = "sudo -u %s %s" % (self.sudo_user, cmd)
         r = "ssh "
         if self.user:
             r += self.user + "@"
@@ -69,6 +72,8 @@ class RemoteAccount(HttpMixin):
         if self.ssh_args:
             r += self.ssh_args + " "
         r += "'" + cmd.replace("'", "'\\''") + "'"
+        if self.logger is not None:
+            self.logger.debug("Execute ssh_command: %s" % r)
         return r
 
     def ssh(self, cmd, allow_fail=False):


lindong28:
@granders Would you have time to take a look? This is a very small patch.

granders:
@lindong28 Thanks for this idea - it will definitely be good to provide hooks to make tests usable in more environments.

I have a couple initial questions about implementation however:

- first off, sticking `sudo -u USER` in front of the command won't guarantee it can be run in general. E.g. if cmd is something like `echo Hello; echo Hello >> /restricted_file.txt`, the second portion of the command would still fail.

- how does this info get into the remote account object? The only place these objects are created now is by the cluster class; so as the implementation currently stands, test writers would have to provide this in test code or in service code, which doesn't seem like quite the right approach.

It seems like what we want is something like Packer's "execute_command" concept (see https://www.packer.io/docs/provisioners/shell.html) which is somehow externally configurable, and can be customized in a single place

What do you think?

lindong28:
@granders Thanks for your comment. Here are my thoughts:

- You are right. The problem can be solved by running command `sudo -u user -- bash -c cmd`. For example, sudo -u user -- bash -c 'whoami; whoami'

- As the implementation currently stands, all parameters of the RemoteAccount, such as hostname and user, are specified in the cluster json file which is used in `JsonCluster`. We can specify sudo_user in this json file as well. Given that the sudo_user field is very similar to the user field, we can specify them in the same place if sudo_user is needed. Does this answer your 2nd question?

This sudo_user field is optional. This is not needed for Vagrant cluster. It is useful only if user wants to connect to an remote physical machine, in which case user needs to provide all the cluster information in the json file and provide it to JsonCluster.

- Yes, I think the idea is very similar to Packer's "execute_command" concept. But we currently don't need the full feature as provided in that package. Are you suggesting to import and use that package for executing cmd remotely in Ducktape? Alternatively we can just gradually add additional options after we have use case for these, and consider using other library if we find it is becoming difficult to add new options. What do you think?

granders:
@lindong28
A few thoughts -

I meant the packer reference mostly as an example, not really a dependency to be brought in. My thought was that it'd be useful to provide a similar hook to what you proposed, but which is more general/flexible (`execute_command`? `ssh_command_prefix`?), and externally configurable. This could support your use case, as well as things like passing environment variables etc.

I think I see what you mean about specifying it in the cluster file. The funky thing there is the interaction with Vagrant - if you can only specify this in the cluster-file, the flow is awkward:
- run once so that Vagrant auto-generates cluster-file
- manually update cluster-file to include `execute_command` thing
- now `execute_command` is is present and we can run normally

To make this less awkward, it seems we'd need another command-line arg.

lindong28:
@granders I read the current implementation of various ssh in class the class `RemoteAccount` and read the documentation on packer's `execute_command` as well. But I still don't the question of "which is more general/flexible". Could you elaborate a bit on the difference between `execute_command` and `ssh_command_prefix` that affects our design?

My proposal is to extend `RemoteAccount` to allow user to optionally specify more configuration to execute command remotely. Both `RemoteAccount` and `execute_command` does that. They differ in the way the ssh command is built after the configuration is provided, i.e. `RemoteAccount` use string concatenation whereas `execute_command` uses configuration template and replace the template with variable or function values. After user provides the configurations, how we build the ssh command is really implementation detail and both solutions work.

I prefer the existing way in which `RemoteAccount` constructs cmd simply because it is easy to do. If in the future we find a good reason to switch to configuration template, it can be changed without affecting user. Does this make sense?

granders:
Hey @lindong28, sorry for the slow reply. Hmm, I think maybe we don't actually disagree too much -

 simple string concatenation seems fine to me. In my mind I was thinking of `execute_command` and `ssh_command_prefix` as synonyms for the same basic effect. 

I'm too concerned about how we achieve this effect - my only real concern here is that I think we don't want to hard-code this prefix - but based on my reading of your most recent response, I think we're in agreement there?

lindong28:
Hey @granders, thanks for your comment. 

I am not sure if I understand your question. Could you explain what you mean by "hard-code the prefix" here?

I think the goal is to provide user with the option to specify sudo_user to execute command remotely. One possible solution, as implemented in the patch, is that we provide sudo_user in the RemoteAccount constructor. And then, when we construct ssh command, we use "-u sudo_user" if sudo_user is specified. Is there any undesirable hard-coded part in this approach?

granders:
@lindong28, oops sorry evidently I misunderstood something. As I understand currently, you're proposing version A (below), and I am proposing version B

Option A:
Usage looks like: `ducktape --sudo-user superperson`
We hardcode `sudo -u %(user) -- bash -c %(cmd)`, and commands on remote machines look something like like: `sudo -u superperson -- bash -c cmd`

Option B:
Usage looks something like `ducktape --command-prefix "sudo -u superperson -- bash -c"`
commands on remote machines look like: `sudo -u superperson -- bash -c cmd`

Since in version B we don't assume anything about content or purpose of the prefix, users could also do things like: 
```
ducktape --command-prefix "export ENV_VAR1=1; export ENV_VAR2=/an/important/path"
```
commands on remote machines now look like:
```
export ENV_VAR1=1; export ENV_VAR2=/an/important/path; cmd
```

In this way, I see version B as more general/flexible.

Thoughts?

lindong28:
@granders Ah, it seems that we were discussing different things..

What I wanted for this patch is to modify the RemoteAccount class so that user can pass in "sudo_user" for ssh command. And you are considering how to modify the command line argument for Ducktape so that user can passs in "sudo_user", right?

I think the goal I have in mind (i.e. RemoteAccount change) is kind of independent of your goal (i.e. command line argument change). We can first work on the RemoteAccount change before the command line argument change.

My proposal for RemoteAccount change is this: we include "sudo_user" as optional parameter in the constructor. The sudo_user will be used in the ssh command if it is specified. Does that sound good?

I actually don't think we should specify sudo_user in the command line argument for Ducktape. This is because the tests may very likely involve multiple RemoteAccount instances with different sudo_user or command_prefix value. It is not possible to have a single value of sudo_user or command_prefix in this case.

lindong28:
@granders Do you think we can add `sudo_user` in the constructor of `RemoteAccount`?

lindong28:
@granders @ewencp Do you have time to review this? @granders and I had a discussion on this in the last kafka meetup and I think we agreed that it is useful to be able to specify sudo_user in the RemoteAccount.

ducktape/cluster/remoteaccount.py
```diff
@@ -62,6 +63,8 @@ def _can_ping_url(self, url, headers):
     def ssh_command(self, cmd):
         if self.local:
             return cmd
+        if self.sudo_user is not None:
```
ewencp: Do we always want to apply the sudo user? Not sure I understand all the use cases, but a lot of ssh commands don't need or want sudo. If we include a `sudo_user`, wouldn't it be better to have the caller opt in (and replace all the SSH commands prefixed with `sudo` with regular calls that have a flag like `sudo=True` or `sudo="username"` or something like that)?

lindong28: Are you suggesting to provide `sudo_user` in the `ssh_command(..)` instead of specifying it in the `RemoteAccount` constructor, so that ducktape user can specify `sudo_user` per ssh_command?

I have considered this as well. But I don't think any use case for this yet. In my opinion `RemoteAccount` allows user to run a `cmd` on a given `machine` (i.e. uri or ip) as the given `user`. This `sudo_user` is very similar to the existing `user`, which is not specified per ssh_command. I couldn't find a scenario where we will want to run `ssh_command` with different `sudo_user` value using a given `RemoteAccount` instance.

So my answer is no, we always want to apply the sudo user for a given `RemoteAccount` if it is specified in the constructor. What do you think?

lindong28: I agree with can allow user to specify `sudo_user` per ssh_command to play on the safe side, just in case user will need that. It's just that I don't have a scenario that user will need that yet.

granders: @lindong28 
There is actually a use case for sudo per ssh_command - we do have some system tests which require sudo as root for a few individual commands.

granders:
Hi @lindong28, a couple of further thoughts:
* I still find this to be a nagging concern: `RemoteAccount` instances are only created by clusters, so this PR adds a feature which is not really used anywhere. Would it make more sense to add this along with a proposed change to the cluster?
* The `sudo` implementation still needs to be updated to work with "compound" commands. This isn't a theoretical concern because many existing tests, for example, have "compound" ssh commands for which this implementation wouldn't work.
* Having thought about this more, one odd thing about having distinct "user" and  "sudo user" attributes linked to the same remote account - this means we always ssh in as one user, then execute the command as a different user. In your use-case, is this because you cannot ssh in as a user that would be able to run the commands you need?

ducktape/cluster/remoteaccount.py
lindong28: @granders Sure, I will update the patch to specify sudo_user per ssh_command. I have two questions:

- Do you think it needs to be specified in the RemoteAccount() as well?
- If I understand it right, you have a use case for which you create RemoteAccount with user = A and needs to execute ssh_command with user = root for in some cases and user = A in other cases. If A != root, how to you currently address the problem without this patch? If A == root, then this patch can address the use case without having sudo_user per ssh_command, right?

granders: @lindong28 
- I'm leaning again toward not specifying in RemoteAccount unless there is a clear use-case 
- I think in general we don't want to ssh in as root, and only run as root where necessary; in the few cases where we need increased privileges, we pretty much just directly run: `node.account.ssh("sudo cmd")`

In general, I think my and Ewen's biggest concerns around this patch are that we don't understand exactly your use-case and why it's the most appropriate solution. Would you mind sketching that out?

lindong28: @granders Here are my thoughts:

- Adding it in RemoteAccount is just convenient if we believe the same sudo_user will be used for all ssh_command from this RemoteAccount instance. It is similar to user which is put in RemoteAccount instead of in ssh_command(). It is not necessary and either way works for my use case.
- In fact, the use case here is that we don't want to ssh in as root but we may need to start/stop process as root, which is already demonstrated in your use case. I agree that you can get around by doing `node.account.ssh("sudo cmd")`. But what if I want to do `node.account.kill_process(...)` to kill process as root user?

Simply speaking, the use case is to be able to execute command on remote machine as root user without logging into that machine as root user, using APIs already provided in RemoteAccount, including ssh_command() and kill_process(). Is this reasonable?

I think the solution to the use case above is to be able to specify sudo_user in either function parameter or in the RemoteAccount constructor. Is there better solution to this problem?

granders: @lindong28 Thanks, for your clarifications. One followup question is - what would this enable you to do that you cannot do today? Or, what specific thing is hard for you to do now, that this would make easier?

As for what this looks like api-wise, here's some thinking out loud about what this should maybe look like:

specify sudo per ssh command might look something like:
*  `node.account.ssh(cmd, super_user="userB")` or maybe
*  `node.account.ssh(cmd, sudo=True)` along with similar flag on other remote account methods
  * In this case, `RemoteAccount` instance as `super_user` field which by default is `root`

My take is the second version is better; the troublesome thing about specifying `super_user` in test code, is that the value of `super_user` is probably dependent on the environment in which the tests are run, so I agree with you that it would make sense to tie super_user to the `RemoteAccount` instance.

What do you think?

lindong28: @granders Today if I want to kill a process on a remote machine as root user, I won't be able to do that with either `RemoteAccount.signal(...)` or `RemoteAccount.kill(...)`. And if I want to start a process as root user, I have do explicitly do `ssh_output("sudo -u root " + cmd)`, which is not as nice as being able to do `ssh_output(cmd, sudo_user=True)`. Does this make sense?

I agree with you that the second format, i.e. `node.account.ssh(cmd, sudo=True)`, is better.

granders: @lindong28 Ok sure that makes sense. 

It sounds like we're converging on behavior where:
- `RemoteAccount` can be instantiated with a specified superuser (which defaults to root)
- ssh commands must specify  `sudo_user=True` to trigger this behavior

lindong28: @granders Cool, I am glad that we finally reached an agreement.

I will go ahead to submit a patch as we discussed.

PR #135, opened 2016-03-18
User defined globals

granders:
The primary contribution here is to allow users to specify a set of global key/values that is made available as an immutable dict attached to the shared session context object.

* Users can specify globals as a JSON string, or as a filename which contains JSON
* Included a few unit tests for this

Additional changes:
* Factored option parsing into a self-contained file, and attempted to make the logic a bit more readable
* Fixed an error which before had prevented calling ducktape with multiple test paths (e.g. `ducktape path_a path_b`), and added associated unit test
* Added several associated parser unit tests
* Minor change - passing parsed args as a dict rather than the ArgParse.NameSpace object.
* On the service class, added a base `Service.name` property to use instead of the ad hoc `my_service.__class__.__name__`


granders:
@ewencp @hachikuji Mind taking a look?

ducktape/command_line/defaults.py
```diff
@@ -15,7 +15,7 @@
 import os
 
 
-class ConsoleConfig(object):
+class ConsoleDefaults(object):
```
ewencp: Why the renaming? This doesn't look like it'll impact any other projects since they don't use it, but also seems unimportant -- it wasn't made to make room for something that better fit the `ConsoleConfig` name or anything like that.

ducktape/command_line/main.py
```diff
@@ -102,12 +43,42 @@ def extend_import_paths(paths):
...
+        return None
+
+    try:
+        # First try parsing directly as json
+        user_globals = json.loads(globals_str)
+    except ValueError:
+        # The string is not JSON - try interpreting string as filename instead
+        user_globals = json.loads(open(globals_str, "r").read())
+
+    assert user_globals is not None
```
ewencp: Why the separate assertion? Won't this be caught by the subsequent check for a `dict`.

ducktape/command_line/parse_args.py
```diff
@@ -0,0 +1,115 @@
...
+                             "this root directory.")
+    parser.add_argument("--exit-first", action="store_true", help="exit after first failure")
+    parser.add_argument("--no-teardown", action="store_true",
+                        help="don't kill running processes or remove log files when a test has finished running. "
+                             "This is primarily useful for test developers who want to interact with running "
+                             "services after a test has run.")
+    parser.add_argument("--version", action="store_true", help="display version")
+    parser.add_argument("--parameters", action="store",
+                        help="inject these arguments into the specified test(s). Specify parameters as a JSON string.")
+    parser.add_argument("--globals", action="store",
```
ewencp: Does this mean globals always need to be passed in on the command line? Or is the expectation that this is only for overrides of defaults? Just curious because it seems best to figure out an approach that doesn't require always passing a parameter on the command line.

ducktape/services/service.py
```diff
@@ -69,23 +68,58 @@ def __repr__(self):
                              (self.num_nodes, [n.account.hostname for n in self.nodes]))
 
     @property
+    def name(self):
+        """Human-readable identifier (almost certainly) unique within a test run."""
+        return self.__class__.__name__ + "-" + str(self.id)
```
ewencp: Is it worth adding a way to override this, for example via constructor parameter? Seem like `self.id` is not going to be all that useful, although I guess in a lot of our current cases we only have one service of each type (and that is something of a constraint of ducktape's current model for mapping services to cluster nodes).

ducktape/services/service.py
```diff
@@ -206,16 +240,3 @@ def idx(self, node):
                 return idx
         return -1
 
-    @staticmethod
-    def run_parallel(*args):
```
ewencp: I don't think this is safe to remove -- it's used in Kafka.

tests/command_line/check_main.py
```diff
@@ -51,13 +53,105 @@ def check_symlink(self):
...
+    def check_parseable_json_string(self):
+        globals_dict = get_user_defined_globals(globals_json)
+        assert globals_dict == json.loads(globals_json)
+
+    def check_unparseable(self):
+        """Check for expected error"""
+        try:
+            global_dict = get_user_defined_globals(invalid_globals_json)
+            assert False, "Should have failed to parse invalid JSON"
+        except:
```
ewencp: It's preferable to expect a specific type of exception. See https://pytest.org/latest/assert.html#assertions-about-expected-exceptions Most of these tests could use the same treatment

ducktape/command_line/defaults.py
granders: I renamed since it's essentially internal, and I thought the name better reflected its purpose and the way it is actually used. Could roll back if necessary though

ducktape/command_line/parse_args.py
granders: It means user-defined globals need be either passed in the command line as a filename or a json string, or passed in a config file.

Other flags that may effectively be global are not included here.

Have I actually answered your question here?

ducktape/services/service.py
granders: Maybe `name` is the wrong word for this property, and maybe we don't need both an "id" and a "name", but the idea is it's useful to have unique but human-readable identifier of a service instance

Use cases I had in mind:
* folder into which log files are collected after running a test
  * There is a currently a bug where if a test has multiple instances of a service, the logs get collected underneath the same folder.
  * E.g. mirror maker system tests in kafka
* It is useful having a unique service name when creating a path on remote machines where logs etc should be stored

ducktape/services/service.py
granders: good catch

ducktape/command_line/defaults.py
ewencp: Not necessary, just wasn't clear to me we considered it internal. This is one of the problems w/ Python code -- it's not always obvious what is internal.

ducktape/command_line/parse_args.py
ewencp: I think so, but to clarify: how does a project make use of the globals exactly? Is the expectation that we define defaults for the globals directly in Python code and the ones passed in here (or via config file) are used to override those, e.g. something like

```
my_var = globals.get("my-config", DEFAULT_MY_CONFIG)
```

And then only if you were working in a non-default environment would you need to pass in something for the `--globals` command line config? Basically, I'm trying to make sure `--globals` in the use case you're envisioning is only used in special cases and that, e.g., if I were just running the kafka tests from a standard checkout, I wouldn't need to add a `--globals` flag.

ducktape/services/service.py
ewencp: Right, those use cases seem valid. I was just wondering if a test-specific override could be handy, e.g. rather than getting something like `ConsoleConsumer-0` I could get something like `ConsoleConsumer-ValidateMessages`. (I don't have a good example of where we use the same service multiple times right now, but hopefully you get the idea).

In any case, no need to add the override now either, just a thought.

ducktape/command_line/defaults.py
granders: Yeah good point - the fact that "we consider it to be internal" is not actually enforced

ducktape/command_line/main.py
granders: Removed

ducktape/command_line/main.py
```diff
@@ -102,12 +43,41 @@ def extend_import_paths(paths):
...
-def setup_results_directory(results_root, new_results_dir):
+def get_user_defined_globals(globals_str):
+    """Parse user-defined globals into an immutable dict using globals_str
+
+    :param globals_str Either a file, in which case, attempt to open the file and parse the contents as JSON,
+        or a JSON string representing a JSON object. The parsed JSON must represent a collection of key-value pairs,
+        i.e. a python dict.
+    :return dict containing user-defined global variables
+    """
+    empty_global = pysistence.make_dict()
```
ewencp: Don't create this if you don't need it -- just `return pysistence.make_dict()` in the `if` block

ducktape/command_line/main.py
```diff
@@ -102,12 +43,41 @@ def extend_import_paths(paths):
...
+    if globals_str is None:
+        return empty_global
+
+    try:
+        # First try parsing directly as json
+        user_globals = json.loads(globals_str)
+    except ValueError:
+        # The string is not JSON - try interpreting string as filename instead
+        user_globals = json.loads(open(globals_str, "r").read())
```
ewencp: This probably needs exception handling too? In the case of a bad JSON string you'll get an IOError with a likely confusing `"No such file or directory: '<malformed json here>'"`

ducktape/command_line/main.py
```diff
@@ -102,12 +43,41 @@ def extend_import_paths(paths):
...
+
+    # Now check that the parsed JSON is a dictionary
+    if not isinstance(user_globals, dict):
+        raise ValueError("User defined globals must be a JSON object which parses to a dict. "
```
ewencp: This is misleading since it can also be a filename.

tests/templates/service/check_render.py
```diff
@@ -41,7 +41,7 @@ def check_file_template(self):
 
 class TemplateRenderingService(Service):
     NO_VARIABLE = "fixed content"
-    SIMPLE_VARIABLE = "Hello {{name}}!"
+    SIMPLE_VARIABLE = "Hello {{a_field}}!"
```
ewencp: This change made me realize we might be using `name` as a field in services elsewhere. If we make this change I see at least `connect.py` in kafkatest will need modifications.

ducktape/command_line/parse_args.py
granders: In an overdue response to your question... that's right, there isn't any built in logic for specifying defaults on the ducktape side of the world. 

In practice, in kafka and muckrake, I'm using it as you described - providing defaults in code so that the standard use-case does not require `--globals` flag

tests/templates/service/check_render.py
granders: On second look, having both `name` and `id` when they're conceptually very similar is odd. 
I think I will consolidate into one property called `service_id`

granders:
@ewencp Thanks for feedback so far. Ready for another look.
An example of how this fits with actual system tests:
https://github.com/apache/kafka/pull/1245

ducktape/command_line/main.py
```diff
@@ -102,12 +43,56 @@ def extend_import_paths(paths):
...
+    """
+    if globals_str is None:
+        return pysistence.make_dict()
+
+    from_file = False
+    if os.path.isfile(globals_str):
+        # The string appears to be a file, so try loading JSON from file
+        # This may raise an IOError if the file can't be read or a ValueError if the contents of the file
+        # cannot be parsed.
+        print "globals parameter appears to be a file. Attempting to load user defined globals from file %s..." % globals_str
```
ewencp: debugging code? normally we wouldn't use print in library code like this

ewencp:
@granders One minor comment, otherwise LGTM.

ducktape/command_line/main.py
granders: @ewencp 
There are already a couple of other `print` calls restricted to the `main` module which are only used when it's about to exit since the `session_context` object is not available throughout.

I figure either we add another logger instance here, or just print. Thoughts?

ConfluentJenkins: I thought this was just debugging code though. Do we even need to be
printing this at all?

granders: fair enough; removed

PR #136, opened 2016-03-25
Delete ref to services so memory can be garbage collected

granders:
Fixes memory leak causing occasional out of memory errors in kafka system tests.


granders:
@ewencp The trickiest bit here was actually the unit test which seems to provide a reasonable heuristic.

I'm doing another profiled run of the whole test suite to double-check this fix as well, so it will be worth looking at those results before merging anything.

granders:
before-fix
after-fix

granders:
@ewencp attached above are plots of memory usage (without patch, then with this patch) sampled during full kafka system test runs on my aws test driver node

granders:
In both cases, the big spike is during connect distributed test

ducktape/tests/runner.py
```diff
@@ -87,33 +85,43 @@ def run_all_tests(self):
             # Results from this test, as well as logs will be dumped here
             mkdir_p(self.current_test_context.results_dir)
 
+            result_info = {
```
ewencp: Not really a big deal, but why in a dictionary vs using variables for each of these? Being able to use `kwargs` makes passing the final variables to `TestResult` easy, but every other time you set them pretty awkward since you have to do it through a dict.

ducktape/tests/runner.py
```diff
@@ -195,6 +203,14 @@ def teardown_single_test(self, teardown_services=True):
             exceptions.append(e)
             self.log(logging.WARN, "Error freeing nodes: %s" % e.message + "\n" + traceback.format_exc(limit=16))
 
+        # Remove reference to services. This is important to prevent potential memory leaks if users write services
+        # which themselves have references to large memory-intensive objects
+        try:
+            del self.current_test_context.services
+        except BaseException as e:
```
ewencp: What are we expecting to catch here? Lack of `services` field or something? Seems odd that a `del` like this would be a problem.

tests/runner/check_serial_runner_memory.py
```diff
@@ -0,0 +1,139 @@
...
+        with it.
+        """
+        # Get a list of test_context objects for the test runner
+        ctx_list = []
+        test_methods = [MemoryLeakTest.test_leak]
+        for f in test_methods:
+            ctx_list.extend(MarkedFunctionExpander(session_context=self.session_context, cls=MemoryLeakTest, function=f).expand())
+        assert len(ctx_list) == N_TEST_CASES  # Sanity check
+
+        runner = InstrumentedSerialTestRunner(self.session_context, ctx_list, memory_data_file=self.tmp_file)
```
ewencp: So we're using a file for this to handle the multiprocessing part since you fork this into a separate process? Otherwise using a file seems odd.

ewencp:
@granders A couple of small comments, but up to you whether they need to be addressed -- mostly just questions. Let me know if you need follow up review.

tests/runner/check_serial_runner_memory.py
granders: Yeah it was purely for sharing data between processes - but there are simpler ways to do this.

PR #138, opened 2016-04-27
Add wait_node to service class

### LINKED ISSUE #137: Add thread joining to BackgroundTheadService shutdown routines
We currently require each extension of the BackgroundThreadService to implement thread joining inside the `stop_node` function, but it would be nicer to keep that logic encapsulated. As an example of where exposing this causes trouble, consider the following check which is used in Kafka test services:
```python
if len(self.worker_threads) >= self.idx(node):
    self.worker_threads[self.idx(node) - 1].join()
```
This check is incorrect because BackgroundThreadService doesn't take into account the fact that a test case itself can call `start_node()` and blindly appends a new thread to the `worker_threads` list. We can (and should) fix this problem, but it would be more ideal to keep these details out of extensions.

Note also that this would let us control stop timeouts internally.

hachikuji:
This resolves #137. After discussion, it seemed better to expose the capability wait on a single node rather than adding it to the default implementation of either `stop_node` or `clean_node`.


ewencp:
add to whitelist

ducktape/services/service.py
```diff
@@ -150,6 +150,12 @@ def wait(self):
         """
         pass
 
+    def wait_node(self, node, timeout_sec=None):
```
granders: Do we really want this in the base service class? I was imagining it would belong in `BackgroundThreadService`

ducktape/services/background_thread.py
```diff
@@ -46,6 +46,9 @@ def _protected_worker(self, idx, node):
     def start_node(self, node):
         idx = self.idx(node)
 
+        if idx in self.worker_threads and self.worker_threads[idx].is_alive():
+            raise "Cannot restart node since previous thread is still alive"
```
granders: I think it's better to raise in instance of `RuntimeError` than a string

granders:
Thanks @hachikuji!
My main question is whether we want `wait_node` in the base Service class

ducktape/services/background_thread.py
hachikuji: Ack.

ducktape/services/service.py
hachikuji: I have no strong preference, but it seemed reasonable to add it since we already expose node-specific methods in `Service`.  What do you think?

ewencp: @hachikuji One thing to keep in mind is that adding the method and docs kind of implies that every class needs to implement it. Maybe that's not a problem in this case since you pretty much always need this functionality? But its worth keeping in mind that putting it in the `Service` class basically commits all `Service` implementations to honor the interface. (Although, at a minimum, in this case it seems that if you wanted to let people opt out this is at least a better default impl than raising a `NotImplementedError`.)

hachikuji:
@granders @ewencp Ready for another look.

ducktape/services/service.py
```diff
@@ -173,12 +176,33 @@ def start_node(self, node):
...
+            now = time.time()
+            if end > now:
+                self.logger.debug("%s: waiting for node", self.who_am_i(node))
+                if not self.wait_node(node, end - now):
+                    alive_nodes.append(node)
+            else:
+                alive_nodes.append(node)
+
+        if alive_nodes:
+            raise TimeoutError("Timed out waiting %s seconds for background threads to finish. " % str(timeout_sec) +
```
granders: Make this message slightly more generic now that we've pulled it up to `Service`?

ducktape/services/service.py
```diff
@@ -173,12 +176,33 @@ def start_node(self, node):
         """Start service process(es) on the given node."""
         raise NotImplementedError("%s: subclasses must implement start_node." % self.who_am_i())
 
-    def wait(self):
+    def wait(self, timeout_sec=600):
```
granders: should `Service.run` take an optional timeout now, too?

ducktape/services/service.py
```diff
@@ -173,12 +176,33 @@ def start_node(self, node):
...
         """Wait for the service to finish.
         This only makes sense for tasks with a fixed amount of work to do. For services that generate
         output, it is only guaranteed to be available after this call returns.
         """
-        pass
+        alive_nodes = [] # nodes still alive following completion of wait
```
granders: Hope this isn't too bikesheddy - I have a slight preference for calling it `unfinished_nodes` since we've pulled the logic up to `Service`. Minor point though

granders:
@hachikuji A few comments, but I like the updates

ducktape/services/service.py
hachikuji: I like `unfinished_nodes`. I'll make the change.

ducktape/services/service.py
hachikuji: Ack

ducktape/services/service.py
hachikuji: Ack

ducktape/services/service.py
hachikuji: After some offline discussion, we decided not to do this since it would suggest a timeout for the entire run() operation, which might require timeouts in start() and stop() as well.

granders:
LGTM

granders:
Oops, one thing - can you increment version to `0.5.0`?

PR #140, opened 2016-05-13
Ensure non-standard file descriptors closed in Popen

hachikuji:
This is intended to address an issue which causes `ReplicationVerifiacationToolTest` to hang indefinitely.  Might be jumping the gun a little bit concluding that this is the root problem (though I have run that test case for a couple hours without a reoccurrence), but this change seems to make sense regardless whether it fixes the specific hanging process issue we've encountered in this test. 

Our theory for the cause of the problem, which is based on analysis of an active instance of the hanging test, is the following:

1. `Popen()` is called for a short-lived command (e.g. listing pid files).
2. Before the pipe to this process has been fully read, a separate thread calls `Popen` for a long-running process (in this case `ReplicationVerificationTool`).
3. Because `close_fds` is false by default, this long-running process inherits all file descriptors (including the pipe to the short-lived process).
4. When the short-lived process returns, the pipe remains open because the long-running process still has it open. This causes read() to block which ends up hanging the test runner.

This fix addresses this problem by ensuring that the only files which remain open in subprocesses are STDIN, STDOUT, and STDERR.


ewencp:
@hachikuji LGTM. I don't think we really have the same issue there, but there are also two Popen calls in `ducktape/cluster/vagrant.py`.

granders:
Probably want to increment to `0.5.1` in `ducktape/__init__.py`
Could do the close_fds in Popen thing in vagrant.py, too, but I agree that probably won't make much difference

LGTM

hachikuji:
@granders @ewencp Added the `close_fds` flag in the vagrant popen calls and bumped the version.

granders:
Thanks @hachikuji!

PR #142, opened 2016-06-08
Close log file handles

granders:
File handles opened by loggers currently never get closed. Depending on the system configs, this can result in hitting max file handle limits

This patch closes open file handles after use


granders:
@hachikuji @ewencp Mind taking a look?
I have hit file handle limits a few times when running a large number of tests on my laptop, and `lsof` showed that the majority of these were files opened by test loggers.

ducktape/command_line/main.py
```diff
@@ -171,6 +171,7 @@ def main():
     # Run the tests
     runner = SerialTestRunner(session_context, tests)
     test_results = runner.run_all_tests()
+    session_context.close_logger()
```
ewencp: Would it make sense to just have a `close()` which, if necessary, can handle cleaning up other resources too?

ducktape/tests/runner.py
```diff
@@ -203,6 +203,12 @@ def teardown_single_test(self, teardown_services=True):
             exceptions.append(e)
             self.log(logging.WARN, "Error freeing nodes: %s" % e.message + "\n" + traceback.format_exc(limit=16))
 
+        try:
+            self.current_test_context.close_logger()
+        except BaseException as e:
```
ewencp: How does this happen?

ewencp:
@granders Looks good, just a couple of minor questions.

ducktape/command_line/main.py
granders: Yeah I hear your point. It was a little unclear what `close` ought to mean wrt test_context/session_context

test teardown involves:
- try to stop services associated with test
- collect logs associated with test
- clean processes/logs/scratch space
(Maybe the following reasonably belong in "close"?)
- release service nodes back to cluster
- delete reference to services
- close log handles

ducktape/tests/runner.py
granders: I was being cautious. Cursory internet investigation does not show any errors coming from closing a file though.

ewencp:
@granders This looked nearly complete and seems like it is likely to end up having conflicts with #146. Should we get this merged?

PR #143, opened 2016-06-28
Added cluster annotation and unit tests

granders:
This is an incremental PR with an aim toward parallel test runner
It adds cluster size annotation, but no logic for actually making use of the extra metadata.


granders:
@ewencp @hachikuji Mind taking a look?

hachikuji:
@granders Would it make sense to make the annotation more generic? For example:
```python
@cluster(num_nodes=100)
```
Potentially that might allow us to include additional attributes in the future?

granders:
@hachikuji Fair point; two possible approaches are:
- additional annotations within `marks/resource.py`, e.g. `cluster_size`, `cluster_blah`. I.e., what is currently implemented.
- look for particular keywords underneath `@cluster` annotation. This approach is definitely much more compact if we want to add additional cluster metadata

ducktape/tests/test.py
```diff
@@ -142,32 +142,42 @@ def __init__(self, **kwargs):
...
-        """Construct a new TestContext object from another TestContext object
-        Note that this is not a true copy, since a fresh ServiceRegistry instance will be created.
-        """
+        """Construct a new TestContext object from another TestContext object"""
         ctx_copy = TestContext(**self.__dict__)
         ctx_copy.__dict__.update(**kwargs)
         return ctx_copy
 
     @property
+    def cluster_size(self):
```
ewencp: Can we just get rid of this and use the field directly by accounting for the value missing in the constructor?

tests/mark/check_cluster_size.py
```diff
@@ -0,0 +1,148 @@
...
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+
+from ducktape.mark import parametrize, matrix, ignore
+from ducktape.mark._mark import MarkedFunctionExpander
+from ducktape.mark import resource
```
ewencp: Why not just import `cluster_size` instead of repeating `resource.cluster_size` throughout this file

ewencp:
Agreed about the annotation. This leaves things open for extension, but as long as `size` remains the first arg, you can use `@cluster(100)`. Also, I'd suggest `size` over `num_nodes` :)

hachikuji:
I suggested `num_nodes` because we already use it throughout the codebase, but `size` works too.

ewencp:
Yeah, `num_nodes` is fine for consistency.

@granders compact is probably better if its going to be for other cluster related stuff. The one example I can think of is constraints if we end up with heterogeneous clusters, e.g. by platform (even if generically by matching tags) or available resources (e.g. needs at least n MB RAM). Those get trickier though as they could possibly even vary per service/node, but the simpler we can keep things the better.

granders:
@ewencp @hachikuji Thanks for the suggestions, agree this is more compact/flexible

granders:
@ewencp @hachikuji Updated to be more generic, let me know if you have better ideas around naming

hachikuji:
@granders LGTM on the API. The implementation seems reasonable, but I'm not too familiar with python decorator usage.

tests/mark/check_cluster_use_metadata.py
```diff
@@ -0,0 +1,163 @@
...
+
+        test_context_list = MarkedFunctionExpander(function=f).expand()
+        assert len(test_context_list) == 1
+        assert test_context_list[0].expected_num_nodes == num_nodes_1
+
+    def check_parametrized_with_multiple_cluster_annotations(self):
+        num_nodes_1 = 10
+        num_nodes_2 = 20
+
+        # num_nodes_2 should *not* override num_nodes_1
```
granders: this comment is backwards

tests/mark/check_cluster_use_metadata.py
```diff
@@ -0,0 +1,163 @@
...
+        assert len(test_context_list) == 3
+        assert test_context_list[0].expected_num_nodes == num_nodes_1
+        assert test_context_list[1].expected_num_nodes == num_nodes_1
+        assert test_context_list[2].expected_num_nodes == num_nodes_2
+
+    def check_matrix_with_multiple_cluster_annotations(self):
+        num_nodes_1 = 10
+        num_nodes_2 = 20
+
+        # num_nodes_2 should *not* override num_nodes_1
```
granders: reverse num_nodes_2 and num_nodes_1 in comment

ducktape/mark/resource.py
```diff
@@ -0,0 +1,75 @@
...
+        assert len(context_list) > 0, "cluster use annotation is not being applied to any test cases"
+
+        for ctx in context_list:
+            if not ctx.cluster_use_metadata:
+                # only update if non-None and non-empty
+                ctx.cluster_use_metadata = self.metadata
+        return context_list
+
+
+def cluster(**kwargs):
```
ewencp: do you want to specify `def cluster(num_nodes, **kwargs)`? That allows positional (i.e. `@cluster(5)`) and keyword (`@cluster(num_nodes=5)`) variants.

ducktape/tests/test.py
```diff
@@ -133,6 +134,9 @@ def _escape_pathname(s):
     return re.sub("^\.|\.$", "", s)
 
 
+CLUSTER_SIZE_KEYWORD = "num_nodes"
```
ewencp: nit, but should this be here or defined alongside the decorator? Possibly related to question about positional vs keyword args above.

tests/mark/check_cluster_use_metadata.py
```diff
@@ -0,0 +1,163 @@
...
+
+        test_context_list = MarkedFunctionExpander(function=f).expand()
+        assert len(test_context_list) == 1
+        assert test_context_list[0].expected_num_nodes == num_nodes
+
+    def check_beneath_parametrize(self):
+        """ Annotations such as cluster, which add metadata, but do not create new tests, add the metadata to
+        all test cases physically below the annotation.
+
+        In the case of a parametrized test, when @cluster has no parametrize annotations below it,
```
ewencp: Not sure I understand this explanation. Why wouldn't it just be applied to the implicit unparameterized test and then be propagated to the parametrized version? I would expect this to work for matrix too. For annotations that are really just attaching some property to the test, its pretty unintuitive that ordering would matter.

ducktape/mark/resource.py
granders: My preference is to keep the mapping of keyword -> value explicit - in my mind this keeps the intent of  code using the annotation clearer if we add more keywords

`@cluster(5, another_val=2)`
vs
`@cluster(num_nodes=5, another_val=2)`

ducktape/tests/test.py
granders: I was imagining the concept of expected `num_nodes` as being something defined within/alongside `TestContext`, and having `ClusterUseMetadata` provide a dumb container for key->val pairs.

More practically, it helps avoid circular import since ducktape.mark._mark imports `TestContext`

tests/mark/check_cluster_use_metadata.py
granders: My assumption was that we wanted to support different test cases within a given parametrized test having different expected node usage. It means each individual parametrization can be annotated with metadata independently from the others.

I wanted to keep a consistent model, where
- metadata annotations apply to test cases strictly physically below
- metadata annotations resolve collisions by allowing the annotation physically below to take precedence

This is very similar to how I imagine handling tags on tests.

These rules allow us to attach different metadata to different test cases:
```
@cluster(num_nodes=5)
@parametrize(x=1)
@cluster(num_nodes=3)
@parametrize(x=2)
def test_method(...):
  # for test case x=1, 5 nodes are allocated
  # for test case x=2, 3 nodes are allocated
```

Also, if we allow the reordering you mentioned, it's ambiguous how we should treat a case like this
```
@cluster(num_nodes=5)
@parametrize(x=1)
@parametrize(x=2)
@cluster(num_nodes=3)
def test_method(...):
  ...
```

granders:
This is similar to how `ignore` annotation works today - it allows you to ignore specific test cases within a parametrized test.

The tradeoff to this approach is that any metadata tags applying to all test cases has to go at the very top

```
@ignore
@cluster(num_nodes=5)
@parametrize(...)
@parametrize(...)
@parametrize(...)
def test():
  ...
```

granders:
@ewencp I moved `CLUSTER_SIZE_KEYWORD` from its fishy location

The other comments are more about how the api should should look/be used. I think the approach I've taken is more self consistent and reasonably intuitive for some of the reasons I outlined above, but let me know if you still disagree :)

ducktape/mark/resource.py
ewencp: Sure. The drawback is that there's no validation of the parameters this way, whereas if you explicitly list the parameters you can also get that for free (here we just accept anything, even if its a typo and/or we won't ever use it). I don't feel strongly either way though.

ducktape/tests/test.py
ewencp: Again, no strong preference. Just kinda weird that the docstring explaining where you would actually use this name (as a parameter to the decorator) and the string version of it are defined in separate modules.

tests/mark/check_cluster_use_metadata.py
ewencp: Interesting. Not clear to me that we want/need multiple cluster annotations since it seems like it means you're either using variably-sized services or a toggling services on/off since otherwise you'd have a constant # of nodes per test. Variably-sized services seems like the more likely use case to arise (e.g. some sort of scalability test that ramps up # of nodes), but I'm not sure whether we'll run into that in practice or not.

I'm fine with this as is, just seemed unintuitive that it requires specific placement.

ewencp:
@granders LGTM

tests/mark/check_cluster_use_metadata.py
granders: Yeah, toggling services on/off or variable size services seem the most likely scenario.

PR #144, opened 2016-07-05
Added --compress option for compression of remote service logs

granders:
Allow possibility of compressing service logs before they are copied from remote workers


granders:
@hachikuji @ewencp Mind taking a look?

PR #146, opened 2016-07-14
Spawn subprocess

granders:
PR into parallel-test-runner branch since I plan to do a few more incremental pull requests before merging parallel-test-runner into master.

Tests still run sequentially, but this PR updates ducktape to run tests in subprocesses, using message passing to coordinate starting of new tests.


granders:
@ewencp @hachikuji 
Would you mind doing an initial once-over?

ducktape/command_line/main.py
```diff
@@ -158,18 +159,19 @@ def main():
 
     # Initializing the cluster is slow, so do so only if
     # tests are sure to be run
+    cluster = None
```
ewencp: Why this assignment? Python isn't scoped the same way as other languages, and if an exception occurs you're going to exit anyway.

ducktape/tests/loader.py
```diff
@@ -33,25 +33,31 @@ class LoaderException(Exception):
...
     def parse_discovery_symbol(self, discovery_symbol):
-        """Parse command-line argument into a tuple (directory, module.py, cls_name, function_name).
+        """Parse a single 'discovery symbol'
+
+        :return tuple of form (directory, module.py, cls_name, function_name)
 
         :raise LoaderException if it can't be parsed
+
+        Examples:
+            "path/to/directory" -> ("/absolute/path/to/directory", "", "", "")
```
ewencp: Is the `/absolute/path/to/directory` part of this accurate anymore? You don't have an `abspath` call and `os.path.dirname` does not guarantee an absolute path.

ducktape/tests/loader.py
```diff
@@ -141,7 +173,12 @@ def import_modules(self, file_list):
                 module_name = '.'.join(path_pieces)
                 # Try to import the current file as a module
                 try:
-                    module_list.append(importlib.import_module(module_name))
+                    module_list.append(
+                        {
```
ewencp: Given how this is actually used:

```
        modules_and_files = self.import_modules(test_files)
        for mf in modules_and_files:
            test_context_list.extend(self.expand_module(mf))
```

it might be better to just use tuples, which admittedly leave a bit of the info implicit, but can be clearer and less awkward:

```
        for module,file in self.import_modules(test_files):
            test_context_list.extend(self.expand_module(module, file))
```

ducktape/tests/loader.py
```diff
@@ -192,16 +229,24 @@ def import_modules(self, file_list):
 
         return module_list
 
-    def expand_module(self, module):
+    def expand_module(self, module_and_file):
```
ewencp: This and the dict are kind of awkward. Why not choose either a concrete named type (with expected structure) or separate this into two separate parameters?

It might also help to rename some of these methods to have a `_` prefix so we know when they are internal (and therefore it might not be *as* bad for them to have somewhat weird arguments like this one).

ducktape/tests/message.py
```diff
@@ -0,0 +1,95 @@
...
+
+class Request(object):
+    READY = "READY"  # reply: {test_metadata, cluster, session_context}
+    SETTING_UP = "SETTING_UP"
+    RUNNING = "RUNNING"
+    TEARING_DOWN = "TEARING_DOWN"
+    FINISHED = "FINISHED"
+    LOG = "LOG"
+
+    def __init__(self, source_id):
```
ewencp: I'm not sure I understand what this class is supposed to represent. From its name I thought it had to do with requests being sent between the different processes, but it seems this object only has a `source_id` and each of the methods seems to generate an `event` more like what I was expecting to be sent back and forth?

ducktape/tests/message.py
```diff
@@ -0,0 +1,95 @@
...
+        event = {
+            "source_id": self.source_id,
+            "event_type": event_type,
+            "event_time": time.time()
+        }
+
+        assert len(set(event.keys()).intersection(set(payload.keys()))) == 0, \
+            "Payload and base event should not share keys. base event: %s, payload: %s" % (str(event), str(payload))
+
+        event. update(payload)
```
ewencp: nit: stray space

ducktape/tests/result.py
```diff
@@ -34,41 +33,40 @@ def __str__(self):
...
         """
         @param test_context  standard test context object
         @param test_status   did the test pass or fail, etc?
         @param summary       summary information
         @param data          data returned by the test, e.g. throughput
         """
 
-        self.test_context = test_context
-        self.session_context = self.test_context.session_context
+        self.test_id = test_context.test_id
```
ewencp: Is all this copying to make this serializable or something? Seems like we're just extracting everything out of `test_context`.

ducktape/tests/runner.py
```diff
@@ -14,212 +14,175 @@
...
 
     # When set to True, the test runner will finish running/cleaning the current test, but it will not run any more
     stop_testing = False
 
-    def __init__(self, *args, **kwargs):
-        super(SerialTestRunner, self).__init__(*args, **kwargs)
+    def __init__(self, cluster, session_context, session_logger, tests, port=ConsoleDefaults.TEST_DRIVER_PORT):
+        # session_logger, message logger,
+        self.session_logger = session_logger
+        self.request_logger = RequestLogger(session_context).logger
```
ewencp: Any reason not to just share the session logger?

ducktape/tests/runner.py
```diff
@@ -14,212 +14,175 @@
...
+        if event["event_type"] == Request.READY:
+            self.handle_ready(event)
+        elif event["event_type"] in [Request.RUNNING, Request.SETTING_UP, Request.TEARING_DOWN]:
+            self.handle_lifecycle(event)
+        elif event["event_type"] == Request.FINISHED:
+            self.handle_finished(event)
+        elif event["event_type"] == Request.LOG:
+            self.handle_log(event)
+        else:
+            self.handle_other(event)
```
ewencp: Shouldn't this just be an error since it is unexpected?

ducktape/tests/runner.py
```diff
@@ -14,212 +14,175 @@
...
+            self.serde.serialize(ready_reply(self.session_context, self.current_test_context, self.cluster)))
 
-                self.current_test_context, self.current_test = None, None
+        # Test is "active" once we receive the ready request
+        self.active_tests[event["source_id"]] = event["event_type"]
 
-            if self.stop_testing:
-                break
+    def handle_log(self, event):
+        self.log(event["log_level"], event["message"])
```
ewencp: If you don't do the ack message immediately, you probably want to protect against exceptions in this code. This bit isn't that bad, but you might want to standardize the pattern if you want to just always ack first.

Also, is this log missing some information about who is running it, i.e. the `source_id` I see elsewhere? Seems like that is how you plan to track the active test state properly when you have parallel tests (and presumably also the lifecycle events, which don't seem to do anything right now).

ducktape/tests/runner.py
```diff
@@ -14,212 +14,175 @@
...
+        self.socket.send(self.serde.serialize(event))
```
ewencp: Why are we sending the event back in reply? Looks like we do this repeatedly. Why wouldn't this just be empty or some simple ack?

ducktape/tests/runner_client.py
```diff
@@ -0,0 +1,281 @@
...
+    client = RunnerClient(port, logger_name, log_dir, debug, max_parallel)
+    client.run()
+
+
+class RunnerClient(object):
+    """Run a single test"""
+
+    def __init__(self, port, logger_name, log_dir, debug, max_parallel):
+        self.serde = SerDe()
+        self.logger = TestLogger(logger_name, log_dir, debug, max_parallel).logger
```
ewencp: Why would `max_parallel` ever be a relevant parameter to the `TestLogger`? In the code it looks like maybe you are using this as a proxy for when you want to log debug level to stdout. Shouldn't this just use the `--debug` CLI flag (which may need to be passed through a couple of layers to get here)?

ducktape/tests/runner_client.py
```diff
@@ -0,0 +1,281 @@
...
+        self.runner_port = port
+
+        self.id = "test-runner-%d-%d" % (os.getpid(), id(self))
+        self.request = Request(self.id)
+        self.sender = Sender("localhost", str(self.runner_port), self.logger)
```
ewencp: Shouldn't hostname be passed through the constructor/`run_client` if you eventually want this running on a different host?

ducktape/tests/runner_client.py
```diff
@@ -0,0 +1,281 @@
...
+        self.test = None
+        self.test_context = None
+
+    def send(self, event):
+        return self.sender.send(event)
+
+    def collect_test_context(self, directory, file_name, cls_name, method_name, injected_args):
+        # TODO - different logger for TestLoader object
+        loader = TestLoader(self.session_context, self.logger, injected_args=injected_args, cluster=self.cluster)
+        loaded_context_list = loader._discover(directory, file_name, cls_name, method_name)
```
ewencp: `_discover` seems misnamed since you're using it from another class?

ducktape/tests/runner_client.py
```diff
@@ -0,0 +1,281 @@
...
+        test_context = loaded_context_list[0]
+        test_context.cluster = self.cluster
+        return test_context
+
+    def run(self):
+        self.log(logging.INFO, "Loading test %s" % str(self.test_metadata))
+        self.test_context = self.collect_test_context(**self.test_metadata)
+
+        self.send(self.request.running())
+        if len(self.cluster) != self.cluster.num_available_nodes():
```
ewencp: Strictly speaking I don't think this is always true. `LocalhostCluster` initializes the available nodes to `sys.maxint` and also always returns `sys.maxint` for `len`. JsonCluster uses its explicit list to keep everything in sync.

We may just not have clearly defined the semantics, but I don't think this test will always work. A `DockerCluster` might be another case similar to `LocalhostCluster` where there isn't a fixed size up front, and it could effectively have an infinitely large cluster.

ducktape/tests/runner_client.py
```diff
@@ -0,0 +1,281 @@
...
+            # Sanity check - are we leaking cluster nodes?
+            raise RuntimeError(
```
ewencp: Should this be doing anything to notify the parent process with some `self.send(something)`?

ducktape/tests/runner_client.py
```diff
@@ -0,0 +1,281 @@
...
+                start_time,
+                stop_time)
+
+            self.log(logging.INFO, "Summary: %s" % str(result.summary))
+            self.log(logging.INFO, "Data: %s" % str(result.data))
+
+            test_reporter = SingleResultFileReporter(result)
+            test_reporter.report()
+
+        # Tell the server we are finished
```
ewencp: Again, in the case of test failure we would have hit the above `except` and `finally`, but we're not going to process this bit of code. Intended or is this going to cause the parent process to hang?

ducktape/tests/runner_client.py
```diff
@@ -0,0 +1,281 @@
...
+
+        if not getattr(self, 'test'):
+            msg = "%s: %s" % (self.__class__.__name__, str(msg))
+            self.logger.log(log_level, msg)
+        else:
+            msg = "%s: %s: %s" % (self.__class__.__name__, self.test_context.test_name, str(msg))
+            self.logger.log(log_level, msg)
+            self.test.logger.log(log_level, msg)
+
+        self.send(self.request.log(msg, level=log_level))
```
ewencp: I thought we were just going to be forwarding logs, but now it seems like we potentially log to 3 separate places? Why so many?

ducktape/tests/runner_client.py
```diff
@@ -0,0 +1,281 @@
...
+                        return self.serde.deserialize(reply)
+                    else:
+                        # send another request...
+                        break
+                else:
+                    self.logger.debug("NO-POLLIN")
+                    self.close()
+                    self._init_socket()
+                    waiting_for_reply = False
+                time.sleep(.5)
```
ewencp: Why `time.sleep`? Why not just block indefinitely or set `REQUEST_TIMEOUT_MS` much, much larger? It looks like maybe zmq doesn't do anything in terms of reliability by default. This is pretty concerning... Knowing that failures can occur is good, and having a way to actually handle them gracefully is better, but firing and praying doesn't seem great, especially since I'm not sure all the code on the other half of this connection is idempotent. In fact, the logging code *definitely* isn't idempotent.

ducktape/tests/test.py
```diff
@@ -173,20 +173,74 @@ def _escape_pathname(s):
...
+class TestLogger(Logger):
+    def __init__(self, logger_name, log_dir, debug, max_parallel):
+        self._logger_name = logger_name
+        self.log_dir = log_dir
+        self.debug = debug
+        self.max_parallel = max_parallel
+
+    @property
+    def logger_name(self):
+        return self._logger_name
```
ewencp: When there's no computation going on, is there any reason we're not just setting it as the field name and not including these extra lines of code for wrapping that field?

ducktape/tests/test.py
```diff
@@ -292,41 +360,10 @@ def test_name(self):
...
-            ch.setLevel(logging.DEBUG)
-        else:
-            # default - pipe warning level logging to stdout
-            ch.setLevel(logging.WARNING)
-        self._logger.addHandler(ch)
-
-
+    def logger(self):
+        """Lazily instantiate logger"""
+        return TestLogger(
```
ewencp: It looks like all uses of `TestLogger` are like this -- you construct the class, then immediately just take the `logger` property. That makes me think it should probably be a function, not a class...

setup.py
```diff
@@ -42,7 +42,7 @@ def run_tests(self):
       url="http://github.com/confluentinc/ducktape",
       packages=find_packages(),
       package_data={'ducktape': ['templates/report/*']},
-      install_requires=['jinja2', 'requests', 'pysistence'],
+      install_requires=['jinja2', 'requests', 'pysistence', 'zmq'],
```
ewencp: The `zmq` package seems to suggest you should be using `pyzmq`: https://pypi.python.org/pypi/zmq/0.0.0

tests/runner/check_serial_runner.py
```diff
@@ -12,29 +12,55 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-from ducktape.tests.test import TestContext, Test
-from ducktape.tests.runner import SerialTestRunner
-from ducktape.mark import ignore, parametrize
+from ducktape.tests.test import TestContext
+from ducktape.tests.runner import TestRunner
```
ewencp: filename should probably change given that `SerialTestRunner` no longer exists?

tests/tests/check_session.py
```diff
@@ -12,16 +12,40 @@
...
+    def check_pickleable(self):
+        """Check that session_context object is pickleable
+        This is necessary so that SessionContext objects can be shared between processes when using python
+        multiprocessing module.
+        """
+
+        kwargs = {
+            "session_id": "hello-123",
+            "results_dir": self.tempdir,
+            "cluster": LocalhostCluster(),
```
ewencp: Only issue with this test may be that `LocalhostCluster` is basically the most trivial implementation. Not sure we'll be confident that others would work based on this test.

ducktape/tests/result.py
granders: yeah - Test context isn't currently serializable, so we can't have a reference to a test_context object if we want to serialize TestResult

ducktape/tests/runner.py
granders: Not necessarily; it's more a question of user interaction
i.e. is it useful for debugging to have a separate `request_log` file? Maybe maybe not

ducktape/tests/runner.py
granders: Good call - this was intended to be very temporary - it was just a convenient way to ack without extra effort

ducktape/tests/runner.py
ewencp: It just seemed like we were already putting logs from the test runners into the session/test logger anyway. I figured the number of requests here should be minimal and mostly just bookend the test, right?

ducktape/tests/runner_client.py
granders: The question I was thinking about here is what should the behavior be if debug is enabled when running tests in parallel?

- should we let the test runner client subprocesses compete for lock on stdout and dump all the interleaved messages?
- should we just print session log debug info? This is the less likely use case and tries to keep some sanity to the output.

Either way, the most likely use case for wanting the debug output is when you're debugging your own test(s), and running test(s) serially

ducktape/tests/runner_client.py
granders: yep it'll cause the parent process to hang :)
I plan to add a "liveness" check ("is the subprocess pid still present?") in the event loop. Even with such a check, the parent process will still hang if the test itself hangs and the subprocess never exits.

It's interesting that even to handle the non-parallel case, a liveness check on the subprocesses can help handle certain failures cleanly.

ducktape/tests/runner_client.py
granders: Oops, that' should've been two places (I'll delete `self.test.logger.log(log_level, msg)` since `self.test.logger` and `self.logger` are the same object)

Two places continues the previous logging behavior, where notable events from the "thing running the test" are propagated both the the session_log and the test_log

ducktape/tests/runner_client.py
ewencp: Not sure I even understand the second bullet point. Loggers are thread safe anyway, right? The default behavior will just be the first bullet. In any case, I'm confused how this relates to the `max_parallel` parameter?

ducktape/tests/runner_client.py
granders: good point, will ditch the sleep

I'm not sure we want a big timeout here while we're still running everything on the same machine, or even very many retries. Lack of a response when everything is running on the same machine most likely suggests that the parent process died?

Idempotence of the "server" logic is a good point, but practically speaking, it's not clear to me that this is a case we need to handle gracefully? Given that everything is on the same machine, on the server side I wonder if it's reasonable to treat multiple occurrences of certain events from from the same client client as a fatal error and leave it at that.

ducktape/tests/test.py
granders: ack

ducktape/tests/runner_client.py
ewencp: My concern is more that we're starting with a networking library that doesn't really offer us the semantics we want. For example, why aren't we just using some TCP sockets with some reasonable timeouts and getting nice reliability guarantees. If we have socket disconnections and we expect reliable intra-DC connections to be the norm, it's fine to not handle them particularly gracefully (e.g. kill the test or even the entire test run).

granders:
@ewencp I addressed most of your comments, and added a couple JIRAs for:
- handling duplicate messages from client
- prevent runner hang if runner client dies

ducktape/tests/event.py
```diff
@@ -0,0 +1,150 @@
...
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+import copy
+import os
+import time
+
+
+class ClientEventSupplier(object):
```
ewencp: not sure about in python, but Factory is probably the more common idiom than Supplier

ducktape/tests/event.py
```diff
@@ -0,0 +1,150 @@
...
+    # Types of messages available
+    TYPES = {READY, SETTING_UP, RUNNING, TEARING_DOWN, FINISHED, LOG}
+
+    def __init__(self, test_id, source_id):
+        self.test_id = test_id
+        # id of event source
+        self.source_id = source_id
+        self.event_id = 0
+
+    def event(self, event_type, payload=None):
```
ewencp: is this meant to be public or should it be named `_event`

ducktape/tests/loader.py
```diff
@@ -26,32 +27,103 @@ class LoaderException(Exception):
     pass
 
 
+# A helper container class
+ModuleAndFile = collections.namedtuple('ModuleAndFile', 'module file')
```
ewencp: Interesting, I didn't know you could do `'module file'` instead of `['module', 'file']`

ducktape/tests/runner.py
```diff
@@ -14,216 +14,153 @@
...
+class Receiver(object):
+    def __init__(self, port):
+        self.port = port
+        self.serde = SerDe()
 
-        self.logger.debug("Instantiating " + self.who_am_i())
+        self.zmq_context = zmq.Context()
+        self.socket = self.zmq_context.socket(zmq.REP)
+        self.socket.bind("tcp://*:%s" % str(self.port))
+        # self.poller = zmq.Poller()
```
ewencp: Stray commented code, probably should just be deleted

ducktape/tests/runner.py
```diff
@@ -14,216 +14,153 @@
...
+                self.session_context.debug,
+                self.session_context.max_parallel
+            ])
+        self.proc_list.append(proc)
+        proc.start()
+
+    def handle(self, event):
+        self.log(logging.DEBUG, str(event))
+
+        if event["event_type"] == ClientEventSupplier.READY:
```
ewencp: If the number of events stays small here, this approach works fine and is perfectly readable. If this grows substantially, you might just want to use a dictionary-based dispatch, i.e. you build a static dictionary like:

```
event_handlers = { ClientEventSupplier.READY: self.handle_ready, .... }
```

And then dispatch is just:

```
self.event_handlers[event["event_type"]](event)
```

ducktape/tests/runner.py
```diff
@@ -14,216 +14,153 @@
...
+            print "~" * int(2 * terminal_width / 3)
 
+    def handle_lifecycle(self, event):
+        self.receiver.send(self.event_response.event_response(event))
 
+    def log_event(self, log_level, event):
+        """Log an event received from the client."""
+        pass
+
+    def log(self, log_level, msg):
```
ewencp: Probably doesn't matter here as its just for this class, but if we're proxying for loggers, you might want to provide `*args` and `**kwargs` support as in the `logging` module (https://docs.python.org/3/library/logging.html#logging.Logger.log) since it doesn't require you to do the interpolation step yourself.

ducktape/tests/runner_client.py
```diff
@@ -0,0 +1,283 @@
...
+        self.cluster = ready_reply["cluster"]
+
+        # Wait to instantiate the test object until running the test
+        self.test = None
+        self.test_context = None
+
+    def send(self, event):
+        return self.sender.send(event)
+
+    def collect_test_context(self, directory, file_name, cls_name, method_name, injected_args):
```
ewencp: should probably be private. probably same for `send` above too, as well as some other methods below -- almost none of this class needs to be public. (would probably be good to start following good practice naming private/public fields as well)

ducktape/tests/test.py
```diff
@@ -173,20 +173,79 @@ def _escape_pathname(s):
...
+        formatter = logging.Formatter(ConsoleDefaults.TEST_LOG_FORMATTER)
+        info_fh.setFormatter(formatter)
+        debug_fh.setFormatter(formatter)
+
+        self._logger.addHandler(info_fh)
+        self._logger.addHandler(debug_fh)
+
+        ch = logging.StreamHandler(sys.stdout)
+        ch.setFormatter(formatter)
+        if self.debug and self.max_parallel == 1:
```
ewencp: Still concerned about why we're doing this. If the user asks for debug, seems like they should get debug level output... `max_parallel` shouldn't have any impact on their request.

ducktape/tests/test.py
```diff
@@ -292,41 +365,7 @@ def test_name(self):
...
-            ch.setLevel(logging.DEBUG)
-        else:
-            # default - pipe warning level logging to stdout
-            ch.setLevel(logging.WARNING)
-        self._logger.addHandler(ch)
-
-
+    def logger(self):
+        """Lazily instantiate logger"""
+        return test_logger(
```
ewencp: This isn't just lazy, it actually reinstantiates if on every call. I suspect you want to cache this once it is constructed.

ewencp:
@granders A bunch more minor cleanup comments, but LGTM

ducktape/tests/test.py
granders: Ack, ok that's reasonable

granders:
Thanks for the feedback @ewencp

PR #147, opened 2016-07-21
Added cluster methods for requesting and freeing subclusters

granders:
These methods will be used for preallocation of a subcluster before the run of a given test when running tests in parallel.

### DIFF
--- a/ducktape/cluster/cluster.py
+++ b/ducktape/cluster/cluster.py
@@ -38,7 +38,19 @@ class Cluster(object):
         """Size of this cluster object. I.e. number of 'nodes' in the cluster."""
         raise NotImplementedError()
 
-    def request(self, nslots):
+    def request_subcluster(self, num_nodes):
+        """Return an instance of Cluster with the specified num_nodes.
+
+        All implementations of this method within ducktape make use of the num_nodes parameter, but in some
+        conceivable implementations, it may be reasonable to ignore.
+        """
+        raise NotImplementedError()
+
+    def free_subcluster(self, subcluster):
+        """Free all nodes allocated to subcluster back to the original cluster."""
+        raise NotImplementedError()
+
+    def request(self, num_nodes):
         """Request the specified number of slots, which will be reserved until they are freed by the caller."""
         raise NotImplementedError()
 
--- a/ducktape/cluster/cluster.py
+++ b/ducktape/cluster/cluster.py
@@ -22,15 +22,10 @@ class ClusterSlot(object):
             setattr(self, k, v)
 
     def __eq__(self, other):
-        if other is None:
-            return False
-
-        return self.__dict__ == other.__dict__
+        return other is not None and self.__dict__ == other.__dict__
 
     def __hash__(self):
-        # convert self.__dict__ to a hashable type
-        TupleRepresentation = collections.namedtuple('TupleRepresentation', [k for k in self.__dict__.keys()])
-        return hash(TupleRepresentation(**self.__dict__))
+        return hash(tuple(self.__dict__.items()))
 
 
 class Cluster(object):
@@ -43,14 +38,6 @@ class Cluster(object):
 
     def __len__(self):
         """Size of this cluster object. I.e. number of 'nodes' in the cluster."""
-        raise NotImplementedError()
-
-    def alloc_subcluster(self, num_nodes):
-        """Try to reserve num_nodes from this cluster, and return a cluster composed only of those nodes."""
-        raise NotImplementedError()
-
-    def free_subcluster(self, cluster):
-        """Free all nodes allocated to subcluster back to the original cluster."""
         raise NotImplementedError()
 
     def alloc(self, num_nodes):
@@ -77,7 +64,7 @@ class Cluster(object):
         raise NotImplementedError()
 
     def __eq__(self, other):
-        if other is None:
-            return False
+        return other is not None and self.__dict__ == other.__dict__
 
-        return self.__dict__ == other.__dict__
+    def __hash__(self):
+        return hash(tuple(self.__dict__.items()))
--- a/ducktape/cluster/cluster.py
+++ b/ducktape/cluster/cluster.py
@@ -16,14 +16,22 @@ import collections
 
 
 class ClusterSlot(object):
-    def __init__(self, parent, account, **kwargs):
-        self.parent = parent
+    def __init__(self, account, **kwargs):
         self.account = account
         for k, v in kwargs.items():
             setattr(self, k, v)
 
-    def free(self):
-        self.parent.free(self)
+    def __eq__(self, other):
+        if other is None:
+            return False
+
+        return self.__dict__ == other.__dict__
+
+    def __hash__(self):
+        # convert self.__dict__ to a hashable type
+        TupleRepresentation = collections.namedtuple('TupleRepresentation', [k for k in self.__dict__.keys()])
+        return hash(TupleRepresentation(**self.__dict__))
+
 
 
 class Cluster(object):
@@ -68,3 +76,9 @@ class Cluster(object):
 
     def free_single(self, slot):
         raise NotImplementedError()
+
+    def __eq__(self, other):
+        if other is None:
+            return False
+
+        return self.__dict__ == other.__dict__
--- a/ducktape/cluster/cluster.py
+++ b/ducktape/cluster/cluster.py
@@ -25,7 +25,7 @@ class ClusterSlot(object):
         return other is not None and self.__dict__ == other.__dict__
 
     def __hash__(self):
-        return hash(tuple(self.__dict__.items()))
+        return hash(tuple(sorted(self.__dict__.items())))
 
 
 class Cluster(object):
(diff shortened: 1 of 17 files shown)


granders:
@ewencp @hachikuji 
Fairly small incremental patch into parallel-test-runner branch
The main question on my mind is whether this is a reasonable api

One alternative which might be cleaner:
* Instead of adding new methods to the cluster class, change behavior of `request` and `free` to return subclusters.
* Make a tiny corresponding adjustment to the base `Service` class where nodes are requested, and to cleanup logic, where nodes are freed.

ducktape/cluster/cluster.py
```diff
@@ -38,7 +46,19 @@ def __len__(self):
         """Size of this cluster object. I.e. number of 'nodes' in the cluster."""
         raise NotImplementedError()
 
-    def request(self, nslots):
+    def request_subcluster(self, num_nodes):
+        """Return an instance of Cluster with the specified num_nodes.
+
+        All implementations of this method within ducktape make use of the num_nodes parameter, but in some
```
hachikuji: This comment is kind of weird. You mean ducktape does not actually depend on the returned cluster having enough nodes?

ducktape/cluster/cluster.py
```diff
@@ -38,7 +46,19 @@ def __len__(self):
         """Size of this cluster object. I.e. number of 'nodes' in the cluster."""
         raise NotImplementedError()
 
-    def request(self, nslots):
+    def request_subcluster(self, num_nodes):
```
hachikuji: Would `alloc_subcluster` fit nicer with `free_subcluster`?

granders: I was following the `request` convention we established for getting nodes.

These methods are only really used internally, so it wouldn't have impact to change both `request` and `request_subcluster` to `alloc` and `alloc_cluster` for example

ducktape/cluster/cluster.py
granders: Hm, yeah I can remove this part of the comment.

I was imagining a cluster implementation where the size of the cluster is unbounded, aka effectively infinite.

ducktape/cluster/cluster.py
```diff
@@ -16,14 +16,21 @@
...
-        self.parent = parent
+    def __init__(self, account, **kwargs):
         self.account = account
         for k, v in kwargs.items():
             setattr(self, k, v)
 
-    def free(self):
-        self.parent.free(self)
+    def __eq__(self, other):
+        if other is None:
```
ewencp: This implementation is fine, but given that its repeated it might be worth pointing out that you can just do `return other is not None and self.__dict__ == other.__dict__`.

ducktape/cluster/cluster.py
```diff
@@ -16,14 +16,21 @@
...
+            return False
+
+        return self.__dict__ == other.__dict__
+
+    def __hash__(self):
+        # convert self.__dict__ to a hashable type
+        TupleRepresentation = collections.namedtuple('TupleRepresentation', [k for k in self.__dict__.keys()])
```
ewencp: This is going to create the type every time this method is called. If you want `TupleRepresentation`, you should move this out of the method (and probably use `_TupleRepresentation` instead as its private). An alternative implementation would just be `return hash(tuple(x.items()))`

ducktape/cluster/cluster.py
```diff
@@ -38,21 +45,39 @@ def __len__(self):
         """Size of this cluster object. I.e. number of 'nodes' in the cluster."""
         raise NotImplementedError()
 
-    def request(self, nslots):
-        """Request the specified number of slots, which will be reserved until they are freed by the caller."""
+    def alloc_subcluster(self, num_nodes):
```
ewencp: Hmm, this sounds like a convenience around `alloc`. Why wouldn't there be a generic implementation of it?

ducktape/cluster/json.py
```diff
@@ -52,33 +79,52 @@ def __init__(self, cluster_json=None, *args, **kwargs):
...
         return len(self.available_nodes) + len(self.in_use_nodes)
 
     def num_available_nodes(self):
         return len(self.available_nodes)
 
-    def request(self, nslots):
-        if nslots > self.num_available_nodes():
+    def alloc_subcluster(self, num_nodes):
+        nodes = self.alloc(num_nodes)
+        return JsonCluster(
```
ewencp: Do we need each subcluster to actually use the same cluster type? I figured we'd just `alloc` from the cluster, then use the result to create a generic cluster since they all just need the same key properties (as in `_account_to_dict`. Then the subcluster part isn't anything special from the perspective of the `Cluster` implementation, you still just `alloc` and `free` sets of nodes.

I think only the freeing step is what you need the original set of nodes for. In that case, couldn't we just make whatever class implements `Subcluster` accept the original set of nodes in its constructor so `free_subcluster` below can use the original set rather than "allocating" them back from the subcluster?

ducktape/cluster/remoteaccount.py
```diff
@@ -39,6 +40,20 @@ def __str__(self):
         r += self.hostname
         return r
 
+    def __repr__(self):
+        return str(self.__dict__)
+
+    def __eq__(self, other):
```
ewencp: Same comments on `__eq__` and `__hash__` implementations as above

ducktape/tests/runner.py
```diff
@@ -64,95 +64,120 @@ def __init__(self, cluster, session_context, session_logger, tests, port=Console
 
         self.proc_list = []
         self.staged_tests = tests
+        self.test_id_to_test_context = {t.test_id: t for t in tests}
```
ewencp: Might be easier to just reduce these maps to a single state object for each test. Not really critical now, but at a minimum `self._test_state[test_id]` is probably clearer than these verbose names and it looks like in at least one or two places we index into both of these dicts, one right after the other.

ducktape/tests/runner.py
```diff
@@ -64,95 +64,120 @@ def __init__(self, cluster, session_context, session_logger, tests, port=Console
...
+        Side effect: store association between the test_id and the preallocated subcluster.
+
+        :param test_context
+        :return None
+        """
+
+        expected_num_nodes = test_context.expected_num_nodes
+        if expected_num_nodes is None:
+            # If there is no information on cluster usage, allocate entire cluster
+            expected_num_nodes = len(self.cluster)
```
ewencp: log something? If we're using this runner, at least if it has parallelism > 1, then we probably want to tell the user they need to annotate their tests to get parallelism

ducktape/tests/runner.py
```diff
@@ -64,95 +64,120 @@ def __init__(self, cluster, session_context, session_logger, tests, port=Console
...
-        self.finished_tests[event["test_id"]] = event
+        # Transition this test from running to finished
+        test_id = event["test_id"]
+        del self.active_tests[test_id]
+        self.finished_tests[test_id] = event
         self.results.append(event['result'])
 
+        # Free nodes used by the test
+        self.cluster.free_subcluster(self.test_id_to_cluster[test_id])
+        del self.test_id_to_cluster[test_id]
```
ewencp: Should the `test_id_to_test_context` entry also be cleaned up?

granders: no, it's just a straight mapping from `test_id -> test_context`

ducktape/cluster/remoteaccount.py
granders: Regarding the suggestion to use`tuple(self.__dict__.items())`: are we guaranteed that the order of `self.__dict__.items()` will be consistent every time it is called?

granders:
@ewencp Thanks for the discussion. Updated to keep things a bit simpler, ready for another look.

ducktape/cluster/remoteaccount.py
ewencp: No, but you can sort them if you need that. Your current implementation doesn't guarantee ordering either, does it? `keys()` has the same ordering limitations as `items()`.

ducktape/cluster/finite_subcluster.py
```diff
@@ -0,0 +1,51 @@
...
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+from ducktape.cluster.cluster import Cluster
+
+
+class FiniteSubcluster(Cluster):
+    """This cluster class gives us a mechanism for allocating finite blocks of nodes from another cluster.
+    """
+    def __init__(self, node_list):
```
ewencp: nit: why not just nodes? naming for parameters matters since the caller can always use named parameters

ducktape/cluster/json.py
```diff
@@ -52,33 +79,43 @@ def __init__(self, cluster_json=None, *args, **kwargs):
...
 
     def __len__(self):
         return len(self.available_nodes) + len(self.in_use_nodes)
 
     def num_available_nodes(self):
         return len(self.available_nodes)
 
-    def request(self, nslots):
-        if nslots > self.num_available_nodes():
+    def _account_to_dict(self, account):
```
ewencp: No longer used, can remove this

ewencp:
@granders A couple more nits and we need to figure out the ordering thing for the dicts, but otherwise LGTM

PR #148, opened 2016-07-25
Run tests in parallel

granders:


granders:
@ewencp This adds:
- simple "on-demand" scheduling mechanism 
- actual parallelism
- signal handling for cleaning up runner and runner client processes

ducktape/tests/runner.py
```diff
@@ -61,46 +68,98 @@ def __init__(self, cluster, session_context, session_logger, tests, port=Console
...
     def who_am_i(self):
         """Human-readable name helpful for logging."""
         return self.__class__.__name__
 
+    @property
+    def _ready_to_trigger_more_tests(self):
+        """Should we pull another test from the scheduler?"""
+        return not self.stop_testing and \
+            len(self.active_tests) < self.max_parallel and \
+            len(self.scheduler) > 0 and \
```
ewencp: The `len(self.scheduler) > 0` check is baked into the subsequent `peek` check, so you can skip this step if you get rid of the RuntimeError

ducktape/tests/runner.py
```diff
@@ -61,46 +68,98 @@ def __init__(self, cluster, session_context, session_logger, tests, port=Console
...
+
+        for proc in self._client_procs.values():
             proc.join()
 
         return self.results
 
+    def expected_num_nodes(self, test_context):
+        """Helper method for deciding how many nodes we expect the given test to use."""
+        expected = test_context.expected_num_nodes
+        return len(self.cluster) if expected is None else expected
```
ewencp: Thinking about this a bit more -- would it make sense to have  separate setting for max cluster size? I'm wondering how we're going to manage transitioning to having the cluster size annotations. If we actually allocated the whole cluster then we basically can't turn on parallel tests until we've got everything annotated. If we use some other value which can be set via command line flag or env var, then we can set a cluster size that matches today's cluster size given serial execution but still get parallelism without wasting the entire cluster's resources. Thoughts?

ducktape/tests/runner.py
```diff
@@ -122,17 +182,15 @@ def _preallocate_subcluster(self, test_context):
         :param test_context
         :return None
         """
-
-        expected_num_nodes = test_context.expected_num_nodes
-        if expected_num_nodes is None:
+        expected = self.expected_num_nodes(test_context)
+        if test_context.expected_num_nodes is None and self.max_parallel > 1:
```
ewencp: `max_parallel` is checked again immediately below this line

ducktape/tests/scheduler.py
```diff
@@ -0,0 +1,81 @@
...
+                                         key=lambda tc: self._expected_nodes[tc.test_id],
+                                         reverse=True)
+
+    def peek(self):
+        """Locate and return the next object to be scheduled, without removing it internally.
+
+        :return test_context for the next test to be scheduled
+        :raise RuntimeError if the scheduler is empty
+        """
+        if len(self) == 0:
```
ewencp: Since this can return `None` anyway, is there really any benefit to throwing the exception?

ducktape/tests/scheduler.py
```diff
@@ -0,0 +1,81 @@
...
+    def next(self):
+        """Get the next test.
+
+        This action removes the test_context object from the scheduler.
+        """
+        tc = self.peek()
+        self._test_context_list.remove(tc)
+        return tc
+
+    def put(self, test_context):
```
ewencp: I think this is unused

ducktape/tests/scheduler.py
granders: Ah the eternal exception vs sentinel value debate :)

In this case I think we need both, at least as currently written, because I'm distinguishing between these cases:

- `None` means: "scheduler has tasks, but there aren't enough available cluster nodes for any of these tasks"
- exception means: scheduler is empty

An alternative could be to have distinct exceptions for the two cases

ducktape/tests/runner.py
granders: ack

ducktape/tests/scheduler.py
granders: ack

ducktape/tests/runner.py
granders: Good point

Could have max cluster usage setting which would effectively say: no test should use more than this many nodes.

ducktape/tests/runner.py
granders: See my comment in the peek check

ewencp: Yeah, my point was just that since you only ever use `peek()` having already checked `len(self.scheduler) > 0` (as here), the exception is never actually used and you could just remove that code. In any case, either approach works, returning `None` in both cases would just require moving the check of `len(self.scheduler)` to the `while` loop condition.

ducktape/tests/runner.py
ewencp: Shouldn't this use the value from the new command line flag now?

ewencp: Or rather, it seems `TestContext.expected_num_nodes` may handle it so the conditional is not needed.

ducktape/tests/runner.py
```diff
@@ -156,30 +221,40 @@ def _handle_ready(self, event):
...
+
+        if self._should_print_delimiter:
             terminal_width, y = get_terminal_size()
-            print "~" * int(2 * terminal_width / 3)
+            self._log(logging.INFO, "~" * int(2 * terminal_width / 3))
+
+    @property
+    def _should_print_delimiter(self):
+        return self.session_context.max_parallel == 1 and \
+            not self.stop_testing and \
```
ewencp: Do you really want to filter on `stop_testing`? Won't that just sort of cut off some of the output for a test that's in progress? In fact, is it even relevant if you've already checked for `max_parallel == 1` since that condition guarantees there's only 1 test running at a time and if you marked `stop_testing` that must mean it already failed?

ewencp:
@granders Had a couple of minor comments, but LGTM

ducktape/tests/runner.py
granders: This exists to make sure the twiddly separator appears only between test output, and not after the last test output has been printed. But yeah it can be simplified a bit

PR #149, opened 2016-07-28
ST-133 fix

granders:
Fixed string formatting on log message and added regression check


granders:
Ping @ewencp 
Hopefully a quick review. 
Also added a unit test which which fails without the fix

ewencp:
LGTM

PR #150, opened 2016-08-09
use paramiko for ssh

granders:


granders:
Ping @ewencp WIP, but far enough along to get feedback

Vagrantfile
```diff
@@ -22,7 +22,7 @@ VAGRANTFILE_API_VERSION = "2"
 # General config
 enable_dns = false
 num_workers = 3
-ram_megabytes = 1280
+ram_megabytes = 300
```
ewencp: Hmm, wasn't this size determined by JVM heap size for Kafka or something?

ducktape/cluster/remoteaccount.py
```diff
@@ -12,15 +12,46 @@
...
+        self.msg = msg
+
+    def __str__(self):
+        return "%s: %s" % (self.account_str, self.msg)
+
+
+class RemoteCommandError(RemoteAccountError):
+    """This exception is raised when a process run by ssh*() returns a non-zero exit status.
+    """
+    def __init__(self, account, cmd, exit_status, remote_err_msg):
```
ewencp: nit: `remote_err_msg` is pretty verbose where `msg` conveys basically the same info given the context

ducktape/cluster/remoteaccount.py
```diff
@@ -30,7 +61,99 @@ def __init__(self, hostname, user=None, ssh_args=None, ssh_hostname=None, extern
...
+        self._logger = logger
+
+    def _log(self, level, msg, *args, **kwargs):
+        msg = "%s: %s" % (str(self), msg)
+        self.logger.log(level, msg, *args, **kwargs)
+
+    @property
+    def ssh_config(self):
+        if not self._ssh_config:
+            self._ssh_config = self._parse_ssh_opts()
```
ewencp: Can we just do this in the constructor?

ducktape/cluster/remoteaccount.py
```diff
@@ -30,7 +61,99 @@ def __init__(self, hostname, user=None, ssh_args=None, ssh_hostname=None, extern
...
+        return self._ssh_config
+
+    @property
+    def ssh_client(self):
+        if not self._ssh_client:
+            o = self.ssh_config.lookup(self.hostname)
```
ewencp: What exactly is `o`?

ducktape/cluster/remoteaccount.py
```diff
@@ -30,7 +61,99 @@ def __init__(self, hostname, user=None, ssh_args=None, ssh_hostname=None, extern
...
+
+            client = SSHClient()
+            client.set_missing_host_key_policy(AutoAddPolicy())
```
ewencp: Do we actually want to add it? Adding host keys for ec2 is actually kind of annoying because over time they collect and since ec2 will recycle the host names over time, eventually you get host key conflicts. I put this in my `~/.ssh/config` to avoid this:

```
Host *.compute.amazonaws.com
  StrictHostKeyChecking no
  UserKnownHostsFile /dev/null
  LogLevel QUIET
```

A quick look suggests paramiko doesn't have a policy that just accepts the host key but doesn't add it, but maybe we could just implement a new `MissingHostKeyPolicy` that does that?

ducktape/cluster/remoteaccount.py
```diff
@@ -30,7 +61,99 @@ def __init__(self, hostname, user=None, ssh_args=None, ssh_hostname=None, extern
...
+            self._ssh_client = None
+        if self._sftp_client:
+            self._sftp_client.close()
+            self._sftp_client = None
+
+    def _parse_ssh_opts(self):
+        if self.ssh_args is None:
+            return SSHConfig()
+
+        args = self.ssh_args
```
ewencp: Doesn't necessarily have to happen here, but as long as we're moving away from the command line SSH, maybe we can make SSH args provided in a more flexible format from the Cluster implementations. In particular, I think we're actually undoing some of the construction of the ssh command that we do in VagrantCluster.

ducktape/cluster/remoteaccount.py
```diff
@@ -30,7 +61,99 @@ def __init__(self, hostname, user=None, ssh_args=None, ssh_hostname=None, extern
...
+        args = args.split("-o")
+        args = [a.strip() for a in args]
+        args = [a.replace("'", "") for a in args]
+        args = [a.replace("\"", "") for a in args]
+        args = [a.replace("\\", "") for a in args]
+        args = [a for a in args if len(a) > 0]
+
+        args_dict = {"Host": self.hostname}
+        for a in args:
```
ewencp: This entire block could just be `dict(a.split(' ') for a in args)` and then it also catches anything that has too many space-separated elements.

ducktape/cluster/remoteaccount.py
```diff
@@ -30,7 +61,99 @@ def __init__(self, hostname, user=None, ssh_args=None, ssh_hostname=None, extern
...
+             pair = a.split(' ')
+             args_dict[pair[0]] = pair[1]
+        ssh_info_lines = ["%s %s" % (k, v) for k, v in args_dict.iteritems()]
+
+        f = tempfile.NamedTemporaryFile(delete=False)
```
ewencp: You probably just need something file-like here, not an actual file, since this just ends up as `config.parse(fd)` below. If `config.parse()` doesn't work with a `string` try the `StringIO` module.

ducktape/cluster/remoteaccount.py
```diff
@@ -82,40 +205,92 @@ def ssh_command(self, cmd):
         return r
 
     def ssh(self, cmd, allow_fail=False):
+        """Run the given command on the remote host, and block until the command has finished running.
```
ewencp: While we're rethinking APIs, I'd spend some time thinking about the right set of SSH primitives. As it stands today, we have a few different patterns. There's probably some baseline SSH method we can provide that provides good flexibility, then a small number of utilities built on top of that (and ideally better named than our current set). Originally when this started out I thought we'd be able to refactor easily, but we never got around to doing that before we split up the fmwk and

ducktape/cluster/remoteaccount.py
```diff
@@ -82,40 +205,92 @@ def ssh_command(self, cmd):
...
-        subprocess.CalledProcessError. If allow_fail is True, returns the exit
-        status of the command.
-        """
-        return self._ssh_quiet(self.ssh_command(cmd), allow_fail)
+        self._log(logging.DEBUG, "Running ssh command: %s" % cmd)
+
+        client = self.ssh_client
+        stdin, stdout, stderr = client.exec_command(cmd)
+
+        exit_status = stdin.channel.recv_exit_status()
```
ewencp: I don't think this is safe if the output is too large. There's a warning in the Paramiko docs about potentially hanging if the amount of data left on `stdin` is larger than its window.

ducktape/cluster/remoteaccount.py
```diff
@@ -82,40 +205,92 @@ def ssh_command(self, cmd):
...
+        """
+        self._log(logging.DEBUG, "Running ssh command: %s" % cmd)
+
+        client = self.ssh_client
+        stdin, stdout, stderr = client.exec_command(cmd)
 
         def output_generator():
-            for line in iter(proc.stdout.readline, ''):
+
+            for line in iter(stdout.readline, ''):
```
ewencp: This also doesn't seem like it'll be safe since stderr could get too large and block stdout (at least that is my reading of how `Channel`s work). Though it is a bit hard to tell since the docs & examples aren't very clear on correct usage. It seams like you need to use the `Channel`'s `recv_ready` and `recv_stderr_ready` and `send_ready` methods to avoid blocking, although I'm not sure you combine these APIs to be able to read stdout/stderr efficiently.

Note that this has also changed the semantics of this command. Previously it combined stderr into stdout. The current implementation will just swallow all stderr output unless there's a non-zero exit status. I'm not sure if anything relies on this behavior, but we're definitely swallowing log information from some applications that can be useful (you can print to stderr without the application failing).

ducktape/cluster/remoteaccount.py
```diff
@@ -82,40 +205,92 @@ def ssh_command(self, cmd):
...
+        :raise RemoteCommandError If allow_fail is False and the command returns a non-zero exit status, raises
+            RemoteCommandError.
+        """
+        self._log(logging.DEBUG, "Running ssh command: %s" % cmd)
+
+        client = self.ssh_client
+        stdin, stdout, stderr = client.exec_command(cmd)
+
+        try:
+            stdoutdata = stdout.read()
```
ewencp: same comment about stderr here.

ducktape/cluster/remoteaccount.py
```diff
@@ -142,97 +317,182 @@ def kill_process(self, process_grep_str, clean_shutdown=True, allow_fail=False):
         for pid in pids:
             self.signal(pid, sig, allow_fail=allow_fail)
 
-    def scp_from_command(self, src, dest, recursive=False):
```
ewencp: We might reasonably remove the `scp_x_command` methods, but are we actually safe removing the `scp_from` and `scp_to` methods? Seems like these are used and are public API? Window of support & deprecation? Or are we just going to sed when we update the dependency version in kafkatest and muckrake?

ducktape/cluster/remoteaccount.py
```diff
@@ -142,97 +317,182 @@ def kill_process(self, process_grep_str, clean_shutdown=True, allow_fail=False):
...
+    def copy_between(self, src, dest, dest_node):
+        """Copy src_path to dest_path on dest_node
 
-        if isinstance(src, basestring):
-            src = remotehost + ":" + src
-        else:
-            # assume src is iterable
-            # e.g. "ubuntu@host:path1 ubuntu@host:path2"
-            src = " ".join([remotehost + ":" + path for path in src])
+        :param src_path Path to the file or directory we want to copy
```
ewencp: parameter names don't match

ducktape/cluster/remoteaccount.py
```diff
@@ -142,97 +317,182 @@ def kill_process(self, process_grep_str, clean_shutdown=True, allow_fail=False):
...
+        finally:
+            if os.path.isdir(temp_dir):
+                shutil.rmtree(temp_dir)
+
+    def copy_from(self, src, dest):
+        if os.path.isdir(dest):
+            # dest is an existing directory, so assuming src looks like path/to/src_name,
+            # in this case we'll copy as:
+            #   path/to/src_name -> dest/src_name
+            src_name = src
```
ewencp: This code is repeated a bunch of times, could be refactored to make each implementation simpler.

ducktape/cluster/remoteaccount.py
```diff
@@ -142,97 +317,182 @@ def kill_process(self, process_grep_str, clean_shutdown=True, allow_fail=False):
...
+            # for obj in `ls src`, if it's a file, copy with copy_file_from, elif its a directory, call again
+            for obj in self.sftp_client.listdir(src):
+                obj_path = os.path.join(src, obj)
+                if self.isfile(obj_path) or self.isdir(obj_path):
+                    self.copy_from(obj_path, dest)
+                else:
+                    # TODO what about uncopyable file types?
+                    pass
+
+    def copy_to(self, src, dest):
```
ewencp: If we're renaming these, it might make sense to just unify copying vaguely like scp does, e.g. something like

`def copy(self, src_host, src, dest_host, dest)`

Where the two host parameters can be some special value to indicate localhost. Or maybe there's a way to shorten it to use fewer parameters (e.g. a single src command, but if it is `string` it is a filename if it is `(host, filename)` it is remote).

ducktape/cluster/remoteaccount.py
```diff
@@ -142,97 +317,182 @@ def kill_process(self, process_grep_str, clean_shutdown=True, allow_fail=False):
...
-        r += self.hostname + ":" + dest_dir
-        return r
+    def islink(self, path):
+        try:
+            # stat should follow symlinks
+            path_stat = self.sftp_client.lstat(path)
+            return stat.S_ISLNK(path_stat.st_mode)
+        except:
+            # TODO figure out which errors are legit (e.g. if file does not exist, what will sftp.stat do?)
+            is_file = False
```
ewencp: `is_file` seems superfluous in the next few methods -- just `return` instead

ducktape/cluster/remoteaccount.py
```diff
@@ -142,97 +317,182 @@ def kill_process(self, process_grep_str, clean_shutdown=True, allow_fail=False):
...
 
-    def rsync_to(self, flags, src_dir, dest_dir):
```
ewencp: The rsync variants actually appear to be unused. Are we sure about removing them entirely? The benefit of having them is that rsync can be *a lot* faster if the files are mostly synced already.

ducktape/cluster/remoteaccount.py
```diff
@@ -142,97 +317,182 @@ def kill_process(self, process_grep_str, clean_shutdown=True, allow_fail=False):
...
+            f.write(contents)
+
+    def mkdir(self, path, mode=0755):
+
+        self.sftp_client.mkdir(path, mode)
+
+    def mkdirs(self, path, mode=0755):
+        self.ssh("mkdir -p %s && chmod %o %s" % (path, mode, path))
+
+    def remove(self, path, allow_fail=False):
```
ewencp: Hmm, I know we use `allow_fail` elsewhere (something I'm not ecstatic about since it's kinda wordy), but with this a `force` would be way more intuitive since I'm so used to the `-f` on `rm`.

ducktape/cluster/remoteaccount.py
```diff
@@ -142,97 +317,182 @@ def kill_process(self, process_grep_str, clean_shutdown=True, allow_fail=False):
...
+        """Remove the given file or directory"""
+
+        if not self.exists(path):
```
ewencp: Most of this seems better done by just executing the command and seeing if it is successful / adding the `-f` flag to it and then checking the exit status....

ducktape/cluster/vagrant.py
```diff
@@ -62,6 +62,10 @@ def __init__(self, *args, **kwargs):
             cluster_json["nodes"] = nodes
             json.dump(cluster_json, open(cluster_file, 'w+'), indent=2, separators=(',', ': '), sort_keys=True)
 
+        # Release any ssh clients used in querying the nodes for metadata
```
ewencp: Not sure I understand why we need this -- what would have opened an SSH connection at this point?

ducktape/tests/logger.py
```diff
@@ -38,3 +38,12 @@ def configured(self):
 
     def configure_logger(self):
         raise NotImplementedError("configure_logger property must be implemented by a subclass")
+
+    @staticmethod
+    def close_logger(logger):
```
ewencp: Why is this a static instead of just a member variable? Just to avoid the `if logger is not None` at the call site?

ducktape/tests/runner.py
```diff
@@ -132,24 +137,30 @@ def run_all_tests(self):
...
+                    except Exception as e:
+                        err_str = "Exception receiving message: %s: %s" % (str(type(e)), str(e))
+                        err_str += "\n" + traceback.format_exc(limit=16)
+                        self._log(logging.ERROR, err_str)
+
+                        # All processes are on the same machine, so treat communication failure as a fatal error
+                        raise
+            except KeyboardInterrupt:
+                # If SIGINT is received, stop triggering new tests, and let the currently running tests finish
+                self.stop_testing = True
```
ewencp: you might want to log something here too -- it'll at least let them know that you're working on shutting things down.

ducktape/tests/runner_client.py
```diff
@@ -157,7 +158,7 @@ def run(self):
     def setup_test(self):
         """start services etc"""
         self.log(logging.INFO, "Setting up...")
-        self.test.setUp()
+        self.test.setup()
```
ewencp: This seems like a pretty drastic and incompatible change that will affect a lot of tests?

ducktape/tests/test.py
```diff
@@ -52,12 +54,22 @@ def min_cluster_size(self):
         """
         return self.test_context.services.num_nodes()
 
-    def setUp(self):
+    def setup(self):
```
ewencp: Why are we making this change? I thought we had discussed previously the noun vs verb usage? Does anyone besides py.test use the noun form? I know jUnit and Python's unittest module both use the verb form.

And if we are going this path, then a [deprecation warning](https://docs.python.org/2/library/warnings.html) is warranted. Otherwise people have to dig into the code to understand what they are supposed to be overriding.

ducktape/tests/test.py
```diff
@@ -260,6 +273,7 @@ def __init__(self, **kwargs):
         self.log_collect = {}
 
         self._logger = None
+        self._scratch_dir = None
```
ewencp: What's the context for this change? It doesn't seem related to this patch at all and it is only used by tests. Why is this on the test rather than on services? Do the tests often have local temp data they need to work with?

ducktape/tests/test.py
```diff
@@ -380,3 +401,16 @@ def logger(self):
...
         return self._logger
 
+    def close(self):
+        """Release resources, etc."""
+        # Remove reference to services. This is important to prevent potential memory leaks if users write services
+        # which themselves have references to large memory-intensive objects
+        del self.services
+
+        # Remove local scratch directory
+        if self._scratch_dir and os.path.exists(self.scratch_dir):
```
ewencp: Mixing `_scratch_dir` and `scratch_dir`? If this needs to know about the internal field and that it may be `None`, just use it consistently.

systests/cluster/test_remote_account.py
```diff
@@ -61,13 +64,326 @@ def write_to_log(self, msg):
...
+
+class FileSystemTest(Test):
+    """
+    Note that in an attempt to isolate the file system methods, validation should be done with ssh/shell commands.
+    """
+
+    def setup(self):
+        self.service = GenericService(self.test_context, 1)
+        self.node = self.service.nodes[0]
+        self.scratch_dir = self.service.scratch_dir
```
ewencp: I am confused about this test -- I thought `Test` was defining `scratch_dir` as a property so it would get created on demand but here we're overwriting it to be the fixed value in the `GenericService` above?

ducktape/tests/test.py
granders: I suppose verb form would be `set_up`, but the camelCase (`setUp`, `tearDown`) has bugged me for ages.

ducktape/tests/test.py
granders: oops, ack

systests/cluster/test_remote_account.py
granders: These tests predate the addition of a generally available scratch directory
Will change

ducktape/tests/runner_client.py
granders: I attempted (and perhaps succeeded?) to make this backwards compatible:

- If `setUp` is overridden, internally we call `setup`, which will call the user's `setUp`
- If `setup` is overridden, internally we will call the user's `setup` method

Vagrantfile
granders: These are system tests of ducktape itself, so there's no jvm, just smallish bash commands on the remote workers

ducktape/cluster/remoteaccount.py
granders: ack

ducktape/cluster/remoteaccount.py
granders: ack

ducktape/cluster/remoteaccount.py
granders: ack - renamed `config_dict`

ducktape/cluster/remoteaccount.py
granders: Ok, looks like we can effectively ignore missing host key with a new `MissingHostKeyPolicy`

ducktape/tests/test.py
ewencp: It's actually originally inherited from the unittest module: https://docs.python.org/2/library/unittest.html#unittest.TestCase.setUp which afaik violates PEP styling in favor of matching the extremely widespread styling of jUnit and most xUnit variants.

ducktape/tests/runner_client.py
ewencp: Yeah, I figured that out once I got down to the Test class. See comment re: deprecation given overloads.

Vagrantfile
ewencp: Ack, duh, just carryover from another Vagrantfile of course.

ducktape/cluster/remoteaccount.py
granders: ack - although I'm removing these usages (of `scp_*`) from kafkatest, I've added back with deprecated warning

ducktape/cluster/remoteaccount.py
granders: ack

ducktape/cluster/remoteaccount.py
granders: ack

ducktape/cluster/remoteaccount.py
granders: hm... I see your point, but this serves to remove both files and directories. Removing a directory without prompts requires `-f` no matter what, in which case a `force` option is superfluous, and we won't get an error from the exit status

ducktape/cluster/remoteaccount.py
granders: Then it becomes effectively a static method, and I think it would be irritating to import `RemoteAccount` to use it.

If it's a method, you're effectively pinning `src_host` or `dest_host` to `self`, but then you need a modifier e.g. `to`/`from`/`between` to clarify which is which.

ducktape/cluster/remoteaccount.py
granders: See my other comment re `force`

ducktape/cluster/remoteaccount.py
granders: Yeah I agree this is a lot faster in the right situation. This is mostly practical - there aren't currently known uses, and I think it will take a notrivial amount of time and experimentation to get it to work w/paramiko.

ducktape/cluster/remoteaccount.py
ewencp: Wait, why does removing directories require `-f`? Why not just `-r`? `-d` is limited to empty directories. `-f` says "ignore nonexistent files and arguments, never prompt" but I think the cases you cover are handled fine just by `rm -r`?

granders: Ack - yeah I checked again and you're right
I have `rm` aliased with `rm -i` so my intuition for the behavior here was off

ewencp: Haha, awesome. Kind of funny since it is probably intended as a safety net, but might also create a habit of using `-r` unnecessarily :)

ducktape/cluster/remoteaccount.py
granders: In this case, my reading is that `stdout.readline()` would return empty string iff there is nothing more to read, in which case it ought to be safe here?

ducktape/cluster/remoteaccount.py
granders: Added `combine_stderr` argument

ducktape/cluster/remoteaccount.py
ewencp: @granders Not sure how that makes it safe? My point was that you wont hit EOF, so it won't return, but then a big chunk of data on stderr could fill the entire buffer such that the connection can't get anymore stdout data across the channel. So readline won't return and you'll never have a chance to process the stderr data as you are blocked.

granders: ack

granders:
@ewencp Addressed a bunch of comments, ping for another review

ducktape/command_line/defaults.py
```diff
@@ -46,7 +46,8 @@ class ConsoleDefaults(object):
     # Log this to indicate a test is misbehaving to help end user find which test is at fault
     BAD_TEST_MESSAGE = "BAD_TEST"
 
-    # Test driver listens on this port for requests
-    TEST_DRIVER_PORT = 5556
-
+    # Ducktape will try to pick a random open port in the range [TEST_DRIVER_PORT_MIN, TEST_DRIVER_PORT_MAX]
+    # this range is *inclusive*
+    TEST_DRIVER_MIN_PORT = 5556
+    TEST_DRIVER_MAX_PORT = 5586
```
ewencp: This range seems small. 30 may be the # of hosts we use now, but I expect that number to get quite a bit larger.

ducktape/cluster/remoteaccount.py
edenhill: Was this issue resolved? I'm seeing stdout capture hangs when there's lots of stderr output

ewencp: I think it was unclear whether it was perfectly safe (partly because paramiko docs are unclear), but we never hit the issue in practice. Sounds like you have hit it, so we should probably adjust this to use the non-blocking methods.

ducktape/tests/runner_client.py
edenhill: There does seem to be an assumption in the code that stderr is only used for minimal fatal error messages (I think I saw some place that only read stderr if the process exited abnormally), but in practice stderr might be used for debugging and thus have enormously huge output.

ewencp: Right, the reason there was still some uncertainty is because it was unclear exactly how the separate streams running over the same SSH connection would work, i.e. if one ultimately could end up blocking things up such that another can't make any progress. If that does appear to be a problem now, we can definitely fix it in ducktape.

PR #152, opened 2016-10-05
Parallel test runner

granders:


granders:
Ping @ewencp 
This should all be stuff you've seen before in the previous PRs into this feature branch

@shikhar If you wouldn't mind taking a look at `remote_account.py`, that would be great

PR #153, opened 2016-11-21
Windows changes

alexlod:
**IMPORTANT:** NOT MEANT TO BE MERGED.

This is an intermediate commit that I want to get @ewencp's feedback on before I go too far down the rabbit hole.

Ewen, I can't do much on this until the week of 12/5, so no rush reviewing.

The two main components of this are:
 * I introduced a `LinuxRemoteAccount` and `WindowsRemoteAccount` that inherit from `RemoteAccount` Almost all of `LinuxRemoteAccount` is copy-pasted from the old `RemoteAccount`
 * I introduced the concept of a `node_spec` which is a dict indicating how many windows and linux VMs are required

Currently, ducktape tests are passing except for `check_runner_memory.py` which blocks indefinitely (so I commented out the test). The muckrake mini test does not run correctly.


ewencp:
add to whitelist

alexlod:
Very helpful review, @ewencp! Thanks! Unfortunately I won't have a chance to address this until the week of 12/5. But I'll pick back up then.

ducktape/cluster/remoteaccount.py
```diff
@@ -139,8 +75,29 @@ def __init__(self, ssh_config, externally_routable_ip=None, logger=None):
...
+
+        if ssh_config.host and "windows" in ssh_config.host:
+            return WindowsRemoteAccount(ssh_config=ssh_config,
+                                        externally_routable_ip=externally_routable_ip)
+        else:
+            return LinuxRemoteAccount(ssh_config=ssh_config,
+                                      externally_routable_ip=externally_routable_ip)
+
+    def has_operating_system(self, operating_system):
+        # import here to avoid a circular import. TODO: Is there a better way to do this?
```
alexlod: I like your latter suggestion, but it feels to me that the feature isn't needed today and would introduce complexity. Although maybe the complexity feels larger to me because I'm rusty in Python ;). If it's ok with you, I'll get around the circular dependency with your first suggestion and leave the latter suggestion for a future improvement when it's needed.

alexlod:
@ewencp: I applied all of your suggestions into a new commit, except for the one suggestion I commented on. Very good feedback, thanks!

Would you prefer I close this PR and open a new one when I've finished all the work? Or would you prefer I continue pushing to this PR?

Thanks again!

ducktape/cluster/cluster.py
```diff
@@ -62,3 +66,20 @@ def __eq__(self, other):
...
+            if node.operating_system == operating_system:
+                count += 1
+        return count
+
+    @staticmethod
+    def _next_available_node(nodes, operating_system):
+        node_to_return = None
+        for node in nodes:
+            if node.operating_system == operating_system:
+                node_to_return = node
```
ewencp: You could just return this immediately instead. Currently you'll iterate over the entire collection even if the first node you check matches. (You can also write these in a simpler more pythonic way, e.g. `return next(node for node in nodes if node.operating_system == operating_system)` for `_next_available_node` and `_node_count_helper` could be `return len([node for node in nodes if node.operating_system == operating_system])`.)

ducktape/services/service.py
```diff
@@ -51,11 +52,17 @@ class Service(TemplateRenderer):
     # }
     logs = {}
 
-    def __init__(self, context, num_nodes, *args, **kwargs):
+    # TODO: I changed this from num_nodes to nodes but I may not have updated all places where this constructor is called.
```
ewencp: It's actually difficult to validate this since this is user facing code, right? One option to maintain compatibility would be to keep the num_nodes argument and convert it to the new node spec if it is used. It would be safe to convert it to `{'linux':num_nodes}` since Linux support was all that was previously provided.

Without the compatibility we'd probably actually want to bump the version # to 0.7 since it'd be an incompatible change. To maintain compatibility we'd probably want to add the `nodes` argument with a default value.

ewencp: Oh, and now I see that we do the conversion below. I've lost track of previous discussions around this since they happened awhile ago. I guess maybe the compatibility won't be a real issue if we never invoke this constructor with that parameter as a keyword argument.

ducktape/cluster/remoteaccount.py
ewencp: It's even fine to leave it as is if it keeps things simpler and we're not worried about it doing something bad, e.g. introducing a performance issue.

ewencp:
@alexlod Continuing in the same PR seems fine to me. But if you'd prefer to merge and work on follow up PRs, we can stick this on a feature branch and target the PRs there until the entire feature set is ready to be merged to master.

ducktape/services/service.py
alexlod: I could also just leave `num_nodes`.  You tell me what's best here -- no change, change the param name, or bump the version to `0.7.0`. I'm flexible.

alexlod:
@ewencp: working in a feature branch sounds good, because then new PRs will be easier to read. How about this:

 * You review my new commits in this PR. The current state is unit tests are passing, and the muckrake minitest is passing. So it's in a pretty good (still incomplete) place.
 * After your review, I'll close this PR and open a new one, with a single commit, against a feature branch. We'll merge that new PR
 * From here on out, I'll submit new PRs on the feature branch

Cool? Thanks!

ducktape/services/service.py
ewencp: It doesn't look critical (we use a named parameter for `num_nodes` to the `@cluster` annotation, but I haven't found it in the implementation of services). But let's avoid the breaking change since we caught it. I was actually thinking of this solution:

```
def __init__(self, context, num_nodes=None, nodes=None, *args, **kwargs):
```

and then checking that one of `num_nodes` or `nodes` is not `None`. I think that keeps compatibility but lets people use `nodes` as a named parameter for the new node spec.

ducktape/cluster/finite_subcluster.py
```diff
@@ -12,17 +12,16 @@
...
 
-class FiniteSubcluster(Cluster):
+class FiniteSubcluster(MixedOsCluster):
     """This cluster class gives us a mechanism for allocating finite blocks of nodes from another cluster.
     """
     def __init__(self, nodes):
         self.nodes = nodes
-        self._available = set(self.nodes)
-        self._in_use = set()
+        self.available_nodes = set(self.nodes)
```
ewencp: Are these meant to be public? I don't see them used outside this class. In Python, you prefix with an `_` to indicate a field is intended to be private to the class. We haven't followed this strictly but have gradually been trying to migrate to this approach over time.

ducktape/services/service.py
alexlod: Sounds good. I'll change `nodes` to `node_spec` too.

ducktape/cluster/mixed_os_cluster.py
```diff
@@ -0,0 +1,45 @@
...
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+from .cluster import Cluster
+from .remoteaccount import RemoteAccount
+
+
+class MixedOsCluster(Cluster):
```
ewencp: Hmm, why have this in addition to the `Cluster` class? Couldn't we just assume all clusters are potentially mixed clusters?

ducktape/tests/result.py
```diff
@@ -48,7 +48,7 @@ def __init__(self,
             self.nodes_used = test_context.services.num_nodes()
         else:
             self.services = {}
-            self.nodes_used = 0
+            self.nodes_used = {"": 0}
```
ewencp: Is the key here the OS? If so, shouldn't this string be non-empty? Or just be an empty dictionary?

ewencp:
@alexlod Sounds good. You don't need to open a new PR, if you want it as a single commit you can just use the Squash and Merge option when merging.

alexlod:
Oh cool! But I don't want this PR to be merged into master. So I'll close this PR, rebase the commits to one, then push to a feature branch. And all new PRs will be against that feature branch.

ducktape/cluster/finite_subcluster.py
alexlod: Yeah these should be private. I just spaced about this. Good catch.

ducktape/cluster/mixed_os_cluster.py
alexlod: I introduced the new class because the methods I introduced don't make sense for the various test cluster types. But I suppose that doesn't make much of a difference. I'll move them to `Cluster`.

ducktape/tests/result.py
alexlod: Well, when I use `nodes_used` below, I ignore the key. So the key doesn't matter. However, I realize that's confusing, so I'll make the key LINUX, which is the default elsewhere. That will be less confusing.

ducktape/cluster/cluster.py
alexlod: Nice Python tricks! Thanks for the pointer.

alexlod:
@ewencp: last chance to comment on this PR. I'll close it and squash my commits in the next hour or so.

alexlod:
Closing this. All future PRs will be against this feature branch: https://github.com/confluentinc/ducktape/tree/windows

ewencp:
@alexlod New PR is fine but FYI, you can now also edit the target branch for PRs -- just hit the edit button at the top of the PR.

alexlod:
Good to know for next time. Thanks @ewencp!

PR #154, opened 2016-11-26
Add support for statically selecting subsets of tests to run.

ewencp:


ducktape/command_line/parse_args.py
```diff
@@ -58,6 +58,12 @@ def create_ducktape_parser():
                         help="Upper bound on number of tests run simultaneously.")
     parser.add_argument("--repeat", action="store", type=int, default=1,
                         help="Use this flag to repeat all discovered tests the given number of times.")
+    parser.add_argument("--subsets", action="store", type=int, default=1,
```
raghavgautam: We are assuming that all the subsets are taking approximately same execution time. This is a very strong assumption.
I have seen wide skews. To give you an example:
Here 18 tests ran in 36 mins.
https://travis-ci.org/raghavgautam/kafka/jobs/178960665
And here 6 tests ran in 32 mins.
https://travis-ci.org/raghavgautam/kafka/jobs/178960668
Probably it will be better to allow users to run a range of tests say from 1-15, 15-20, 20-45 etc.

ducktape/tests/loader.py
```diff
@@ -80,7 +86,15 @@ def load(self, test_discovery_symbols):
                 raise LoaderException("Didn't find any tests for symbol %s." % symbol)
 
         self.logger.debug("Discovered these tests: " + str(all_test_context_list))
-        return all_test_context_list * self.repeat
+
+        # Select the subset of tests. Select every nth test instead of blocks of n to avoid groups of tests that are
+        # parametrizations of the same test being grouped together since that can lead to a single, parameterized,
+        # long-running test causing a very imbalanced workload across different subsets. Imbalance is still possible,
+        # but much less likely using this heuristic.
+        subset_test_context_list = list(itertools.islice(all_test_context_list, self.subset, None, self.subsets))
```
raghavgautam: How do we know that the order of tests in the list will remain same ? The =os.walk()= call provides no such guarantees. 
https://docs.python.org/2/library/os.html#os.walk
https://github.com/confluentinc/ducktape/blob/master/ducktape/tests/loader.py#L287
Probably we need to do sort to ensure this.

ewencp: Good point

ducktape/command_line/parse_args.py
ewencp: The idea was to keep this simple enough that it can be maintained easily. I realize there is variation in the test run times, that's why I listed out the times [in this comment](https://github.com/apache/kafka/pull/2172#issuecomment-263045897) and pointed out that if you select every Nth test instead of blocks of N, you avoid grouping together paramatrizations of the same test method that are likely to take similar (potentially long) amounts of time. Given how the test run times currently look, I'm not convinced we can't get a good distribution using a simple scheme like this.

The idea was to make it more likely you'll get an even distribution without having to manually maintain groupings of tests that are highly specific to the setup you're running the tests on. Using explicit ranges like you mention has a serious drawback as well -- when you add new tests, it's very easy to accidentally forget to update this mapping and then those tests won't even be running.

I'm actually not that happy about adding this as a fixed set of parameters to ducktape itself rather than, e.g., making the test runner pluggable or something like that so we have the flexibility to change this moving forward. But my biggest concern is that we do something that doesn't require constant maintenance and updates to the breakdown of tests into groups. If that means saving info about test run times and allowing the test runner to take advantage of that when breaking tests into groups, that'd be significantly preferable to anything that requires manually maintaining lists of tests...

raghavgautam: If I use your data and split it into 15 splits, the runtimes vary from 1710 sec to 2566 sec.
`
>>> for i in range(15): sum(itertools.islice(data, i, None, 15))
...
2566.3769999999995
2150.7529999999997
2244.764
2363.585
2320.717
2524.1629999999996
2092.3149999999996
2101.9869999999996
2063.3659999999995
2141.384
1718.5469999999998
1894.3680000000004
1710.8980000000001
1831.2150000000001
2200.502
`
I think that a completely automated solution of this problem is not clear at this point. I am inclined towards giving the end user as much control as possible and let them figure out how they want to do splits. (of course the user may choose to use data to generate splits - he may write a script to automate this split generation)

That said, if you feel that the Nth approach will work, let us go ahead with it and we can tackle the problem when we face it.

raghavgautam:
@ewencp It will good if we can conclude the discussion. I am going on vacation day on Thursday after which my availability will be intermittent.

ewencp:
@raghavgautam How about this as a compromise: I can update this to optionally use an extra file that provides timing information from previous test runs. It can just be a URL that we need to point to. By also outputting timing information with each test run, we'll have an easy way to update the timing info when it gets too outdated. If no timing information is available (the URL arg is not provided), then we just fall back to the current assumptions. Despite being a bit more complex than I'd like, I think this will work well because:

1. We don't need any of that timing information just to get things started. The fallback mode just works and is easy to use.
2. We don't have constant maintenance. We'll want to update that timing information every once in awhile, but just adding a new test doesn't require doing anything. Falling back to using, e.g., the average known test time will probably work for keeping things relatively balanced for many months at a time (just based on the rate at which we tend to add tests).
3. Doesn't require any weird customization of the test layout, which at the moment was already leading people to ask whether they could put a test in a specific directory (i.e. should I be putting it in `clients`, `clients1`, `clients2` or even creating a new `clients3`?)
4. While we don't have the flexibility in the travis.yml file atm (as far as I can tell from their docs), this doesn't tie us to any particular value for the level of parallelism. If someone wants to use the same support to run the tests with larger parallelism on their own infrastructure, it's easy for them to do so.

What do you think? If this works, I can try to get support for this in pretty quickly.

raghavgautam:
Sounds good.

ewencp:
@aayars Can you take a quick look at this too?

ducktape/tests/loader.py
```diff
@@ -15,9 +15,12 @@
 import collections
 import importlib
 import inspect
+import itertools
 import os
 import re
+import requests
 
+from operator import attrgetter
```
aayars: Nitpick: Move "from operator..." to appear above built-in python imports

PR #160, opened 2017-03-17
ST-18 Ducktape framework documentation

rancp:
@ewencp @apovzner @norwood


ewencp:
add to whitelist

docs/install.rst
```diff
@@ -0,0 +1,31 @@
+.. _topics-install:
+
+=======
+Install
+=======
+
+.. note:: As a general rule, it's recommended to use an isolation tool such as ``virtualenv``.
```
ewencp: I would recommend including the virtualenv steps in the install instructions. It's pretty much universally a good idea, so we want to encourage folks to do it by default. This note is still helpful (gives the opt-out if they don't want to use it), but for anyone unfamiliar with python, we should encourage them to work in a properly isolated env. Those familiar enough to make the choice themselves will be able to cut out the steps they don't need.

docs/install.rst
```diff
@@ -0,0 +1,31 @@
...
+
+        $ sudo yum install gcc libffi-devel python-devel openssl-devel
+
+* Install ducktape::
+
+    pip install ducktape
+
+
+
+* On macOS you may need to ``brew install openssl`` and then install using::
```
ewencp: Hmm, I actually added this note to the README and I'm surprised I didn't just put it up with the OSX section. Maybe consolidate. Unfortunately, I think it depends on the version of macOS, though maybe at this point we should just assume people are on the latest version?

docs/misc.rst
```diff
@@ -0,0 +1,59 @@
...
+Alternatively, if you've installed pytest ``sudo pip install pytest`` you can run
+it directly on the ``tests`` directory::
+
+    py.test tests
+
+
+Windows
+=======
+
+Ducktape supports Service`s that run on Windows, but only in EC2.
```
ewencp: nit: there's a backtick in `Service's` instead of an apostrophe that's mucking up with formatting

docs/misc.rst
```diff
@@ -0,0 +1,59 @@
...
+
+When a ``Service`` requires a Windows machine, AWS credentials must be configured on the machine running ducktape.
+
+Ducktape uses the [boto3](https://aws.amazon.com/sdk-for-python/) Python module to connect to AWS. And ``boto3`` supports many different [configuration options](https://boto3.readthedocs.io/en/latest/guide/configuration.html#guide-configuration).
```
ewencp: I think this link doesn't work because of the previous incorrect backtick

docs/misc.rst
```diff
@@ -0,0 +1,59 @@
...
+
+Here's an example bare minimum configuration using environment variables::
+
+    export AWS_ACCESS_KEY_ID="ABC123"
+    export AWS_SECRET_ACCESS_KEY="secret"
+    export AWS_DEFAULT_REGION="us-east-1"
+
+The region can be any AWS region, not just ``us-east-1``.
+
+Contribute
```
ewencp: I think these contribute and license sections are accidentally copied from the index file.

docs/new_services.rst
```diff
@@ -0,0 +1,41 @@
+.. _topics-new_services:
+
+New Services
+============
+
+"Service" refers generally to any process, possibly long-running, which you
```
ewencp: More importantly, it's *multiple* processes. The expectation for a `Service` is that it might be a distributed system, so we by default expect multiple processes/nodes.

docs/new_services.rst
```diff
@@ -0,0 +1,41 @@
...
+want to run on the test cluster.
+
+These can be services you would actually deploy
+(e.g., Kafka brokers, ZK servers, REST proxy) or processes used during testing
+(e.g. producer/consumer performance processes). You should also make each
```
ewencp: Rather than "you should", saying something like "Services that are distributed systems can support a variable number of nodes, allowing one implementation to handle a variety of tests" might give Service authors a bit more flexibility. In at least a few cases, there really isn't any way to support more than one instance at a time.

docs/new_services.rst
ewencp: nit: Adjust the quoting of Service so it's styled as code

docs/new_services.rst
```diff
@@ -0,0 +1,41 @@
...
+service class support starting a variable number of instances of the service so
+test code is as concise as possible.
+
+Each service is implemented as a class and should at least implement the following::
+
+    start_node(self, node) - start the service (possibly waiting to ensure it started successfully)
```
ewencp: We might be missing docstrings atm, but linking to the actual docs might be nice here

docs/new_services.rst
```diff
@@ -0,0 +1,41 @@
...
+
+Most of the code you'll write for a service will just be series of SSH commands
+and tests of output. You should request the number of nodes you'll need using
+the ``num_nodes`` or ``node_spec`` parameter to the Service base class's constructor.
+Then, in your Service's methods you'll have access to ``self.nodes`` to access the
+nodes allocated to your service. Each node has an associated
+``ducktape.cluster.RemoteAccount`` instance which lets you easily perform remote
+operations such as running commands via SSH or creating files. By default, these
+operations try to hide output (but provide it to you if you need to extract
+some subset of it) and *checks status codes for errors* so any operations that
```
ewencp: formatting issue here, probably md vs rst issue

docs/new_tests.rst
```diff
@@ -0,0 +1,53 @@
+.. _topics-new_tests:
+
+=========
+New Tests
+=========
+
+Writing ducktape Tests
+======================
+
+Subclass ``ducktape.tests.test.Test and implement`` as many ``test`` methods as you
```
ewencp: nit: backticks are around extra words "and implement" here

docs/new_tests.rst
```diff
@@ -0,0 +1,53 @@
...
+        def test_services_start(self):
+            self.zk.start()
+            self.kafka.start()
+
+Test Parameters
+===============
+
+Use test decorators to parametrize tests, examples are provided below
+
+.. autofunction:: ducktape.mark._mark.parametrize
```
ewencp: In `__init__.py` we import the stuff from `_mark` such that it will be available without the underscore, e.g. via `ducktape.mark.parametrize` instead of `ducktape.mark._mark.parametrize`. Since we don't really want to use the latter import, will the `autofunction` handle the shorter, cleaner version here?

docs/requirements.txt
```diff
@@ -0,0 +1,10 @@
+avro
```
ewencp: Why do we need a `requirements.txt`? Shouldn't the `setup.py` already handle this? Is this required for the readthedocs integration or something? I just want to avoid duplicating dependency information if possible.

docs/run_tests.rst
```diff
@@ -0,0 +1,145 @@
...
+
+Options
+=======
+
+To see a complete listing of options run::
+
+    ducktape --help
+
+
+Here is a list of run options:
```
ewencp: Is this manually generated or auto-generated? Just curious what our maintenance cost will be as we add more config options.

docs/requirements.txt
rancp: Agreed, just noticed there's an option for RTD to install the project, removing this dependency now

docs/install.rst
rancp: Good point, I think we should add the suggestion as a step and let them choose what tool they want to use since there are alternatives to virtualenv

docs/new_services.rst
```diff
@@ -0,0 +1,21 @@
+.. _topics-new_services:
+
+New Services
```
apovzner: Maybe call it "Writing a new service" or "Writing new services"? Somehow, "New Services" title was confusing to me.

apovzner:
LGTM. 

I am ok with merging this and then expand some sections with other PRs.

README.md
```diff
@@ -1,219 +1,24 @@
...
-
-Use the `--collect-only` flag to discover tests without running any:
-
-    ducktape <path_to_testfile_or_directory> --collect-only
-
-Use the `--parameters` flag followed by a JSON string to specify which parameters to use when running the discovered test:
+* Isolation by default so system tests are as reliable as possible.
+* Utilities for pulling up and tearing down services easily in clusters in different environments (e.g. local,
+* custom cluster, Vagrant, K8s, Mesos, Docker, cloud providers, etc.)
+* Easy to write unit tests for distributed systems
```
ewencp: I'd probably phrase this a bit differently since this is kind of confusing (a unit test by definition can't really be for a distributed system). Perhaps something more like:

```
* Write tests for distributed systems in a simple unit test-like style
```

norwood:
this is awesome

docs/requirements.txt
```diff
@@ -0,0 +1,6 @@
+Sphinx
```
ewencp: You might want to `pip freeze` to get exact versions here. We've had problems in the past with compatibility for some of these libraries.

granders:
This is great @rancp, I'm very happy to see these docs!

PR #161, opened 2017-03-27
add pep8/style checks

rancp:
@ewencp @norwood


rancp:
Jenkins job is now enforcing styles as part of the build, we can disable this feature if needed

setup.py
```diff
@@ -30,19 +30,21 @@ def run_tests(self):
...
-        'console_scripts': ['ducktape=ducktape.command_line.main:main'],
+          'console_scripts': ['ducktape=ducktape.command_line.main:main'],
       },
       license="apache2.0",
       url="http://github.com/confluentinc/ducktape",
       packages=find_packages(),
       package_data={'ducktape': ['templates/report/*']},
       install_requires=['jinja2', 'requests', 'paramiko', 'pysistence', 'pyzmq'],
-      tests_require=['pytest', 'mock', 'psutil==4.1.0', 'memory_profiler==0.41', 'statistics', 'requests-testadapter', 'boto3', 'pycrypto', 'pywinrm'],
+      tests_require=['pytest', 'mock', 'psutil==4.1.0', 'memory_profiler==0.41',
```
ewencp: If you add `flake8` to `setup_requires`, per https://flake8.readthedocs.io/en/2.0/setuptools.html, you can get a `setup.py` command for doing flake8. This is also nice since in the Jenkins jobs you don't need to separately install flake8.

ewencp: Even better is if it can be done as part the `test` phase (such that anyone running tests would see errors, which means more of them get caught before the PR is filed), though I'm not sure what `setup.py` voodoo is required to make that happen.

rancp: Good point, that would be helpful, adding now

rancp:
retest this please

norwood:
🌶 🎱

PR #163, opened 2017-03-30
ST-281: added section about test debugging to ducktape docs

apovzner:
@rancp @ewencp

### DIFF
--- a/docs/debug_tests.rst
+++ b/docs/debug_tests.rst
@@ -1,0 +1,81 @@
+.. _topics-debug_tests:
+
+===========
+Debug Tests
+===========
+
+The test results go in ``results/<date>—<test_number>``. For results from a particular test, look for ``results/<date>—<test_number>/test_class_name/<test_method_name>/`` directory. The ``test_log.debug`` file will contain the log output from the test, and logs of services used in the test will be in ``service_name/node_name`` sub-directory.
+
+If there is not enough information in the logs, you can re-run the test with ``—no-teardown`` argument::
+
+    ducktape dir/tests/my_test.py::TestA.test_a —no-teardown
+
+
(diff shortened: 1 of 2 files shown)


docs/debug_tests.rst
```diff
@@ -0,0 +1,81 @@
...
+
+	node.account.kill_process("zookeeper", allow_fail=False)
+	wait_until(lambda: not self.alive(node), timeout_sec=5, err_msg="Timed out waiting for zookeeper to stop.")
+
+The test will progress as soon as condition is met, and timeout ensures that the test does not run indefinitely if termination never ends.
+
+Test Order Dependency
+^^^^^^^^^^^^^^^^^^^^^
+
+Make sure that your services properly cleanup the state in ``clean_node`` implementation. This can cause the next run of the test to fail or fail intermittently if other tests happen to clean same directories for example.
```
ewencp: I'd also recommend that they clean up aggressively. One of the benefits of the isolation that ducktape assumes is that you can assume you have complete control of the machine, so killing all java processes you can find rather than being more targeted is safe, deleting the entire temporary working space is ok, etc.

docs/debug_tests.rst
```diff
@@ -0,0 +1,81 @@
...
+If you are using Vagrant for example, you can then log into that node via::
+
+    vagrant ssh worker1
+
+
+
+Use Logging
+===========
+
+Distributed system tests can be difficult to debug. You want to add a lot of logging for debugging and tracking progress of the test. A good approach would be to log an intention of an operation with some useful information before any operation that can fail. Also, put enough information to a message of ``assert`` to help figure out what went wrong. Consider an example of testing ElasticSearch service::
```
ewencp: maybe also mention that it can be a good idea to use a higher logging level than you would in prod so more info is available, e.g. maybe make your log levels default to DEBUG instead of INFO. This is good thing to adjust in the Services themselves rather than at a test level.

docs/debug_tests.rst
```diff
@@ -0,0 +1,81 @@
...
+
+===========
+Debug Tests
+===========
+
+The test results go in ``results/<date>—<test_number>``. For results from a particular test, look for ``results/<date>—<test_number>/test_class_name/<test_method_name>/`` directory. The ``test_log.debug`` file will contain the log output from the test, and logs of services used in the test will be in ``service_name/node_name`` sub-directory.
+
+If there is not enough information in the logs, you can re-run the test with ``—no-teardown`` argument::
+
+    ducktape dir/tests/my_test.py::TestA.test_a —no-teardown
```
ewencp: If you use `.. sourcecode:: <language>` on these blocks, you can get syntax highlighting to make them more readable.

PR #166, opened 2017-06-06
Collect all logs when a test fails

ewencp:


ewencp:
Would like some more thoughts because I think this is a substantial change to defaults -- I don't think it's too dangerous, but many GB of data can be the delta of this change. On the other hand, this change makes debugging failures a *lot* simpler.

ewencp:
retest this please

PR #167, opened 2017-08-11
remoteaccount: add java_pids, kill_java_processes

cmccabe:
Add java_pids, a function to find the process IDs of running java
processes based on a classname string.

Add kill_java_processes, a function to kill running java processes based
on their classname.

These functions are more precise than grepping for java and then
grepping for a further string in the process name.  That process can
catch many java processes besides the intended one.

### DIFF
--- a/ducktape/cluster/remoteaccount.py
+++ b/ducktape/cluster/remoteaccount.py
@@ -371,6 +371,22 @@ class RemoteAccount(HttpMixin):
         for pid in pids:
             self.signal(pid, sig, allow_fail=allow_fail)
 
+    def java_pids(self, match):
+        cmd = """jcmd | grep """ + match + """ | awk '{print $1}'"""
+        return [int(pid) for pid in self.ssh_capture(cmd, allow_fail=True)]
+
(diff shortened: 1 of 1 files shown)


ducktape/cluster/remoteaccount.py
```diff
@@ -371,6 +371,22 @@ def kill_process(self, process_grep_str, clean_shutdown=True, allow_fail=False):
         for pid in pids:
             self.signal(pid, sig, allow_fail=allow_fail)
 
+    def java_pids(self, match):
+        cmd = """jcmd | grep """ + match + """ | awk '{print $1}'"""
```
xvrl: you could simplify to`"jcmd | awk '/%s/ {print $1}'" % match`

cmccabe: Ah, I figured there was something I could do like that with awk.  Will try it...

cmccabe:
[clabot:check]

ewencp:
I think this is fine, but I'd also suggest keeping in mind that we have non-java stuff to test as well and the framework is (ideally) generic, so adding too many java-isms to the core APIs would hopefully be avoided.

PR #168, opened 2017-08-16
Introduce ClusterSpec

cmccabe:
Introduce the ClusterSpec object to represent a specification for a cluster.
Currently, ClusterSpec can express that a cluster must have a certain number of
nodes of various operating systems.  In the future, we may add support for
tagging nodes.

Cluster: remove OS-specific methods and consolidate interfaces into three
methods which return ClusterSpec objects: available(), used(), and all().
Remove deprecated "request" method, since it is not being used.
The Cluster#alloc method now takes a cluster spec, not an int or a map.

JsonCluster: rename _available_nodes to _available_accounts, since this
collection stores account objects rather than node objects.

Service: add a num_nodes property for tests which expect it.

Test: add a new min_cluster_spec method.  Provide a better compatibility path
for Test subclasses which currently override min_cluster_size.

TestContext#close: do not throw an exception on double close-- this makes unit
test failures hard to read.

Rename the ClusterSlot class to ClusterNode.

Add a NodeContainer class which can efficiently store nodes in a per-OS
fashion.  The class also has utility methods for removing nodes matching a cluster
spec.  This simplifies writing Cluster subclasses

### DIFF
--- a/ducktape/cluster/cluster_spec.py
+++ b/ducktape/cluster/cluster_spec.py
@@ -1,0 +1,119 @@
+# Copyright 2017 Confluent Inc.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+# http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+from __future__ import absolute_import
+
+import json
+
+from ducktape.cluster.node_container import NodeContainer
+
+LINUX = "linux"
+
+WINDOWS = "windows"
+
+SUPPORTED_OS_TYPES = [LINUX, WINDOWS]
+
+
+class NodeSpec(object):
+    """
+    The specification for a ducktape cluster node.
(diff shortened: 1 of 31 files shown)


aayars:
Is there an associated documentation update for this PR?

cmccabe:
That's a good point, @aayars.  I will add some documentation to `docs/new_services.rst` and maybe a few other places where we talk about `Service#num_nodes`

cmccabe:
I added some new documentation about ClusterSpec.

ducktape/cluster/cluster_spec.py
```diff
@@ -0,0 +1,97 @@
...
+    @staticmethod
+    def empty():
+        return ClusterSpec([])
+
+    @staticmethod
+    def simple_linux(num_nodes):
+        """
+        Create a ClusterSpec containing some simple Linux nodes.
+        """
+        """Convenience method to create a cluster of all linux nodes."""
```
cmccabe: Oops, bad merge

ducktape/cluster/node_container.py
```diff
@@ -0,0 +1,160 @@
...
+
+class NodeNotPresentError(Exception):
+    pass
+
+
+class InsufficientResourcesError(Exception):
+    pass
+
+
+class NodeContainer(object):
```
cmccabe: Hmm.  I don't think it's a good fit for just a list or a set-- it stores nodes grouped by OS and possibly grouped in other ways in the future.  Also all the stuff discussed here: http://www.mit.edu/~6.005/fa14/classes/14-inheritance/ :)

PR #169, opened 2017-08-17
TestContext should be robust against multiple close calls

cmccabe:
TestContext should be robust against multiple close calls.  This helps
when debugging unit test issues that result in close being called more
than once.

### DIFF
--- a/tests/tests/check_test.py
+++ b/tests/tests/check_test.py
@@ -39,6 +39,14 @@ class DummyTestNoDescription(Test):
         pass
 
 
+class CheckLifecycle(object):
+    def check_test_context_double_close(self):
+        context = TestContext(session_context=ducktape_mock.session_context(),
+                              cls=DummyTest, function=DummyTest.test_function_description)
+        context.close()
+        context.close()
+
+
 class CheckEscapePathname(object):
 
     def check_illegal_path(self):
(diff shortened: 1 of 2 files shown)


cmccabe:
Good idea-- I added a test case

tests/tests/check_test.py
```diff
@@ -39,6 +39,14 @@ def test_this(self):
         pass
 
 
+class CheckLifecycle(object):
+    def check_test_context_double_close(self):
```
hachikuji: Is it worth asserting that "services" is undefined in `context`?

cmccabe:
> It looks like this conditional may have been guarding against python errors in the unlikely case that it's called twice or something.

In general, calling close twice is not an error.  This is a common convention in Java, and it appears that Python follows it as well.  For example, the Python file object can be closed more than once with no errors.  https://docs.python.org/2.7/library/stdtypes.html#file.close

> Is there a reason for removing?

The reason for removing is that I had some unit test failures which I was unable to debug until I fixed this issue.  Basically because of a bug in my code, the TestContext got closed, and then there was a `finally` block in the unit test that closed it again.  And the resulting exception obscured the original problem.

tests/tests/check_test.py
cmccabe: added

PR #174, opened 2017-08-25
Fix ClusterNode#name and add a unit test

cmccabe:


tests/cluster/check_json.py
```diff
@@ -123,3 +123,15 @@ def check_exhausts_supply(self):
...
+        cluster = JsonCluster(
+            {"nodes": [
+                {"ssh_config": {"host": "localhost1"}},
+                {"ssh_config": {"host": "localhost2"}},
+                {"ssh_config": {"host": "localhost3"}}]})
+        hosts = set(["localhost1", "localhost2", "localhost3"])
+        nodes = cluster.alloc(cluster.available())
+        for node in nodes:
+            hosts.remove(node.name)
+        assert 0 == len(hosts)
```
ewencp: you could just do `assert hosts == set(node.name for node in nodes)` and remove a few lines here (and I think the intent is a lot more obvious as well)

cmccabe: Sounds good.  Changed.

PR #177, opened 2017-09-21
Improve visibility of exceptions from background threads

ewencp:


ducktape/services/background_thread.py
```diff
@@ -33,11 +34,13 @@ def _protected_worker(self, idx, node):
...
-        except BaseException as e:
+        except BaseException:
             with self.lock:
                 self.logger.info("BackgroundThreadService threw exception: ")
-                self.logger.info(traceback.format_exc(limit=16))
-                self.worker_errors[threading.currentThread().name] = e
+                tb = traceback.format_exc()
+                self.logger.info(tb)
+                self.worker_errors[threading.currentThread().name] = tb
+                self.errors += "%s: %s" % (threading.currentThread().name, tb)
```
aayars: Perhaps add a line break, if multiple errors are possible.

PR #179, opened 2017-09-27
add defaults parametrize mark

rancp:
makes setting default parameters easier when we have lots of `parametrize` and `matrix` decorators
@confluentinc/c3

### DIFF
--- a/ducktape/mark/_mark.py
+++ b/ducktape/mark/_mark.py
@@ -134,6 +134,43 @@ class Matrix(Mark):
         return super(Matrix, self).__eq__(other) and self.injected_args == other.injected_args
 
 
+class Defaults(Mark):
+    """Parametrize with a default matrix of arguments on existing parametrizations.
+    Assume each values in self.injected_args is iterable
+    """
+
+    def __init__(self, **kwargs):
+        self.injected_args = kwargs
+        for k in self.injected_args:
+            try:
+                iter(self.injected_args[k])
+            except TypeError, te:
+                raise DucktapeError("Expected all values in @defaults decorator to be iterable: " + str(te))
+
+    @property
+    def name(self):
+        return "DEFAULTS"
+
+    def apply(self, seed_context, context_list):
+        new_context_list = []
+        if context_list:
+            for ctx in context_list:
+                for injected_args in cartesian_product_dict(
+                        {arg: self.injected_args[arg] for arg in self.injected_args if arg not in ctx.injected_args}):
+                    injected_args.update(ctx.injected_args)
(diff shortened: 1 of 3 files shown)


tests/mark/check_parametrize.py
```diff
@@ -167,3 +167,190 @@ def function(self, x=1, y=2, z=3):
...
+        assert len(context_list) == 1
+
+        expected_output = {(1, 2, 3, 4)}
+        output = set()
+        for ctx in context_list:
+            output.add(ctx.function())
+
+        assert output == expected_output
+
+    def check_overlap_param_no_override_method(self):
```
ewencp: i'd probably only have the method variant for the first test. i don't think this test really gains us much

tests/mark/check_parametrize.py
```diff
@@ -167,3 +167,190 @@ def function(self, x=1, y=2, z=3):
...
+            return x, y, z
+
+        assert parametrized(function)
+        injected_args_list = [m.injected_args for m in function.marks]
+        assert len(injected_args_list) == 2
+
+        context_list = MarkedFunctionExpander(None, function=function).expand()
+        assert len(context_list) == 4
+
+        expected_output = {(1, 5, 40), (1, 6, 40), (2, 5, 40), (2, 6, 40)}
```
ewencp: this seems a bit confusing. the default value and `@defaults` annotation just seem at odds with each other, is this really the result we want? i think if i was just reading the code in a test I would actually expect the `z` parameterizations from `@defaults` to be used, but i'm not sure either one is really right

ducktape/mark/_mark.py
```diff
@@ -134,6 +134,47 @@ def __eq__(self, other):
...
+    @property
+    def name(self):
+        return "DEFAULTS"
+
+    def apply(self, seed_context, context_list):
+        new_context_list = []
+        if context_list:
+            for ctx in context_list:
+                for injected_args in cartesian_product_dict(self.injected_args):
+                    if set(injected_args).intersection(set(ctx.injected_args)):
```
ewencp: Not sure I understand the logic here. If there is *any* overlap in the set of default args and the args, we basically do nothing to that parameterization? I guess I would have expected that defaults apply argument by argument.

rancp: now I think about it, it is probably more intuitive to have this apply arg by arg, I did that initially and changed it for some reason..

tests/mark/check_parametrize.py
rancp: this will change once I change the logic above

PR #180, opened 2017-11-01
SSH command failure will either raise exception or log error

rhauch:
Current, when an SSH command fails, an exception will be raised only if `allow_fail` is false; if it is true, the result of any failed SSH command will be silently ignored.

This PR changes the behavior to log a debug message with the failure information when `allow_fail` is true. All other behavior is unchanged.

### DIFF
--- a/ducktape/cluster/remoteaccount.py
+++ b/ducktape/cluster/remoteaccount.py
@@ -247,8 +247,12 @@ class RemoteAccount(HttpMixin):
         stdout.read()
         exit_status = stdout.channel.recv_exit_status()
         try:
-            if not allow_fail and exit_status != 0:
-                raise RemoteCommandError(self, cmd, exit_status, stderr.read())
+            if exit_status != 0:
+                if not allow_fail:
+                    raise RemoteCommandError(self, cmd, exit_status, stderr.read())
+                else:
+                    msg = "Running ssh command exited with status %d and message '%s': %s" % (exit_status, stderr.read(), cmd)
+                    self._log(logging.DEBUG, msg)
         finally:
             stdin.close()
             stdout.close()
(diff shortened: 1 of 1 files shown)


ducktape/cluster/remoteaccount.py
```diff
@@ -247,8 +247,12 @@ def ssh(self, cmd, allow_fail=False):
         stdout.read()
         exit_status = stdout.channel.recv_exit_status()
         try:
-            if not allow_fail and exit_status != 0:
-                raise RemoteCommandError(self, cmd, exit_status, stderr.read())
+            if exit_status != 0:
+                if not allow_fail:
+                    raise RemoteCommandError(self, cmd, exit_status, stderr.read())
+                else:
+                    msg = "Running ssh command exited with status %d and message '%s': %s" % (exit_status, stderr.read(), cmd)
```
edenhill: I would rearrange it to print the command line prior to the stderr output since the latter can be multiline.

rhauch: Sorry, can you explain?

rhauch: oh, I think I understand.

edenhill: If the command fails and prints multiple lines of output to stderr the debug log will be like:
```
Running ssh command exited with status 2 and message 'Failed to bla bla
because here is an exception:
  function1
  method2
  otherthing
and some/other/stuff/maybe': /path/to/a/program/that/fails --some args
```

and I personally find that harder to read than prefixing the output with the command:
```
Running ssh command '/path/to/a/program/that/fails --some args' exited with status 2: Failed to bla bla
because here is an exception:
  function1
  method2
  otherthing
and some/other/stuff/maybe'
```


But this is a tiny tiny nit pick of personal preference, it may be ignored :)

rhauch:
@ewencp I've corrected the parens (sorry, fixed locally and didn't push), but is the build still failing because of the warnings?

ewencp:
Hmm, seems like maybe flake8 got updated to a newer version which may have started causing new flake8 errors that kill the build. I think we are not specifying an exact version so it looks like the build just started breaking a few weeks ago. I filed https://github.com/confluentinc/ducktape/pull/181 which should hopefully fix that and allow for a green build here.

ewencp:
retest this please

PR #184, opened 2017-12-11
add env annotation

dguy:
Adding `@env` annotation so we can ignore tests when the environment isn't correct, i.e, incorrect version of java etc

### DIFF
--- a/ducktape/mark/_mark.py
+++ b/ducktape/mark/_mark.py
@@ -17,6 +17,7 @@ from ducktape.errors import DucktapeError
 
 import functools
 import itertools
+import os
 
 
 class Mark(object):
@@ -89,7 +90,6 @@ class Ignore(Mark):
 
     def apply(self, seed_context, context_list):
         assert len(context_list) > 0, "ignore annotation is not being applied to any test cases"
-
         for ctx in context_list:
             ctx.ignore = ctx.ignore or self.injected_args is None or self.injected_args == ctx.injected_args
         return context_list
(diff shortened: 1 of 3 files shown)


dguy:
@apurvam @ewencp I have a test that fails as `ctx.ignore` is always `False`. I’m not sure why, probably something obvious, but I have no idea. Any ideas?

ducktape/mark/_mark.py
```diff
@@ -190,10 +190,33 @@ def __eq__(self, other):
...
+            if os.environ[key] != value:
+                self.should_ignore = True
+
+    @property
+    def name(self):
+        return "ENV"
+
+    def apply(self, seed_context, context_list):
+        for ctx in context_list:
+            ctx.ingore = ctx.ignore or self.should_ignore
```
apurvam: `ctx.ignore` is probably always false because of the typo on the left hand side.. you have written `ctx.ingore` ..

dguy: yay! you don't know how long i've been looking at that! FFS!

dguy:
@apruvam yes that is the idea

dguy:
@ewencp you ok with this?

ducktape/mark/_mark.py
```diff
@@ -190,10 +190,33 @@ def __eq__(self, other):
         return super(Parametrize, self).__eq__(other) and self.injected_args == other.injected_args
 
 
+class Env(Mark):
+    def __init__(self, **kwargs):
+        self.injected_args = kwargs
+        self.should_ignore = False
+        for key, value in kwargs.iteritems():
```
ewencp: something like `self.should_ignore = any(os.environ[key] == value for key,value in kwargs.iteritems())` would probably be more idiomatic if you want to tighten this up a bit.

PR #186, opened 2017-12-18
Support kwargs on Service lifecycle methods and do not throw NotImplementedError by default

ewencp:


ewencp:
Motivation for this is the HttpMetrics mixin I recently wrote. I found that it was difficult to put it into place as a well-behaving mixin that passes along the call to `super` without adding in some hackiness; without something to intercept the call being passed along to the super class, the base class would throw NotImplementedError. in retrospect, we should have just `pass`ed anyway. start_node is the only one that should absolutely be required, and it'll be super obvious if you mess it up and don't actually start the service. the rest should be optional, even stop_node, because the Service could potentially be just a very quick command that then exits.

For the `**kwargs` arguments, they shouldn't be strictly required, but do allow mixins to be a little bit simpler by just inspecting the `**kwargs` rather than modifying them. this isn't critical, but I think we'll end up with code that's less likely to break this way. Because the framework doesn't assume any args will actually be passed implementations of Service don't *have* to maintain the **kwargs.

I believe this should all be backward compatible, but we should run a full set of tests against this version of ducktape to confirm.

ivandasch:
@ewencp Hi. This is great feature IMO. Is there any reason not to merge this PR?

stan-confluent (approved):

PR #189, opened 2018-02-28
Make SSHOutputIter reusable after timeout

rodesai:
This patch makes SSHOutputIter reusable after a call to has_next
times out. SSHOutputIter now accepts a generator rather than an
iterator. It uses the generator to create its iterator and recreates
the iterator when iterator.next() raises a timeout exception

### DIFF
--- a/ducktape/cluster/remoteaccount.py
+++ b/ducktape/cluster/remoteaccount.py
@@ -311,7 +311,7 @@ class RemoteAccount(HttpMixin):
                 stdout.close()
                 stderr.close()
 
-        return SSHOutputIter(output_generator(), stdout)
+        return SSHOutputIter(output_generator, stdout)
 
     def ssh_output(self, cmd, allow_fail=False, combine_stderr=True, timeout_sec=None):
         """Runs the command via SSH and captures the output, returning it as a string.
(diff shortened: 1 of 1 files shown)


ducktape/cluster/remoteaccount.py
```diff
@@ -605,12 +605,13 @@ class SSHOutputIter(object):
...
     """
 
-    def __init__(self, iter_obj, channel_file=None):
+    def __init__(self, iter_obj_f, channel_file=None):
         """
         :param iter_obj: An iterator
         :param channel_file: A paramiko ``ChannelFile`` object
         """
-        self.iter_obj = iter_obj
+        self.iter_obj_f = iter_obj_f
```
ewencp: docs right above this need updating. I'd also probably use `_func` instead to make it clear its a callable.

rodesai:
> is there a chance this loses data because the stdout.readline in output_generator reads some data, then loses connection and times out, then resumes but with half the line gone? still probably better than current behavior, and I'm guessing you're running into this because output flows for some time, then pauses for a long time waiting for another line, then resumes, so maybe the edge case doesn't matter much.

I don't think so. It's the same stdout object across calls to the generator. And its readline implementation would have to internally buffer to handle receiving multiple lines at once. I'll look at the implementation to make sure.

rodesai:
Looks like you were on the money. Paramiko's BufferedReader buffers in a function scoped variable until its consumed a line and then update its internal buffer before returning. Bummer. Maybe still worth keeping this in but calling this limit out in the doc comment.

PR #190, opened 2018-03-02
Use separate tables for test results

hachikuji:
It's annoying searching through test results looking for failures, so this patch puts them in a separate table. The CSS changes probably need tweaking.


hachikuji:
cc @ewencp

ducktape/tests/reporter.py
```diff
@@ -187,17 +187,31 @@ def test_results_dir(self, result):
         test_results_dir = os.path.abspath(result.results_dir)
         return test_results_dir[len(base_dir):]  # truncate the "absolute" portion
 
+    def append_result(self, result_string, result):
```
ewencp: Seems this ended up unused?

hachikuji: Ah, yeah, sorry about that. Had it initially, but felt like overkill in the end.

PR #191, opened 2018-06-06
Get context errors before tearing down test

iso88592:
On non-self test failures ducktape fails with `AttributeError: 'TestContext' object has no attribute 'services''


iso88592:
[clabot:check]

PR #195, opened 2018-08-28
ST-1103: Add Dockerfile to build ducktape image

maxzheng:
This will be published to Docker Hub as confluentinc/ducktape

```
$ docker run -it confluentinc/ducktape ducktape --version
0.7.3
```


Dockerfile
```diff
@@ -0,0 +1,23 @@
+# An image of ducktape that can be used to setup a Docker cluster where ducktape is run inside the container.
+
+FROM ubuntu:14.04
```
mohnishbasha: Cant we use a light weight OS like alpine? Ubuntu is heavy.

Dockerfile
```diff
@@ -0,0 +1,23 @@
...
+    virtualenv /opt/ducktape && \
+    . /opt/ducktape/bin/activate && \
+    pip install -U pip==9.0.3 setuptools wheel && \
+    pip install bcrypt cryptography==2.2.2 pynacl && \
+    mkdir /var/run/sshd && \
+    mkdir /root/.ssh && \
+    apt-get clean && \
+    rm -rf /var/lib/apt/lists/* /tmp/* /var/tmp/*
+
+ARG DUCKTAPE_VERSION=0.7.3
```
mohnishbasha: Add space?

Dockerfile
mohnishbasha: parameterize ubuntu version 14.04

Dockerfile
```diff
@@ -0,0 +1,23 @@
+# An image of ducktape that can be used to setup a Docker cluster where ducktape is run inside the container.
+
+FROM ubuntu:14.04
+
+RUN apt-get update && \
+    apt-get install -y libffi-dev libssl-dev openssh-server python-dev python-pip python-virtualenv && \
+    virtualenv /opt/ducktape && \
+    . /opt/ducktape/bin/activate && \
+    pip install -U pip==9.0.3 setuptools wheel && \
+    pip install bcrypt cryptography==2.2.2 pynacl && \
```
mohnishbasha: parameterize pip and cryptography version

Dockerfile
maxzheng: That is what we use in muckrake, so we have the use the same, otherwise this image wouldn't be useful at all. Downstream projects can use various OS and maybe we will publish different OSes later if the need comes up, but for now, just Ubuntu.

maxzheng: It doesn't change often, but maybe later when we publish multiple OSes.

Dockerfile
maxzheng: Remove pip version as we don't need it in virtualenv. As for cryptography, we don't actually need to pin the version, but due to 14.04, there is a warning to upgrade to Python 2.7.7. Maybe later when we support different OS versions.

PR #200, opened 2018-09-26
Enable automated testing

maxzheng:
Test job works: https://jenkins.confluent.io/job/mzheng-test-multibranch-pipeline/job/enable-automated@master/

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -1,0 +1,3 @@
+python {
+    publish = false  // Release is done manually to PyPI and not supported yet.
+}


Jenkinsfile
```diff
@@ -0,0 +1,3 @@
+python {
+    publish = false  // Release is done manually to PyPI and not supported yet.
```
ewencp: This is fine for now, but see comment on other PR related to this from connect. It'd be good to get this into a general classification of projects (e.g. in this case not automatically released python projects) rather than customizing on very narrow behavior like this. With a bunch of narrow settings like this, it becomes hard to reason about what you should configure and what variations of project configurations/standards we have.

PR #201, opened 2018-10-30
Upgrade to latest requests to fix potential vulnerability.

maxzheng:
Based on our usage, it does not really matter, but it is easy to upgrade so why not.

See https://github.com/confluentinc/ducktape/network/alert/setup.py/requests/open


maxzheng:
Ping @confluentinc/tools

PR #202, opened 2018-11-08
Use builtin dir to populate template properties

cyrusv:
The previously used __dict__ method will only show class properties
for current class, and not for superclasses
This change prepopulates the result dictionary with the builtin dir
which includes superclasses, and then applies the existing logic
in order to maintain behavior.

### DIFF
--- a/ducktape/template.py
+++ b/ducktape/template.py
@@ -21,6 +21,11 @@ import inspect
 
 class TemplateRenderer(object):
 
+    def _get_ctx(self):
+        ctx = {k: getattr(self.__class__, k) for k in dir(self.__class__)}
+        ctx.update(self.__dict__)
+        return ctx
+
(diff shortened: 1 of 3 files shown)


ducktape/template.py
```diff
@@ -31,7 +31,8 @@ def render_template(self, template, **kwargs):
         """
         if not hasattr(template, 'render'):
             template = Template(template)
-        ctx = dict(self.__class__.__dict__)
+        ctx = {k: getattr(self, k) for k in dir(self.__class__)}
+        ctx.update(dict(self.__class__.__dict__))
```
ewencp: Is anything extra being captured here vs what is covered by `dir` collecting props up the class hierarchy? If yes, this case doesn't seem to handle multiple levels of inheritance; if not, we can just remove this.

PR #203, opened 2018-11-20
MINOR: Improve error message in service wait

kkonstantine:


kkonstantine:
Service throws exception with a message that is too hard to interpret. Eg.
```
These nodes are still alive: [<ducktape.cluster.cluster.ClusterNode object at 0x7f05a83834d0>]
```

Re-using the formatted string that is used elsewhere should do the trick.

kkonstantine:
@ewencp `ClusterNode` doesn't seem to have info such as `service_id` (which is the one I'd characterize as more useful here) or `idx`. Just the `node.account.hostname`

Unless I'm missing something, I wouldn't be keen on shuffling too much info around as part of this simple PR. 

@cyrusv @ewencp thank you for the reviews. I'll merge if there's no other suggestion at this point.

PR #204, opened 2018-11-20
ST-1307: Switch from pycrypto to pycryptodome and update to a modern version

ewencp:


ewencp:
I also ran the tests under `systest/` locally to validate it still works properly end-to-end.

ewencp:
ping @confluentinc/tools

PR #206, opened 2018-12-10
added extra ducktape install commmand line for error message 1. (Oper…

woshifudayun:
…ation not permitted.)

### DIFF
--- a/docs/install.rst
+++ b/docs/install.rst
@@ -33,3 +33,7 @@ Install
 
         C_INCLUDE_PATH=/usr/local/opt/openssl/include LIBRARY_PATH=/usr/local/opt/openssl/lib pip install ducktape
 
+    If you got error message `failed with error code 1`, you may need to install ducktape by::
+
+        pip install --user ducktape
+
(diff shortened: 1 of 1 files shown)


docs/install.rst
```diff
@@ -33,3 +33,7 @@ Install
 
         C_INCLUDE_PATH=/usr/local/opt/openssl/include LIBRARY_PATH=/usr/local/opt/openssl/lib pip install ducktape
 
+    If you got error message `failed with error code 1`, you may need to install ducktape by::
+
+        pip install --user ducktape
```
ewencp: This seems weird, why would there be a `ducktape` user? And are you installing in a virtualenv or to the system python install? If the latter, you should probably be in a virtualenv.

maxzheng: This is installing `ducktape` into the user's personal site, which is an alternative to using virtualenv (still bad imo). Perhaps clarify that this is for the non-virtualenv case?

ewencp: Ah, just read through this too quickly. I guess alternative is a gross `sudo pip install ducktape`. Agree with clarifying non-virtualenv since these days you should pretty much not install anything to system python.

woshifudayun:
Thanks Ewen. Sorry I lost track of this PR previously. Yes, only by using `--user` I can install ducktape in my virtualenv previously. That's why I added this.

PR #207, opened 2019-01-23
Fix bug in @cluster decorator

whynick1:
There is a bug in current `@cluster` implementation, which only support `num_nodes`, but not `cluster_spec`.

Suppose we have, `@cluster(cluster_spec=ClusterSpec.simple_linux(3))` for the test, while only 2 linux nodes is up. We expect the test to fail as below
```
xxx_test requires more resources than are available in the whole cluster. linux nodes requested: 3. linux nodes available: 2
```
However, the test will continue, ignoring the "hint". 

Here is the fix.

### DIFF
--- a/ducktape/tests/test.py
+++ b/ducktape/tests/test.py
@@ -27,8 +27,7 @@ from ducktape.utils.local_filesystem_utils import mkdir_p
 from ducktape.command_line.defaults import ConsoleDefaults
 from ducktape.services.service_registry import ServiceRegistry
 from ducktape.template import TemplateRenderer
-from ducktape.mark.resource import CLUSTER_SPEC_KEYWORD
-from ducktape.mark.resource import CLUSTER_SIZE_KEYWORD
+from ducktape.mark.resource import CLUSTER_SPEC_KEYWORD, CLUSTER_SIZE_KEYWORD
 from ducktape.tests.status import FAIL
 
 
@@ -285,7 +284,7 @@ class TestContext(object):
         :param file: file containing this module
         :param injected_args: a dict containing keyword args which will be passed to the test method
         :param cluster_use_metadata: dict containing information about how this test will use cluster resources,
-               to date, this only includes "num_nodes"
+               to date
         """
 
         self.session_context = kwargs.get("session_context")
@@ -302,7 +301,7 @@ class TestContext(object):
         self.ignore = kwargs.get("ignore", False)
 
         # cluster_use_metadata is a dict containing information about how this test will use cluster resources
-        # to date, this only includes "num_nodes"
+        # to date
         self.cluster_use_metadata = copy.copy(kwargs.get("cluster_use_metadata", {}))
 
         self.services = ServiceRegistry()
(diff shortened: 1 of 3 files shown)


ducktape/tests/test.py
```diff
@@ -384,8 +385,11 @@ def expected_cluster_spec(self):
 
         :return:            A ClusterSpec object.
         """
+        cluster_spec = self.cluster_use_metadata.get(CLUSTER_SPEC_KEYWORD)
         cluster_size = self.cluster_use_metadata.get(CLUSTER_SIZE_KEYWORD)
-        if cluster_size is not None:
+        if cluster_spec is not None:
```
whynick1: if cluster_spec provided, return it.
else if nums_nodes provided, use that to create ClusterSpec accordingly.
else if test context not successfully initialized, expect 0 node because test is not ready.
else test is ready, but no valid hint for resource available, set expected_nums_nodes to number of nodes we have.

whynick1: @ewencp

ducktape/tests/test.py
```diff
@@ -27,6 +27,7 @@
 from ducktape.command_line.defaults import ConsoleDefaults
 from ducktape.services.service_registry import ServiceRegistry
 from ducktape.template import TemplateRenderer
+from ducktape.mark.resource import CLUSTER_SPEC_KEYWORD
```
ewencp: nit: you can do `from ducktape.mark.resource import CLUSTER_SIZE_KEYWORD, CLUSTER_SPEC_KEYWORD` here. generally it's nicer to group imports from the same module where possible

ducktape/tests/test.py
whynick1: Done!

ducktape/tests/test.py
whynick1: Done!

whynick1:
* Added unit test
* Fixed nit-pick
@ewencp @maxzheng

criccomini:
@ewencp @maxzheng ping

criccomini:
@maxzheng ping

tests/mark/check_cluster_use_metadata.py
```diff
@@ -38,6 +39,19 @@ def function():
         assert len(test_context_list) == 1
         assert test_context_list[0].cluster_use_metadata == cluster_use_metadata
 
+    def check_basic_usage_cluster_spec(self):
+        num_nodes = 200
```
ewencp: just a style nit for the future: we should be consistent and always use a defined test constant or just leave them literals (which is generally ok in tests where they won't end up duplicated or really be significant constants anyway)

tests/mark/check_cluster_use_metadata.py
```diff
@@ -38,6 +39,19 @@ def function():
...
+
+        @cluster(cluster_spec=ClusterSpec.simple_linux(200))
```
maxzheng: use num_nodes here

PR #208, opened 2019-01-23
Skip compression on non-existent log file

kkonstantine:


kkonstantine:
I had to also fix a unit test on a relative path file that didn't seem to do the right thing. LMK if this looks ok.

kkonstantine:
Thanks for the approval @maxzheng !
In the meantime I added a system test for better coverage of the added functionality and removed the `echo` message of the `else` part, since it has no effect, and is not printed in the logs. 
@ewencp any comments?

kkonstantine:
Ok. All set. It's green. Merging and bumping up the version in a subsequent PR.

PR #210, opened 2019-02-01
adding SSHClient session expire check

brianbushree:
when running long-lived system tests I've run into `SSHException('SSH session not active')`

this change checks for a an active session both via the API and via actually sending a packet before returning the `SSHClient`

### DIFF
--- a/ducktape/cluster/remoteaccount.py
+++ b/ducktape/cluster/remoteaccount.py
@@ -159,22 +159,33 @@ class RemoteAccount(HttpMixin):
         msg = "%s: %s" % (str(self), msg)
         self.logger.log(level, msg, *args, **kwargs)
 
+    def _new_ssh_client(self):
+        client = SSHClient()
+        client.set_missing_host_key_policy(IgnoreMissingHostKeyPolicy())
+
+        self._log(logging.DEBUG, "ssh_config: %s" % str(self.ssh_config))
+
+        client.connect(
+            hostname=self.ssh_config.hostname,
+            port=self.ssh_config.port,
+            username=self.ssh_config.user,
+            password=self.ssh_config.password,
+            key_filename=self.ssh_config.identityfile,
+            look_for_keys=False)
+        self._ssh_client = client
+
     @property
     def ssh_client(self):
-        if not self._ssh_client:
-            client = SSHClient()
-            client.set_missing_host_key_policy(IgnoreMissingHostKeyPolicy())
-
-            self._log(logging.DEBUG, "ssh_config: %s" % str(self.ssh_config))
-
-            client.connect(
-                hostname=self.ssh_config.hostname,
-                port=self.ssh_config.port,
-                username=self.ssh_config.user,
-                password=self.ssh_config.password,
-                key_filename=self.ssh_config.identityfile,
-                look_for_keys=False)
-            self._ssh_client = client
+        if not self._ssh_client or \
+           not self._ssh_client.get_transport() or \
+           not self._ssh_client.get_transport().is_active():
+            self._new_ssh_client()
+        else:
+            try:
+                transport = self._ssh_client.get_transport()
+                transport.send_ignore()
+            except:
+                self._new_ssh_client()
 
         return self._ssh_client
 
--- a/ducktape/cluster/remoteaccount.py
+++ b/ducktape/cluster/remoteaccount.py
@@ -191,10 +191,20 @@ class RemoteAccount(HttpMixin):
 
         return self._ssh_client
 
+    def _new_sftp_client(self):
+        self._sftp_client = self.ssh_client.open_sftp()
+
     @property
     def sftp_client(self):
         if not self._sftp_client:
-            self._sftp_client = self.ssh_client.open_sftp()
+            self._new_sftp_client()
+        else:
+            try:
+                self._sftp_client.listdir(".")
+            except Exception as e:
+                self._log(logging.DEBUG, "exception getting sftp_client (creating new client): %s" % str(e))
+                self._sftp_client.close()
+                self._new_sftp_client()
 
         return self._sftp_client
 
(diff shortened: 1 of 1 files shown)


brianbushree:
@ewencp what's the best way to test this change?

ewencp:
@brianbushree thanks for the patch. i'm a bit surprised we hit this, we must have connections that idle for a very long time. I think this fix is reasonable, another improvement might be to try setting the KeepAlive for the connection, which might allow us to just keep the session active for much longer.

In terms of testing, I think baseline would be to mock out the `SSHClient` class and just get unit test coverage of the `ssh_client` property (and the new method, indirectly) to validate expected behavior. (See https://docs.python.org/3/library/unittest.mock.html for details if you're not familiar with this in python yet.) In terms of end-to-end validation, that's obviously harder with things like this -- if you can adjust timeouts to reproduce the issue consistently, you could potentially write a system test (under `systests/`), but I'm not sure how valuable that is unless we can keep the test pretty short. Not sure how easily you can reproduce this in your actual tests, but just validating you can't reproduce might be good enough on top of the unit tests.

ewencp:
Also, just a heads up that you can tag `tools` for reviews on this repo to get more eyes on PRs.

ducktape/cluster/remoteaccount.py
```diff
@@ -159,22 +159,35 @@ def _log(self, level, msg, *args, **kwargs):
...
+           not self._ssh_client.get_transport() or \
+           not self._ssh_client.get_transport().is_active():
+            self._new_ssh_client()
+        else:
+            try:
+                transport = self._ssh_client.get_transport()
+                transport.send_ignore()
+            except Exception as e:
+                self._log(logging.DEBUG, "exception getting ssh_client (creating new client): %s" % str(e))
+                self._ssh_client.close()
```
maxzheng: Perhap move this line into _set_ssh_client() so it automatically close instead of relying on caller.

ducktape/cluster/remoteaccount.py
```diff
@@ -159,22 +159,35 @@ def _log(self, level, msg, *args, **kwargs):
...
-                port=self.ssh_config.port,
-                username=self.ssh_config.user,
-                password=self.ssh_config.password,
-                key_filename=self.ssh_config.identityfile,
-                look_for_keys=False)
-            self._ssh_client = client
+        if not self._ssh_client or \
+           not self._ssh_client.get_transport() or \
+           not self._ssh_client.get_transport().is_active():
+            self._new_ssh_client()
```
maxzheng: instead of not / negative case, it's more readable/less cognitive load (Pythonic) to use positive first. And also parenthesis is preferred style. So:

```
if (self._ssh_client
    and self._ssh_client.get_transport()
    and self._ssh_client.get_transport().is_active()):
```

I think without `not`s, it is short enough to fit into a single line.

ducktape/cluster/remoteaccount.py
```diff
@@ -159,22 +159,35 @@ def _log(self, level, msg, *args, **kwargs):
         msg = "%s: %s" % (str(self), msg)
         self.logger.log(level, msg, *args, **kwargs)
 
+    def _new_ssh_client(self):
```
maxzheng: maybe _set_ssh_client is better?

ducktape/cluster/remoteaccount.py
brianbushree: that would require another conditional check in `_set_ssh_client()` to ensure it's not `None` before closing

ducktape/cluster/remoteaccount.py
```diff
@@ -159,29 +159,55 @@ def _log(self, level, msg, *args, **kwargs):
...
+            self._sftp_client.close()
+        self._sftp_client = self.ssh_client.open_sftp()
+
     @property
     def sftp_client(self):
-        if not self._sftp_client:
-            self._sftp_client = self.ssh_client.open_sftp()
+        if self._sftp_client:
+            try:
+                self._sftp_client.listdir(".")
```
ewencp: i don't think we want to do this much work to validate this. would the channel used for ftp time out independently (and earlier) than the underlying ssh transport? since this client is layered on top of the `ssh_client`, i think we could do the lighter weight check on just the ssh client. i.e. instead of trying to do something with sftp client, just do the `send_ignore` check and only replace `self._sftp_client` if it that check fails.

if we *do* need to do something with the sftp client itself, maybe do a `stat` instead of a `listdir`. the `stat` should be small and fixed size. `listdir` when cwd is in the wrong place could be quite expensive.

brianbushree: I think you're right. we should just need to check on the `self.ssh_client` and update the `self._sftp_client` when that fails

if not, will use `stat` instead

ducktape/cluster/remoteaccount.py
```diff
@@ -159,29 +159,50 @@ def _log(self, level, msg, *args, **kwargs):
...
+    def _set_sftp_client(self):
+        if self._sftp_client:
+            self._sftp_client.close()
+        self._sftp_client = self.ssh_client.open_sftp()
+
     @property
     def sftp_client(self):
         if not self._sftp_client:
-            self._sftp_client = self.ssh_client.open_sftp()
+            self._set_sftp_client()
```
ewencp: does this actually cover the case you were worried about re: ftp client? this catches if we haven't created the ftp client, but if you already have set it and it times out i think this new logic doesn't work. I think you'd need to make sure you *always* call `self.ssh_client` for every call to this method to ensure you handle any transport timeouts.

brianbushree: you're right. does not handle that case...

I think this should do it.
```
    @property
    def sftp_client(self):
        if not self._sftp_client:
            self._set_sftp_client()
        else:
            self.ssh_client  # test connection
        return self._sftp_client
```

brianbushree:
no problem, wasn't too bad actually! this codebase is pretty clean

PR #214, opened 2019-08-28
Added tox to dependencies

stan-confluent:
**Why**
If you run with a clean virtualenv and do `python setup.py develop`, then (as per readme) try to execute all tests with coverage via `tox`, you'll get an error `Unknown command tox`.

**What**
This PR adds `tox` to `setup.py`. Just picked the latest version, since I couldn't find out if they're using semver or not. I'm open to specifying it in any other way of course, suggestions are welcome!

It also adds `more-itertools`, which is a transitive dependency for `tox`, locked at a specific version compatible with python 2.7. `more-itertools` is specified as 'Any':
```
- tox [required: ==3.13.2, installed: 3.13.2]
    - pluggy [required: >=0.12.0,<1, installed: 0.12.0]
      - importlib-metadata [required: >=0.12, installed: 0.19]
        - zipp [required: >=0.5, installed: 0.6.0]
          - more-itertools [required: Any, installed: 5.0.0]
```
That's why pypi was trying to install a version that's not compatible with python 2.7, so I picked the latest one that still supports it. Seems to work fine :)

**Tested**
Clean virtualenv, `python setup.py develop`, then `tox`, ensured the run was successful.


sonupillai:
LGTM!

PR #215, opened 2019-08-28
Add python3 support

ewencp:
This adds python3 to the standard unit test matrix and is verified to
work with the ducktape system tests on both py27 and py37. Most issues
were found easily with existing unit tests. System tests revealed a
couple of additional problems with str/bytes differences in py3 and
in Vagrant cluster code that is more difficult to unit test.


ducktape/cluster/remoteaccount.py
```diff
@@ -588,11 +588,12 @@ def create_file(self, path, contents):
         with self.sftp_client.open(path, "w") as f:
             f.write(contents)
 
-    def mkdir(self, path, mode=0755):
+    _DEFAULT_PERMISSIONS = 493  # 0755
```
dnozay: `int('755',8)` would work too

ducktape/services/service_registry.py
```diff
@@ -52,7 +52,7 @@ def stop_all(self):
             except BaseException as e:
                 if isinstance(e, KeyboardInterrupt):
                     keyboard_interrupt = e
-                service.logger.warn("Error stopping service %s: %s" % (service, e.message))
+                service.logger.warn("Error stopping service %s: %s" % (service, e))
```
dnozay: nit: you could use ',' for variadic calls rather than interpolating directly.

ducktape/tests/loader.py
```diff
@@ -106,7 +107,7 @@ def load(self, test_discovery_symbols):
             # packing tests into bins based on the least full bin at the time.
             raw_results = _requests_session.get(self.historical_report).json()["results"]
             time_results = {r['test_id']: r['run_time_seconds'] for r in raw_results}
-            avg_result_time = sum(time_results.itervalues()) / len(time_results)
+            avg_result_time = sum(itervalues(time_results)) / len(time_results)
```
dnozay: possible divide by zero?

ducktape/utils/http_utils.py
```diff
@@ -12,7 +12,12 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-import urllib2
+try:
+    # For Python 3.0 and later
+    from urllib.request import build_opener, Request
```
dnozay: q: not available in six?

ducktape/tests/loader.py
ewencp: This shouldn't be possible since this block only makes sense with a result file, which would have to have contents (or would likely be malformed for this use anyway).

ducktape/services/service_registry.py
ewencp: Was just doing minimal conversion (I think there are a bunch of other cases of this as well), but have updated this one. There's probably a bunch of other cleanup we could do as well, especially as some style things changed from py2 -> 3.

PR #216, opened 2019-09-29
Add support to accept "skip_node_allocation" flag as kwargs

sonupillai:
This adds the ability to skip node allocation for a service. With this, another layer of abstraction
can be written over ductape services when the nodes and services are already known and are running and we just need to run some tests on the running services.

Ran all the unit tests using tox and got a clean run.


ducktape/command_line/parse_args.py
```diff
@@ -70,6 +70,8 @@ def create_ducktape_parser():
                         help="URL of a JSON report file containing stats from a previous test run. If specified, "
                              "this will be used when creating subsets of tests to divide evenly by total run time "
                              "instead of by number of tests.")
+    parser.add_argument("--allocate_nodes", action="store_true", help="Use this flag to skip allocating nodes for "
```
scott-hendricks: convention is to have this be `--allocate-nodes`

ducktape/services/service.py
```diff
@@ -76,7 +76,9 @@ def __init__(self, context, num_nodes=None, cluster_spec=None, *args, **kwargs):
         self.context = context
 
         self.nodes = []
-        self.allocate_nodes()
+        self.skip_nodes_allocation = kwargs.get("skip_nodes_allocation", False)
```
scott-hendricks: how does this kwarg get passed, if there's no mechanism to bring the command line argument into this function?

sonupillai: Type in parser. Fixed. Thank you!

ducktape/command_line/parse_args.py
```diff
@@ -70,6 +70,8 @@ def create_ducktape_parser():
                         help="URL of a JSON report file containing stats from a previous test run. If specified, "
                              "this will be used when creating subsets of tests to divide evenly by total run time "
                              "instead of by number of tests.")
+    parser.add_argument("--skip_nodes_allocation", action="store_true", help="Use this flag to skip allocating nodes for "
```
scott-hendricks: convention is to use `--skip-nodes-allocation`

PR #218, opened 2020-03-12
Debuggability improvements to remote commands

jspong:


ducktape/cluster/remoteaccount.py
```diff
@@ -146,24 +146,19 @@ def operating_system(self):
...
-            return logging.getLogger(__name__)
+        if not self._logger:
+            self._logger = logging.getLogger(__name__)
+        return self._logger
 
     @logger.setter
     def logger(self, logger):
         self._logger = logger
 
-    def _log(self, level, msg, *args, **kwargs):
```
ewencp: doesn't removing this remove information from the log since we're no longer logging the account info that this is executing from? seems like that would reduce debuggability?

That might not be the ideal info to log (e.g. you probably want to know the current logical use of the node, e.g. connect worker 2, rather than the specific account, e.g. user@node1, but it's better than nothing?

ducktape/cluster/remoteaccount.py
```diff
@@ -260,21 +255,24 @@ def ssh(self, cmd, allow_fail=False):
...
 
         client = self.ssh_client
         stdin, stdout, stderr = client.exec_command(cmd)
 
-        # Unfortunately we need to read over the channel to ensure that recv_exit_status won't hang. See:
-        # http://docs.paramiko.org/en/2.0/api/channel.html#paramiko.channel.Channel.recv_exit_status
-        stdout.read()
+        stdout_contents = stdout.read().strip()
+        stderr_contents = stderr.read().strip()
+        if stdout_contents:
```
ewencp: Do we want this here? It's going to add a lot of noise since this will log the output of every command, regardless of whether successful. Seems like putting it in the final non-failure allowed block (line 275) covers the 99% use case?

ducktape/cluster/remoteaccount.py
jspong: stacklevel seems to have been added in python 3.8, so we can't use that. we can use a logging adapter to add the output w/o modifying the stack though, so I did that.

ducktape/cluster/remoteaccount.py
jspong: FWIW, i've really been enjoying having all this additional output. we should also consider going through services and get rid of the many areas people explicitly log this functionality

ducktape/cluster/remoteaccount.py
```diff
@@ -313,28 +298,32 @@ def ssh_capture(self, cmd, allow_fail=False, callback=None, combine_stderr=True,
         stdout = chan.makefile('r', -1)
         stderr = chan.makefile_stderr('r', -1)
 
+        exit_status = [None]
+
         def output_generator():
 
             for line in iter(stdout.readline, ''):
+                self.logger.debug(line)
```
ewencp: This is likely to be excessive to log for some use cases. e.g., because the test driver needs to analyze the output, verifiable_producer/consumer stream directly to the driver node: https://github.com/apache/kafka/blob/trunk/tests/kafkatest/services/verifiable_consumer.py#L238. Depending on how fast events are being generated, that could turn into a massive log (and for some tests that try to stress the system and expose potential concurrency issues, we could be generating events quite quickly).

jspong: This makes sense. I added an overridable limit of 5MB. Testing now

andrewegel:
I'm closing this b/c John isn't around any longer, this can always be re-visted later.

PR #220, opened 2020-03-25
Add ability to run a sample of the full test suite

jspong:
This PR adds the --sample which, if provided, will give the size of a random sample of tests to run rather than all the tests that match the test path.

### DIFF
--- a/ducktape/command_line/main.py
+++ b/ducktape/command_line/main.py
@@ -17,6 +17,7 @@ from __future__ import print_function
 import importlib
 import json
 import os
+import random
 from six import iteritems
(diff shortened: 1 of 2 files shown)


jspong:
This behavior can be seen here: https://jenkins.confluent.io/job/jspong-system-test/411/console

ewencp:
I think this failed on style check, the CI failed before the merge too: https://jenkins.confluent.io/job/ducktape/1198/console

jspong:
oops! https://github.com/confluentinc/ducktape/pull/221 will fix this

PR #225, opened 2020-04-14
MINOR: Constrain MarkupSafe transitive dependency to avoid newer, incompatible versions

ewencp:


ewencp:
For reference, I created a 0.7.x off of v0.7.6. We haven't done this in the past for ducktape because changes have been relatively infrequent and so we generally have time to test updates against consumers of the tool. However, in this case, I'd like to publish 0.7.7 asap, so we'll merge it to 0.7.x, I'll do a version bump and release, then merge that into master. From there, we can decide whether to fast-forward 0.7.x to master.

This is probably something we should adopt going forward anyway so it matches the patterns we use elsewhere and we can isolate compatible from breaking changes.

PR #231, opened 2020-06-09
Support excluding tests

stan-confluent:
As a first step towards supporting test suites, this PR adds support for excluding certain tests.

I opted to explicitly filter by test_id in the loader instead of implementing `__eq__` and `__hash__` on TestContext itself - mainly because TestContext not only describes which test to load, but also deals with resources, cluster etc. So I wasn't sure - I could just use test_id or test_metadata in `__eq__` and `__hash__` but that might break things in other places I'm not aware of - or I could implement `__eq__` and `__hash__` using most/all fields, but then again not sure if it would break anything down the road or if some fields would be different for similar tests. 
@ewencp do you have any suggestions on implementing `__eq__` and `__hash__` for TestContext? If we have these methods, we could then just put included and excluded test contexts in sets and do `all = included - excluded` instead of map/filter as I do now.


ducktape/tests/loader.py
```diff
@@ -70,35 +70,68 @@ def __init__(self, session_context, logger, repeat=1, injected_args=None, cluste
...
+        included_contexts_list = self.load_test_contexts(included_test_symbols)
 
-            test_context_list_for_symbol = self.discover(directory, module_name, cls_name, method_name)
-            all_test_context_list.extend(test_context_list_for_symbol)
-
-            if len(test_context_list_for_symbol) == 0:
-                raise LoaderException("Didn't find any tests for symbol %s." % symbol)
-
-        self.logger.debug("Discovered these tests: " + str(all_test_context_list))
+        if not included_contexts_list:
```
ewencp, line 125: This seems redundant given the checks already happening in `load_test_contexts`.

stan-confluent: Thanks! Addressed this in the follow up PR.

ewencp:
Anything to prevent merging this now then?

stan-confluent:
@ewencp - yep, I can merge it.
@andrewegel  - I noticed one of your PRs went into 0.7.x and bumped the version from 0.7.7 to 0.7.8 - but its not in master branch. This PR also uses 0.7.8 version but does not target 0.7.x and targets master instead. Any objections to both of us using 0.7.8? Also, can you please also push your 0.7.x changes to master?

stan-confluent:
I'll close this since there's a test-suite PR for review already.

PR #233, opened 2020-06-17
QE-47: Test suite support

stan-confluent:
Adds support for test suites to ducktape.
Test suite is a yaml file with either a list of included tests or two lists - included and excluded.
Test suites are specified via the same arguments as normal test files and can be loaded side-by-side. 
Test suites are distinguished by `.yml` extension.
If folder is specified, all tests in that folder will be loaded, but not the test suites - this is to avoid dealing with circular imports and to not change existing code too much.
If the same test is specified multiple times (from either test suites or files or both), it will only be loaded once.
`--exclude` parameter specifies global excludes - these are files or symbols only (not test suites), and will be excluded from the final list of tests to run.

This PR does not change the reporting part of ducktape - your final report would still contain a flat list of tests.

Tested with vagrant, ducker-ak and ducker in muckrake. 

One note, I'm not sure how to properly test the "magic" code that updates `sys.path` - I moved it into loader, and things seem to work fine, but if any of the reviewers has a good suggestion on how to test this better, I'd be happy to.


tests/loader/resources/loader_test_directory/sub_dir_a/test_d.py
```diff
@@ -0,0 +1,37 @@
+# Copyright 2015 Confluent Inc.
```
andrewegel, line 1:
```suggestion
# Copyright 2020 Confluent Inc.
```

andrewegel:
Are we breaking any APIs here?

No => Then I don't see a reason to put this in master instead of `0.7.x`
Yes => Good idea to put it into a new major version. I assume then, we'll have to followup to update kafkatest in Kafka projects and Update Muckrake.

My other comment is around documentation: Seems like a good feature to write into the docs so we better understand how to use it.

stan-confluent:
@andrewegel Thanks!
We're not breaking any APIs, but adding functionality in a backwards compatible manner, which, according to https://semver.org/ is supposed to be a minor bump.
Agree on the documentation, working on it in a separate branch and will either update this PR or send a follow up.

As for master vs 0.7.x - shouldn't all the changes go to master first (so that master is a source of truth) and then cherrypicked as needed to the release branches?

tests/loader/resources/loader_test_directory/sub_dir_a/test_d.py
stan-confluent: do we care about copyright on the unit-test resources? Maybe just remove it entirely?

stan-confluent:
Updated docs, PTAL

andrewegel:
> As for master vs 0.7.x - shouldn't all the changes go to master first (so that master is a source of truth) and then cherrypicked as needed to the release branches?

Yeah, trunk-based development works. Internally though that pattern is inverted and automatically done with tooling to merge from older (0.7.x) -> newer (master).

I'm fine with going to master with the intention of back-porting to 0.7.x

tests/loader/resources/loader_test_directory/sub_dir_a/test_d.py
andrewegel: Its your call.

andrewegel:
One thought is to use a muckrake PR to point to this ducktape branch to test your changes internally.

stan-confluent:
Thanks folks!
@andrewegel :
>I'm fine with going to master with the intention of back-porting to 0.7.x
either way works for me really, though I was thinking of making this a 0.8.0 version, on account of this being a relatively large feature, which we'll potentially grow even further. I can merge this to master and then branch out to 0.8.x and continue development from there.
>One thought is to use a muckrake PR to point to this ducktape branch to test your changes internally.
I did that with local ducker. Not 100% sure how to test with jenkins builds - is it just running system tests jenkins job from a muckrake PR that has newer version?

PR #235, opened 2020-07-21
Add Retries to remote calls

andrewegel:
- Add retrying module as dependency
- Add SSHException exponential backoff retries to methods that perform
remote commands
- Change logging level to WARNING in _set_ssh_client

Some Background:
We've been seeing an increase in SSH exceptions from these code paths:
* `SSHException: Unable to open channel`
*  `[Errno 110] Connection timed out`
* `SSHException: Server connection dropped:`

While I don't think we have a true understanding of *why* there's been an increase in these lately, there isn't a reason *not* to add retries to distributed system code paths such as these.


ducktape/cluster/remoteaccount.py
```diff
@@ -250,6 +258,7 @@ def _can_ping_url(self, url, headers):
         except Exception:
             return False
 
+    @retry(retry_on_exception=retry_if_ssh_exception, wait_exponential_multiplier=1000, wait_exponential_max=60000)
```
stan-confluent, line 261: I'm a bit concerned about adding un-adjustable retries to the framework itself permanently - I mean we already know that ssh is flaky, but don't really know why - so while retries will possibly make the tests pass, they won't address the root of the problem. Especially them not being is what doesn't exactly *feel* right.

Also, this doesn't address ssh_capture and ssh_output methods. 

Also I wonder if we could adjust a connection timeout on `self.ssh_client.exec_command` and `self.ssh_client.open_session` and the like instead of retries.

andrewegel: > I'm a bit concerned about adding un-adjustable retries to the framework itself permanently - I mean we already know that ssh is flaky, but don't really know why - so while retries will possibly make the tests pass, they won't address the root of the problem.

You're hard pressed to convince me that adding retries here *isn't* a good idea when testing distributed systems. If this does end up addressing the issues, that means that we simply needed exponential backoff when calling commands over SSH - if it doesn't, well it makes the framework more resilient anyways.

> Especially them not being is what doesn't exactly feel right.

I don't understand what you're saying here.

> Also, this doesn't address ssh_capture and ssh_output methods.

Sure, I'll add these.

> Also I wonder if we could adjust a connection timeout on self.ssh_client.exec_command and self.ssh_client.open_session and the like instead of retries.

I acknowledge that this is me taking shots in the dark to try and address a problem we don't fully understand, but we should be methodical with adjusting changes here. Retrying IMO is more clean than increasing TCP timeouts, and we should try that before we increase timeouts.

setup.py
```diff
@@ -59,7 +59,8 @@ def run_tests(self):
                         'more-itertools==5.0.0',
                         'tox==3.13.2',
                         'six==1.12.0',
-                        'PyYAML==5.3'],
+                        'PyYAML==5.3',
+                        'retrying==1.3.3'],
```
ewencp, line 63: Not necessarily an issue since this is a simple library, but just a heads up that this hasn't seen a release since 2014.

ducktape/cluster/remoteaccount.py
stan-confluent: >Especially them not being is what doesn't exactly feel right.
Especially them not being *adjustable* is what doesn't exactly feel right. :)

ewencp: I think it's fine to add some retries, definitely up for debate though is what the max timeouts should be. Tests can already take quite a long time, adding retries only extends that, and depending on how fast the ssh failures are could make them take a *lot* longer.

I think this boils down to the tradeoff between test cost in the face of network failures vs developer time. For a change like this, it'd probably be a good idea to get an idea of what execution time changes look like over a statistically significant number of runs with the change.

stan-confluent: >adding retries here isn't a good idea when testing distributed systems
Yeah, nothing against retries in general, just not sure adding unconditional and unconfigurable retries on the lowest-level api method in the framework is a good idea. Consider the (hypothetical) case where a test would expect ssh connection to fail. In most cases you probably want some retries, but you should be able to adjust them or turn them off.
>If this does end up addressing the issues, that means that we simply needed exponential backoff when calling commands over SSH - if it doesn't, well it makes the framework more resilient anyways.
I would argue that we need to understand the nature and the cause of our issues with ssh before applying solutions. If it's actually flaky then retries would help, if the issue is of a different kind, then this will simply make the test run longer and fail anyways. 
Again, nothing against retries in general, but we need to understand the problem before applying the solution, all I'm saying :)

setup.py
andrewegel: I saw this too, but that should speak to how well the library works: There hasn't been a bug fix or feature addition in 6 years because of how well it *does* things already.

andrewegel:
It sounds like no one is really in favor of this change.

setup.py
stan-confluent: I used a version of this library from the same author for Java (guava-retrying) and can attest that it works quite well for java - not necessarily any indication of the python implementation, of course, but some data point at least.

stan-confluent:
>It sounds like no one is really in favor of this change.

I think we can keep this change if we can make retries configurable (timeout, backoff or turn them off) from the kafkatest/muckrake code, without having to rebuild ducktape. Maybe something like `node.account.set_ssh_retry_strategy(timeout, backoff)`? Not sure what value should be default though.

But before merging this one, we need to understand the root cause of the recent spike in ssh issues (which seems to be paramiko version?) - since it would at the very least affect the default retry timeout/backoff numbers we'd use.

sdanduConf:
> It sounds like no one is really in favor of this change.

I am in favor of this change generally. I think suggestions on configurability and timeout values are valid, but there is definite flakiness we see when running across a large number of runs and tests with a significant number of workers, so reduction of that flakiness makes sense to me.  I'd suggest we continue this or try a few test runs just to see if there is an obvious increase in runtime.

ewencp:
Yeah, I like the idea, just details need to be ironed out. It's just one of those things where it's so much at the heart of all operations that it's tricky to modify and hard to understand what all the fallout could be.

PR #241, opened 2020-08-04
Fix python3 annoying false positive import error.

### LINKED ISSUE #240: Python3 annoying false error output in ducktape.tests.loader.TestLoader._import_module
In python 2 string representation of `ImportError` looks like:
```
ImportError: No module named home.ivandasch.Job.ducktape.systests.cluster.test_remote_account

```

Since in python3 string representation changes:
```
ImportError: No module named 'home'
```

So original regex can be changed to cover both python 2 and python 3
```
match = re.search(r"No module named '?([^\s\']+)'?", str(e))

if match is not None:
    missing_module = match.groups()[0]

    if missing_module in module_name:
        expected_error = True
```

ivandasch:
Fix #240

### DIFF
--- a/ducktape/tests/loader.py
+++ b/ducktape/tests/loader.py
@@ -273,12 +273,12 @@ class TestLoader(object):
                 # import line), or a SyntaxError.
                 expected_error = False
                 if isinstance(e, ImportError):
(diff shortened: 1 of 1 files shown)


ivandasch:
#240

PR #251, opened 2020-09-07
Fix: set DucktapeJSONEncoder as default encoder in reports.

### LINKED ISSUE #250: Set DucktapeJSONEncoder as default in reports generator.
Currently, there is inconsistency with json encoder in `tests/reporter.py`. Some methods use it, some uses default.
I.e. `JSONReporter` use `DucktapeJSONEncoder`, but other encoders not.

Lets introduce some alias to json.dumps i.e.
```
def json_dumps(obj, **kwargs):
    """
    Dump object to json string with Ducktape json encoder.
    """
    assert 'cls' not in kwargs, "Passing custom json encoder is forbidden."

    return json.dumps(obj, cls=DucktapeJSONEncoder, **kwargs)
```
and use it everywhere in `tests/reporter.py`

ivandasch:
Fix #250

### DIFF
--- a/ducktape/cluster/cluster_spec.py
+++ b/ducktape/cluster/cluster_spec.py
@@ -14,9 +14,8 @@
 
 from __future__ import absolute_import
 
-import json
-
 from ducktape.cluster.node_container import NodeContainer
+from ducktape.json_serializable import json_dumps, json_loads
 
 LINUX = "linux"
 
@@ -40,7 +39,7 @@ class NodeSpec(object):
         dict = {
             "os": self.operating_system,
         }
-        return json.dumps(dict, sort_keys=True)
+        return json_dumps(dict, sort_keys=True)
 
 
 class ClusterSpec(object):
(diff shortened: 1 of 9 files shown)


ewencp (commented):
Did one of the cases not using `DucktapeJSONEncoder` cause a problem?

imo we should really only go one of two ways here. The current approach is that cases where we expect complex data structures that the `json` module won't handle them naturally provide the override. This is used for cases where, e.g., we know we have a `TestResult` which is a "real" class rather than just a dumb data container. The danger is that we might miss a case, but to date, I haven't heard of any issues with this.

The second would be to be very defensive and support the `to_json` extension everywhere. If we wanted to do that, we should make it fully comprehensive -- in `json_serializable.py` we should provide the type of wrapper you have here and use it universally, probably banning direct imports of `json` everywhere else. That guarantees we cover not only the reporter, but also, e.g., `result.py`, `vagrant.py`, etc.

ivandasch:
@ewencp Hi, Ewen
>> Did one of the cases not using DucktapeJSONEncoder cause a problem?
Currently, if you try to use object in @parametrize, test will fail. Namely, we use subclass of `LooseVersion` and we always should convert it to str in our project.
Also, we created custom pararemetrization and pass it from --globals (run all suite against specific version or against version vs version in compatibility suite). 

I suppose that it is possible to remove this restriction and add ability to user pass object with to_json overriden method to parameter.

>> The second would be to be very defensive and support the to_json extension everywhere.
So you suggests to move my helper function to distinct module and use it everywhere? Ok, sounds good for me.

ivandasch:
@ewencp Hi, I changed my PR as you suggested (second variant). Could you please look at it again?

stan-confluent:
pardon my trigger finger :)

stan-confluent:
I'm not super familiar with all of ducktape yet, can you please clarify how using DucktapeJSONEncoder helps with this?
>Currently, if you try to use object in @parametrize, test will fail. Namely, we use subclass of LooseVersion and we always should convert it to str in our project.

?

PR #265, opened 2021-01-19
Merge changes from 07x to 08x

stan-confluent:


setup.py
```diff
@@ -55,6 +55,8 @@ def run_tests(self):
                         'paramiko~=2.7.2',
                         'pyzmq==19.0.2',
                         'pycryptodome==3.9.8',
+                        'bcrypt==3.1.7',
+                        'pysistence==0.4.1',
```
andrewegel, line 59: Don't need this since 0.8.x is Py3x compatible, these pins were for Py2.7

andrewegel (changes requested):
Undo the bcrypt / pysistence pins, those were for Py2.7 compatibility

setup.py
stan-confluent: Do we officially state that 0.8.x drops support for 2.7? I'm all in favor for doing so for the record :)

stan-confluent:
@andrewegel done

setup.py
andrewegel: 0.7.x line will be the last Py2.7 compatibility branch. 0.8.x is for python3. "Official" wording and enforcement is something else (You would want to set `python_requires >= 3.6` in setup.py to "officially" drop Py2.7 support) - And likely some kind of documentation note.

andrewegel (approved):

PR #266, opened 2021-01-20
Support specifying test params in test discovery symbols

stan-confluent:
You can now append `@{"x": 100}` to a test discovery symbol after a method.
This behaves as a filter as opposed to `--parameters` flag which directly passes the params.
For example, this method:
```
@matrix(x=[1, 2, 3])
def some_test(x):
   ...
```
you can run with `@{"x": 1}` but not with `@{"x": 1000}` because there is no such param declared in the `@matrix` annotation. It won't crash, but simply won't find test that matches.

There is, of course, a limitation that parameters must be json-compatible for this to work, similarly to `--parameters` flag.

Additionally:
- removed py27 from tox and clarified that only python 3.7+ is supported. This has been the case for 0.8.x branch already, but not clarified anywhere.
- updated the docs
- fixed a test that was failing on python3.8 due to different behavior of the `@patch` annotation - and the test wasn't that good in the first place, it is more focused now.

### DIFF
--- a/docs/install.rst
+++ b/docs/install.rst
@@ -4,7 +4,9 @@
 Install
 =======
 
-1. Install `cryptography`_ (used by `paramiko` which Ducktape depends on), this may have non-python external requirements
+1. Ducktape requires python 3.7 or later.
+
+2. Install `cryptography`_ (used by `paramiko` which Ducktape depends on), this may have non-python external requirements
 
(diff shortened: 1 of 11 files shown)


imcdo (approved):
LGTM.  looking forward to this feature! Thanks Stan

PR #272, opened 2021-01-28
Print a warning if test is not using all requested nodes

stan-confluent:
This should allow us to monitor sub-optimal usage of cluster resources.


nikesh0822 (approved):
lgtm

PR #275, opened 2021-02-10
Fix 'defaults' mark to preserve cluster metadata

nikesh0822:
**Existing Implementation:**

If the `@default` mark is specified, and if there is a `@cluster` with `num_nodes` before default, this is overridden in cases when matrix or parametrized has `@cluster` mark. 
Example:
```
  @cluster(num_nodes=5)
  @defaults(y=[5])
  @parametrize(x=100, y=200)
  @cluster(num_nodes=3)
  @parametrize(x=10, y=20)
  @parametrize(x=30, y=40)
  @cluster(num_nodes=7)
  @matrix(x=[1, 2])
  def test_thing(self, x, y):
      pass
```
since we have default num_nodes as 5, even though matrix and parameterize have 7 and 3 respectively, it's ignored and hence all 5 tests above will request 5 nodes instead.

**Modified Implementation:**

This PR fixes the above issue and the output for the above scenario after fix would be:

```
 {'x': 1, 'y': 5} -> using matrix --> uses  7 nodes
 {'x': 2, 'y': 5} -> using matrix --> uses 7 nodes
 {'x': 30, 'y': 40} -> using parametrize --> uses 3 nodes
 {'x': 10, 'y': 20} -> using parametrize --> uses 3 nodes
 {'x': 100, 'y': 200} -> using parametrize (default) --> uses 5 nodes
```

### DIFF
--- a/ducktape/mark/_mark.py
+++ b/ducktape/mark/_mark.py
@@ -159,10 +159,12 @@ class Defaults(Mark):
                 for injected_args in cartesian_product_dict(
                         {arg: self.injected_args[arg] for arg in self.injected_args if arg not in ctx.injected_args}):
                     injected_args.update(ctx.injected_args)
-                    if ctx.cluster_use_metadata:
-                        seed_context.cluster_use_metadata = ctx.cluster_use_metadata
                     injected_fun = _inject(**injected_args)(seed_context.function)
-                    new_context_list.insert(0, seed_context.copy(function=injected_fun, injected_args=injected_args))
+                    new_context = seed_context.copy(
+                        function=injected_fun,
+                        injected_args=injected_args,
+                        cluster_use_metadata=ctx.cluster_use_metadata)
+                    new_context_list.insert(0, new_context)
         else:
             for injected_args in cartesian_product_dict(self.injected_args):
                 injected_fun = _inject(**injected_args)(seed_context.function)
--- a/ducktape/mark/_mark.py
+++ b/ducktape/mark/_mark.py
@@ -159,6 +159,8 @@ class Defaults(Mark):
                 for injected_args in cartesian_product_dict(
                         {arg: self.injected_args[arg] for arg in self.injected_args if arg not in ctx.injected_args}):
                     injected_args.update(ctx.injected_args)
+                    if ctx.cluster_use_metadata:
+                        seed_context.cluster_use_metadata = ctx.cluster_use_metadata
                     injected_fun = _inject(**injected_args)(seed_context.function)
                     new_context_list.insert(0, seed_context.copy(function=injected_fun, injected_args=injected_args))
         else:
--- a/tests/loader/resources/loader_test_directory/test_decorated.py
+++ b/tests/loader/resources/loader_test_directory/test_decorated.py
@@ -13,8 +13,11 @@
 # limitations under the License.
 
 from ducktape.tests.test import Test
+from ducktape.mark.resource import cluster
 from ducktape.mark import matrix
 from ducktape.mark import parametrize
+from ducktape.mark import defaults
+
 
 NUM_TESTS = 15
 
@@ -43,4 +46,15 @@ class TestParametrized(Test):
     @parametrize(x="abc", y=[])
     def test_thing(self, x, y):
         """2 tests"""
+        pass
+
+class TestStackedDefaultClusterMatrixParametrized(Test):
+    @cluster(num_nodes=5)
+    @defaults(y=[1])
+    # @cluster(num_nodes=3)
+    @parametrize(x=[1])
+    @cluster(num_nodes=7)
+    @matrix(x=[4])
+    def test_thing(self, x, y):
+        print("works!!!")
         pass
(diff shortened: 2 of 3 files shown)


imcdo (approved):
LGTM

ducktape/mark/_mark.py
```diff
@@ -159,6 +159,8 @@ def apply(self, seed_context, context_list):
                 for injected_args in cartesian_product_dict(
                         {arg: self.injected_args[arg] for arg in self.injected_args if arg not in ctx.injected_args}):
                     injected_args.update(ctx.injected_args)
+                    if ctx.cluster_use_metadata:
+                        seed_context.cluster_use_metadata = ctx.cluster_use_metadata
```
stan-confluent, line 163: Modifying the function argument doesn't look like a good idea, especially since this seed context will be passed to other mark implementations. Can you explain why the fix is the way it is and why do you need to modify seed context in defaults but not in other marks?
(also, you don't do anything in the `else` where context_list is none)

stan-confluent: I'm not sure that functionality we rely upon so much (the order in which marks are applied and how multiple cluster annotations are applied) is actually by design - but could be wrong

nikesh0822: Sure, 
In the current implementation, since cluster mark is on top followed by defaults mark, the named attribute we get from object is (say)
```
[<ducktape.mark._mark.Matrix object at 0x7f9a691a8e20>, 
<ducktape.mark.resource.ClusterUseMetadata object at 0x7f9a691a8d30>, 
<ducktape.mark._mark.Parametrize object at 0x7f9a691a8d00>, 
<ducktape.mark._mark.Parametrize object at 0x7f9a691a8cd0>, 
<ducktape.mark.resource.ClusterUseMetadata object at 0x7f9a691a8c70>, 
<ducktape.mark._mark.Parametrize object at 0x7f9a691a8ca0>, 
<ducktape.mark._mark.Defaults object at 0x7f9a691a8c40>, 
<ducktape.mark.resource.ClusterUseMetadata object at 0x7f9a691a8c10>]
```
when we reach to ‘defaults’ in [this for loop here](https://github.com/confluentinc/ducktape/blob/v0.7.12/ducktape/mark/mark_expander.py#L42-L44), we execute [this mark ‘apply’ for ‘default’](https://github.com/confluentinc/ducktape/blob/6a8d9c28cf62227761e4e531c09b4c95b4d3104c/ducktape/mark/_mark.py#L155-L171) piece of code.
Here, we pass the context_list and seed_context
Even though we have our cluster metadata in context_list since this is the ‘default’ case, we define an empty new context list [here](https://github.com/confluentinc/ducktape/blob/6a8d9c28cf62227761e4e531c09b4c95b4d3104c/ducktape/mark/_mark.py#L156), get injected args from the cartesian product, and set the cluster metadata to empty because cluster metadata for `seed_context` is `{}` and insert that into the new context list [here](https://github.com/confluentinc/ducktape/blob/6a8d9c28cf62227761e4e531c09b4c95b4d3104c/ducktape/mark/_mark.py#L165 and return that new context list https://github.com/confluentinc/ducktape/blob/6a8d9c28cf62227761e4e531c09b4c95b4d3104c/ducktape/mark/_mark.py#L171) opposed to updating the initial context text as we do at other places...
Parametrize: https://github.com/confluentinc/ducktape/blob/6a8d9c28cf62227761e4e531c09b4c95b4d3104c/ducktape/mark/_mark.py#L189
Matrix: https://github.com/confluentinc/ducktape/blob/6a8d9c28cf62227761e4e531c09b4c95b4d3104c/ducktape/mark/_mark.py#L130
After this, the returned context list would have empty cluster metadata. In the next iteration, we would be going over `<ducktape.mark.resource.ClusterUseMetadata object at 0x7f9a691a8c10>` and this looks for `@cluster(num_nodes=...)` that is above the default and apply that to the entire context. If `@cluster(num_nodes=...)`  above `@defaults` is not defines, it will take it as empty and use the entire cluster. 

In this PR, 
we check if [our context](https://github.com/confluentinc/ducktape/blob/6a8d9c28cf62227761e4e531c09b4c95b4d3104c/ducktape/mark/_mark.py#L162) has cluster metadata, we use that `ctx.cluster_use_metadata` instead of `{}` and solve our problem. 
we don't need an `else` because if our `ctx.cluster_use_metadata` is `{}`, then it's the same as using seed_context.cluster_use_metadata which is `{}`

stan-confluent: Yeah, I understand that defaults creates new contexts based on the cartesian product of the context args x default args. And you're right, we don't need to do anything if context_list is empty.

What concerns me is that modifying seed_context doesn't look right.

_(I'll leave aside the the question of why not just do `ctx.copy` instead of `seed_context.copy` - it was like this before, could've been made like that for a reason; or not)_

Why not smth like this?
```python
new_context = seed_context.copy(
    function=injected_fun, 
    injected_args=injected_args,
    cluster_use_metadata=ctx.cluster_use_metadata)
new_context_list.insert(0, new_context)
```

nikesh0822: Agreed. probably for the best 👍 
Pushed commit with updated changes. Thanks!

stan-confluent (approved):
LGTM, thanks Nikesh!

PR #276, opened 2021-02-10
Allow for yaml cluster files to import other yaml cluster files

imcdo:
Currently cluster files can describe a spec for a test suite.  This PR adds the ability to import the suites from other cluster files as well. This will allow users to create sub suites that can be included in larger suites.

Example:
```
# in suite_a.yml
suite-a:
  - test-a.py
  - test-b.py
  
# in suite_b.yml
import:
   - ./suite-a.yml
suite-b:
  -  test-c.py

```
Now ducktape will run both test-suite-a and suite-b when ran against suite_b.yml


imcdo:
> [clabot:check]

nikesh0822 (approved):
Awesome!
LGTM.

nikesh0822:
I see you've added a test calling itself. Would be good to add another cyclic test too. main_suite -> s1, s2 ; s1 -> s2 ; s2 -> s1
This is fine too :)

ducktape/tests/loader.py
```diff
@@ -454,23 +457,45 @@ def _load_test_suite_files(self, test_suite_files):
...
+            raise LoaderException("Malformed test suite file: " + suite_file_path)
+
+        for suite_name, suite_content in file_content.items():
+            if not suite_content:
+                raise LoaderException("Empty test suite " + suite_name + " in " + suite_file_path)
+        return file_content
+
+    def _read_test_suite_from_file(self, root_suite_file_path):
+        suites = []
+        root_suite_file_path = os.path.abspath(root_suite_file_path)
```
stan-confluent, line 479: It is possible to pass multiple test suite files into the loader, so this can be optimized a bit by adding all the suite files in the stack at once.
I would also separate these two steps - one step should walk the graph and build a set of unique files to load; and a second one would simply load them. This is not critical, but might make the code simpler and functions shorter.

imcdo: the issue with that is to walk the graph it needs to load the file. (hence i cache the loaded files to be accessed later) 
I will go ahead and load all files at once good catch.

nikesh0822 (approved):
lgtm

tests/loader/resources/loader_test_directory/test_suite_cyclic_a.yml
```diff
@@ -0,0 +1,9 @@
+import:
+  - test_suite_cyclic_b.yml # 1 test
+
+test_suite_cyclic_a:
+  - ./sub_dir_a/test_c.py # 1 test
+  - ./test_b.py # 3
+# suite total: 4
+
+# total: 5
```
stan-confluent, line 9: Nit: no new line at the end of this one and other test files

ducktape/tests/loader.py
stan-confluent: Gotcha, yeah, that makes sense. 
It can still be done in two steps - one reads the files and only parses the import statement, thus filtering duplicate files; another reads test suites from the loaded files - but in this case the benefits of splitting them are questionable :)

ducktape/tests/loader.py
```diff
@@ -446,31 +452,52 @@ def _is_test_function(self, function):
...
+        files = {file: self._load_suite(file) for file in root_suite_file_paths}
+        stack = root_suite_file_paths
+
+        while len(stack) != 0:
+            curr = stack.pop()
+            loaded = files[curr]
+            if 'import' in loaded:
+                directory = os.path.dirname(curr)
+                # apply path of current file to the files inside
+                abs_file_iter = (os.path.abspath(os.path.join(directory, file))
```
stan-confluent, line 492: what if my test suite imports by absolute path?
```
import:
 - /Users/test/suite.yml
```
?

ducktape/tests/loader.py
```diff
@@ -446,31 +452,52 @@ def _is_test_function(self, function):
...
+                                    for file in loaded.get('import', []))
+                imported = [file for file in abs_file_iter if file not in files]
+                for file in imported:
+                    files[file] = self._load_suite(file)
+                stack.extend(imported)
+                del files[curr]['import']
+
+        for test_suite_file_path, file_content in files.items():
```
stan-confluent, line 500: You could probably extract everything from line 501 onward into a separate method to make testing easier, but I think all our unit tests load full suite files anyway, so not a big deal.

```python
    for test_suite_file_path, file_content in files.items():
        suites.extend(_load_suite(test_suite_file_path, file_content))

# rename existing _load_suite into _load_file since it doesn't actually load suite, only a full file content.
def _load_suite(file_path, file_content):
    # everything from line 501:
    for suite_name, suite_content in file_content.items():
        .....
```

stan-confluent (approved):
lgtm, thanks Ian!
Couple of minor suggestions, would be good to add tests that import:
- absolute paths, 
- parent paths (`../smth.yml`) 
- non-existent files
- files that are not suites
- importing whole folders
- etc

ducktape/tests/loader.py
ewencp: `os.path.join` will use the second parameter only if it's an absolute path, so this should work. Testing is a bit more annoying since the yaml files can't be pre-baked...

ewencp (commented):
The general idea seems good to me (had discussed with @stan-confluent previously), and I think @stan-confluent caught what I would have flagged. So generally lgtm

ducktape/tests/loader.py
imcdo: yeah not sure how I'm supposed to test absolute paths.

ewencp: There would be a couple of options I can think of:

1) Construct yamls dynamically instead of statically and use a tmp directory
2) Mock the file loading code to redirect as long as the expected absolute path matched
3) Decide it's not worth the testing effort and skip those tests. I'm not sure we even *should* support absolute paths. The only reason I can think of to support them is if you need to get a test suite from another project (e.g. CP tests use kafka tests), but I don't think we've planned or designed around that possibility.

imcdo: I have added a dynamic absolute path test.

nikesh0822 (approved):
Thanks for the tests. LGTM

tests/loader/check_loader.py
```diff
@@ -111,6 +113,27 @@ def check_test_loader_raises_on_invalid_test_suite(self, suite_file_path):
...
+        pytest.param(5, [
+            # test that a cyclic import doesn't cause an infinate loop
+            os.path.join(discover_dir(), 'test_suite_cyclic_b.yml')
+        ], None, id='self load in import'),
+        pytest.param(5, [
+            # test that a cyclic import starting with second import
+            os.path.join(discover_dir(), 'test_suite_cyclic_a.yml')
+        ], None, id='self load in import'),
+        pytest.param(5, [
+            # test single import statement with parnet reference
```
stan-confluent, line 134: typo `parnet`

tests/loader/check_loader.py
```diff
@@ -111,6 +113,27 @@ def check_test_loader_raises_on_invalid_test_suite(self, suite_file_path):
...
+            # loads a suite with an imported statement that imports the suites in the
+            # imported file as well
+            os.path.join(discover_dir(), 'test_suite_with_single_import.yml')
+        ], None, id='load dependency path'),
+        pytest.param(2, [
+            # test that a self import doesn't cause an infinate loop
+            os.path.join(discover_dir(), 'test_suite_with_self_import.yml')
+        ], None, id='self load in import'),
+        pytest.param(5, [
+            # test that a cyclic import doesn't cause an infinate loop
```
stan-confluent, line 126: typo `infinite`, here and above

stan-confluent (approved):
Thanks Ian!

imcdo:
https://confluentinc.atlassian.net/browse/QEC-4483

PR #277, opened 2021-02-18
Wait until exception

stan-confluent:
This PR modifies wait_until behavior so that if the underlying condition raises an exception, wait_until would continue retrying. This is enabled via setting retry_on_exc parameter to True. Default value is False, to maintain backward compatibility.


ducktape/utils/util.py
```diff
@@ -19,26 +19,42 @@
...
-def wait_until(condition, timeout_sec, backoff_sec=.1, err_msg=""):
+def wait_until(condition, timeout_sec, backoff_sec=.1, err_msg="", retry_on_exc=False):
     """Block until condition evaluates as true or timeout expires, whichever comes first.
 
     :param condition: callable that returns True if the condition is met, False otherwise
     :param timeout_sec: number of seconds to check the condition for before failing
     :param backoff_sec: number of seconds to back off between each failure to meet the condition before checking again
     :param err_msg: a string or callable returning a string that will be included as the exception message if the
                     condition is not met
+    :oaram retry_on_exc: if True, will retry if condition raises an exception. If condition raised exception on last
```
ewencp, line 30: typo on `:param`

ducktape/utils/util.py
```diff
@@ -19,26 +19,42 @@
...
+                         iteration, that exception will be raised as a cause of TimeoutError.
+                         If False and condition raises an exception, that exception will be forwarded to the caller
+                         immediately.
+                         Defaults to False (original ducktape behavior).
```
ewencp, line 34: Not sure where you want to put it, but might want to add a TODO with a version number to start tracking breaking changes we might want to make eventually if we ever bump past 1.0

ducktape/utils/util.py
```diff
@@ -19,26 +19,42 @@
...
-        if condition():
-            return
-        else:
+        try:
+            if condition():
+                return
+            else:
+                # reset last_exception if last iteration didn't raise any exception, but simply returned False
+                last_exception = None
+        except Exception as e:
```
ewencp, line 48: I think all the exceptions (hah!) of errors that inherit from BaseException instead of Exception are ok to skip here, but should double check that we don't want to cast an even wider net.

ewencp (approved):
a few minor comments, but lgtm

imcdo (approved):
LGTM thanks for the feature.

ducktape/utils/util.py
stan-confluent: Good idea, added a TODO right here; we can simply use `# TODO: [1.0.0]`

ducktape/utils/util.py
stan-confluent: Good idea, I don't think it'll hurt to catch a BaseException here, changed

PR #278, opened 2021-02-24
Fix failure to load if discovery symbol characters in json

imcdo:
Currently in ducktape if you try to load a test with characters used in the loading process in your json, the loader will fail to discover tests.
ex:
`dummy_test.py::DummyTest.test@{'version':'1.0.0.'}` - fails to load with an `Invalid discovery symbol:` loader error

to fix this, I manually split by the rightmost occurrence of the split character, and allow the failures to occur when loading json.

### DIFF
--- a/ducktape/tests/loader.py
+++ b/ducktape/tests/loader.py
@@ -230,34 +230,20 @@ class TestLoader(object):
             "path/to/test_file.py" -> ("path/to/test_file.py", "", "")
             "path/to/test_file.py::ClassName.method" -> ("path/to/test_file.py", "ClassName", "method")
         """
+        def divide_by_symbol(ds, symbol):
+            if symbol not in ds:
+                return ds, ""
+            idx = ds.index(symbol)
+            return ds[:idx], ds[idx + len(symbol):]
+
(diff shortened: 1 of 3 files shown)


ducktape/tests/loader.py
```diff
@@ -230,34 +230,20 @@ def _parse_discovery_symbol(self, discovery_symbol, base_dir=None):
             "path/to/test_file.py" -> ("path/to/test_file.py", "", "")
             "path/to/test_file.py::ClassName.method" -> ("path/to/test_file.py", "ClassName", "method")
         """
+        def divide_by_symbol(ds, symbol):
```
stan-confluent, line 233: python's str.split() method already has this functionality - see `maxsplit` param:
https://docs.python.org/3/library/stdtypes.html#str.split

stan-confluent (approved):
Thanks Ian, that's a really good catch!
Left one suggestion, otherwise the fix looks good conceptually.

ducktape/tests/loader.py
imcdo: cool!, I'll go ahead and use it.

nikesh0822 (approved):
LGTM 👍

PR #279, opened 2021-03-03
Fail test option when node utilization isn't optimal

imcdo:
This PR allows for users to specify that improper node utilization causes a failure.  Currently if you define a cluster that requests more nodes than a test actually uses ducktape logs a warning specifying the number of nodes miss-match.  This PR allows you to specify the `--fail-bad-cluster-utilization` when running tests, which fails the test if the node utilization is miss matched.

### DIFF
--- a/ducktape/tests/runner_client.py
+++ b/ducktape/tests/runner_client.py
@@ -180,10 +180,10 @@ class RunnerClient(object):
             self.test = None
 
     def _check_cluster_utilization(self, result, summary):
-        """Pass through results and summary and modify the values if
-        the number of nodes used by a test is less than the number of nodes
-        requested by a test.  Will also print a warning if the test passes and
-        the node utilization doesn't match.
+        """Checks if the number of nodes used by a test is less than the number of
+        nodes requested by the test. If this is the case and we wish to fail
+        on bad cluster utilization, the result value is failed. Will also print
+        a warning if the test passes and the node utilization doesn't match.
         """
         max_used = self.cluster.max_used()
         total = len(self.cluster.all())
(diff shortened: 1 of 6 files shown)


ducktape/tests/runner_client.py
```diff
@@ -152,6 +153,10 @@ def run(self):
                 if service_errors:
                     summary += "\n\n" + service_errors
 
+            # only check node utilization on test pass
+            if (test_status == PASS):
+                test_status = self._check_cluster_utilization(test_status)
```
stan-confluent, line 158: Please also update the `summary` field with the message.
You can also do it for failed tests as well - just make sure the test summary includes both the original failure and the utilization message

nikesh0822 (approved):
LGTM. Add doc string to `_check_cluster_utilization` would help.

ducktape/tests/runner_client.py
```diff
@@ -176,11 +179,26 @@ def run(self):
             self.test_context = None
             self.test = None
 
-    def _check_cluster_utilization(self):
+    def _check_cluster_utilization(self, result, summary):
+        """Pass through results and summary and modify the values if
```
ewencp, line 183: This could probably be a bit clearer -- it's not obvious from the comment what "modify the values mean" and the subsequent sentence makes it sound like it'll only ever warn.

ewencp (approved):
one nit on comment clarity, but otherwise lgtm

ducktape/tests/runner_client.py
imcdo: Thanks for the feedback hopefully it is clearer now

PR #280, opened 2021-03-22
add timeout parameter and set it to 0mq socket

imcdo:
add a timeout to test runners recv method, so if a test fails to send a message to the receiver, the receiver will throw a timeout error.

### DIFF
--- a/tests/runner/check_sender_receiver.py
+++ b/tests/runner/check_sender_receiver.py
@@ -1,0 +1,69 @@
+# Copyright 2021 Confluent Inc.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+# http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+from ducktape.cluster.localhost import LocalhostCluster
+from tests.ducktape_mock import test_context, session_context
+
+import logging
+import pytest
+from ducktape.tests.runner_client import Sender
+from ducktape.tests.runner import Receiver
+from ducktape.tests.event import ClientEventFactory, EventResponseFactory
+
+
+import multiprocessing as mp
+import os
+
+
+class CheckSenderReceiver(object):
+    def ready_response(self, client_id):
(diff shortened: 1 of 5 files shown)


ewencp (approved):
LGTM as long as the 1 hr setting is liberal enough given current test run times. We need to be a bit careful here since this only puts the timeout in place but doesn't make it configurable.

ewencp (commented):
Looks good, would be good to see if this is testable (we definitely have tests of `Runner`, but not sure we have any examples of mocking out the underlying `Receiver`'s 0mq client.

nikesh0822 (approved):
Apart from tests like Ewen suggested, adding timeout is good to return a failure. lgtm

tests/runner/check_sender_receiver.py
```diff
@@ -0,0 +1,69 @@
...
+
+
+import multiprocessing as mp
+import os
+
+
+class CheckSenderReceiver(object):
+    def ready_response(self, client_id):
+        sender_event_factory = ClientEventFactory("test_1", 0, client_id)
+        sender = Sender(server_host='localhost', server_port=8888,
```
ewencp, line 32: Using fixed ports can cause problems for tests. The port won't always be available. I can't remember how we're setup on Jenkins wrt docker networking, but anywhere there might be concurrent tests running (including in the container via parallel pytest), this could cause flakiness.

Usually if you're doing a real test rather than mocking out some pieces, you want the code to support accepting `0` as the port, then have an accessor to get the bound port. That way it'll just choose randomly and the only way you run into issues is if you just run out of ports entirely.

tests/runner/check_sender_receiver.py
```diff
@@ -0,0 +1,69 @@
...
+    def check_timeout(self):
+        client_id = "test-runner-{}-{}".format(os.getpid(), id(self))
+
+        receiver = Receiver(8888, 8888)
+        receiver.start()
+        try:
+            p = mp.Process(target=self.ready_response, args=(client_id,))
+            p.start()
+            with pytest.raises(TimeoutError):
+                receiver.recv(timeout=0)
```
ewencp, line 67: does 0 have special meaning here? I'd expect it to be `None`, but sometimes libraries inherit these details from the underlying C library, and using 0 is a pretty common timeout setting for "infinite"

imcdo: I think in this case it is the oposite, None indicates infinity, while 0 is a zero second timeout, so it should instantly fail

tests/runner/check_sender_receiver.py
imcdo: sounds good, Ill use the same port range that are used for the test runner.

Issue #284, opened 2021-11-03
Tests in explicitly provided files ignored


vp-elitnet:
After upgrade from 0.7.7 to 0.8.8 a regression was noticed - tests in explicitly provided files are not executed unless the file name has "test_" prefix or "_test" suffix. It seems the change was introduced as part of support for test suites (commit 2036577e). Pattern-based file discovery may be useful in a directory where test and non-test files are mixed. However, rejecting test files explicitly provided by the user (whatever the name) feels like an anti-feature. I suggest restoring the previous behaviour of not matching the names of explicitly provided files against the default pattern.

vp-elitnet:
Resolved by merging #285.

PR #285, opened 2021-11-03
Fix test file discovery regression

vp-elitnet:
Proposed fix for #284.

### DIFF
--- a/tests/loader/resources/loader_test_directory/name_does_not_match_pattern.py
+++ b/tests/loader/resources/loader_test_directory/name_does_not_match_pattern.py
@@ -1,0 +1,7 @@
+from ducktape.tests.test import Test
+
+class TestNotLoaded(Test):
+    """Loader should not discover this - module name does not match default pattern."""
+
+    def test_a(self):
+        pass
--- a/tests/loader/resources/loader_test_directory/name_does_not_match_pattern.py
+++ b/tests/loader/resources/loader_test_directory/name_does_not_match_pattern.py
@@ -1,5 +1,6 @@
 from ducktape.tests.test import Test
 
+
 class TestNotLoaded(Test):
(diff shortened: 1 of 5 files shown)


imcdo:
As many frameworks do (like for instance pytest) They filter files based on a prefixed or postfixed 'test' or something.  I would argue that this is ideal as it allows for not test files to not be collected around other test files, however i see your point that this could be a blocker if you wanted to change the pattern, so i would prefer making the prefix/postfix test file identifier be configurable as it is in something like pytest

imcdo:
looks good, but before this gets merged in, I would like it if you could add documentation for this behavior as it could be a bit unclear if someone can run a test individually but it not be collected when specifying its parent directory https://github.com/confluentinc/ducktape/blob/master/docs/run_tests.rst

stan-confluent:
Thanks for finding the issue and a suggested fix! Can you please also add a test to ensure that passing the folder with a mix of test and non-test files still behaves as expected (ie non-test files are ignored)?
Thanks!

vp-elitnet:
Added behaviour clarification in "Running Tests" section.

> Can you please also add a test to ensure that passing the folder with a mix of test and non-test files still behaves as expected (ie non-test files are ignored)?

Added a minimal additional mock test module to verify this case.

imcdo (approved):
Looks good thanks so much!

stan-confluent (approved):
Thank you!

tests/loader/resources/loader_test_directory/name_does_not_match_pattern.py
```diff
@@ -0,0 +1,7 @@
+from ducktape.tests.test import Test
+
```
imcdo, line 2: Nit: styling required for the pr job:
```suggestion


```

PR #287, opened 2022-01-18
Add SSH failure callbacks

imcdo:
One common point of failure for ducktape is paramiko, when a remote account fails, we generally get a non useful error message that has no attachment to the node that is being ran.  This PR addresses this issue by adding a flag to ducktape that allows you to specify a specific function that takes in an error and a remote account, that will then be ran on an ssh failure.  For instance, if you are running aws instances, you can write your own aws command that validates that the node is still running on aws, and then run this on an ssh failure.

### DIFF
--- a/docs/debug_tests.rst
+++ b/docs/debug_tests.rst
@@ -97,3 +97,13 @@ Tools for Managing Logs
 =======================
 
 Analyzing and matching up logs from a distributed service could be time consuming. There are many good tools for working with logs. Examples include http://lnav.org/, http://list.xmodulo.com/multitail.html, and http://glogg.bonnefon.org/.
+
+Validating Ssh Issues
+=======================
+
+Ducktape supports running custom validators when an ssh error occurs, allowing you to run your own validation against a host.
+this is done simply by running ducktape with the `--ssh-checker-function`, followed by the module path to your function, so for instance::
+    
+    ducktape my-test.py --ssh-checker-function my.module.validator.validate_ssh
+
+this function will take in the ssh error raised as its first argument, and the remote account object as its second.
--- a/ducktape/cluster/json.py
+++ b/ducktape/cluster/json.py
@@ -90,8 +90,9 @@ class JsonCluster(Cluster):
                     "Cluster json has a node without a ssh_config field: %s\n Cluster json: %s" % (ninfo, cluster_json)
 
                 ssh_config = RemoteAccountSSHConfig(**ninfo.get("ssh_config", {}))
-                remote_account = JsonCluster.make_remote_account(ssh_config, ninfo.get("externally_routable_ip"),
-                                                         ssh_exception_checks=kwargs.get("ssh_exception_checks"))
+                remote_account = \
+                    JsonCluster.make_remote_account(ssh_config, ninfo.get("externally_routable_ip"),
+                                                    ssh_exception_checks=kwargs.get("ssh_exception_checks"))
                 if remote_account.externally_routable_ip is None:
                     remote_account.externally_routable_ip = self._externally_routable_ip(remote_account)
                 self._available_accounts.add_node(remote_account)
--- a/ducktape/cluster/json.py
+++ b/ducktape/cluster/json.py
@@ -90,7 +90,8 @@ class JsonCluster(Cluster):
                     "Cluster json has a node without a ssh_config field: %s\n Cluster json: %s" % (ninfo, cluster_json)
 
                 ssh_config = RemoteAccountSSHConfig(**ninfo.get("ssh_config", {}))
-                remote_account = JsonCluster.make_remote_account(ssh_config, ninfo.get("externally_routable_ip"))
+                remote_account = JsonCluster.make_remote_account(ssh_config, ninfo.get("externally_routable_ip"),
+                                                                 ssh_exception_checks=kwargs.get("ssh_exception_checks"))
                 if remote_account.externally_routable_ip is None:
                     remote_account.externally_routable_ip = self._externally_routable_ip(remote_account)
                 self._available_accounts.add_node(remote_account)
@@ -100,15 +101,13 @@ class JsonCluster(Cluster):
         self._id_supplier = 0
 
     @staticmethod
-    def make_remote_account(ssh_config, externally_routable_ip=None):
+    def make_remote_account(ssh_config, *args, **kwargs):
         """Factory function for creating the correct RemoteAccount implementation."""
 
         if ssh_config.host and WINDOWS in ssh_config.host:
-            return WindowsRemoteAccount(ssh_config=ssh_config,
-                                        externally_routable_ip=externally_routable_ip)
+            return WindowsRemoteAccount(ssh_config, *args, **kwargs)
         else:
-            return LinuxRemoteAccount(ssh_config=ssh_config,
-                                      externally_routable_ip=externally_routable_ip)
+            return LinuxRemoteAccount(ssh_config, *args, **kwargs)
 
     def do_alloc(self, cluster_spec):
         allocated_accounts = self._available_accounts.remove_spec(cluster_spec)
--- a/ducktape/cluster/json.py
+++ b/ducktape/cluster/json.py
@@ -91,7 +91,7 @@ class JsonCluster(Cluster):
 
                 ssh_config = RemoteAccountSSHConfig(**ninfo.get("ssh_config", {}))
                 remote_account = JsonCluster.make_remote_account(ssh_config, ninfo.get("externally_routable_ip"),
-                                                                 ssh_exception_checks=kwargs.get("ssh_exception_checks"))
+                                                         ssh_exception_checks=kwargs.get("ssh_exception_checks"))
                 if remote_account.externally_routable_ip is None:
                     remote_account.externally_routable_ip = self._externally_routable_ip(remote_account)
                 self._available_accounts.add_node(remote_account)
--- a/ducktape/cluster/linux_remoteaccount.py
+++ b/ducktape/cluster/linux_remoteaccount.py
@@ -12,18 +12,19 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
+import logging
 from ducktape.cluster.cluster_spec import LINUX
 from ducktape.cluster.remoteaccount import RemoteAccount
 
 
 class LinuxRemoteAccount(RemoteAccount):
 
-    def __init__(self, ssh_config, externally_routable_ip=None, logger=None):
-        super(LinuxRemoteAccount, self).__init__(ssh_config, externally_routable_ip=externally_routable_ip,
-                                                 logger=logger)
+    def __init__(self, *args, **kwargs):
+        super(LinuxRemoteAccount, self).__init__(*args, **kwargs)
         self._ssh_client = None
         self._sftp_client = None
         self.os = LINUX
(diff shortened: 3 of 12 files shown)


ducktape/cluster/json.py
```diff
@@ -100,15 +101,13 @@ def __init__(self, cluster_json=None, *args, **kwargs):
         self._id_supplier = 0
 
     @staticmethod
-    def make_remote_account(ssh_config, externally_routable_ip=None):
+    def make_remote_account(ssh_config, *args, **kwargs):
```
sonupillai, line 104: This is a big change. Please make sure, all callers of this function are updated to access "externally_routable_ip"  config like a kw?

sonupillai (approved):
LGTM and as we discussed but I would recommend getting a second review from Ducktape Gods as well.

ducktape/cluster/json.py
imcdo: they dont need to, as args also get forwarded.

docs/debug_tests.rst
```diff
@@ -97,3 +97,13 @@ Tools for Managing Logs
 =======================
 
 Analyzing and matching up logs from a distributed service could be time consuming. There are many good tools for working with logs. Examples include http://lnav.org/, http://list.xmodulo.com/multitail.html, and http://glogg.bonnefon.org/.
+
+Validating Ssh Issues
```
stan-confluent, line 101: nit: SSH should be all caps, no?

ducktape/cluster/remoteaccount.py
```diff
@@ -26,6 +27,21 @@
...
+    def wrapper(self, *args, **kwargs):
+        try:
+            return method(self, *args, **kwargs)
+        except (SSHException, NoValidConnectionsError, socket.error) as e:
+            if self._custom_ssh_exception_checks:
+                self._log(logging.DEBUG, "starting ssh checks:")
+                self._log(logging.DEBUG, "\n".join(repr(f) for f in self._custom_ssh_exception_checks))
+                for func in self._custom_ssh_exception_checks:
+                    func(e, self)
+            raise
```
stan-confluent, line 43: should it be `raise e`? I don't remember if simple `raise` would preserve the original traceback

ducktape/command_line/main.py
```diff
@@ -181,7 +182,12 @@ def main():
         (cluster_mod_name, cluster_class_name) = args_dict["cluster"].rsplit('.', 1)
         cluster_mod = importlib.import_module(cluster_mod_name)
         cluster_class = getattr(cluster_mod, cluster_class_name)
-        cluster = cluster_class(cluster_file=args_dict["cluster_file"])
+
+        cluster_kwargs = {"cluster_file": args_dict["cluster_file"]}
+        checkers = [load_function(func_path) for func_path in args_dict["ssh_checker_function"]]
```
stan-confluent, line 187: There needs to be a readable error message if the function cannot be loaded for some reason.

ducktape/cluster/linux_remoteaccount.py
```diff
@@ -12,15 +12,15 @@
...
 from ducktape.cluster.cluster_spec import LINUX
 from ducktape.cluster.remoteaccount import RemoteAccount
 
 
 class LinuxRemoteAccount(RemoteAccount):
 
-    def __init__(self, ssh_config, externally_routable_ip=None, logger=None):
-        super(LinuxRemoteAccount, self).__init__(ssh_config, externally_routable_ip=externally_routable_ip,
-                                                 logger=logger)
+    def __init__(self, *args, **kwargs):
```
stan-confluent, line 22: there used to be a `logger` arg, but I don't see it being passed in in the `make_remote_account` method. Was anyone using it?

tests/cluster/check_remoteaccount.py
```diff
@@ -86,6 +93,18 @@ def check_wait_for_http_timeout(self):
             actual_timeout = time.time() - start
             assert abs(actual_timeout - timeout) / timeout < 1
 
+    def check_ssh_checker(self):
```
stan-confluent, line 96: I'm presuming other existing tests provide coverage for when there are no exception checks? Please make sure that we have covered the cases when there's more than one exception check, no exception checks or invalid exception check(s).

ducktape/command_line/parse_args.py
```diff
@@ -77,6 +77,10 @@ def create_ducktape_parser():
     parser.add_argument("--test-runner-timeout", action="store", type=int, default=1800000,
                         help="Amount of time in milliseconds between test communicating between the test runner"
                              " before a timeout error occurs. Default is 30 minutes")
+    parser.add_argument("--ssh-checker-function", action="store", type=str, nargs="+",
```
stan-confluent, line 80: Nit: Can this flag be specified multiple times or do you specify multiple functions after a single flag? It's not clear from the description

stan-confluent (commented):
Thanks Ian, this is great - great idea, clean implementation, 👏 
Couple of comments from me, mostly around testing and negative scenarios, and the usual batch of nitpicks, but looks good overall.

ducktape/cluster/remoteaccount.py
imcdo: i originally did `raise e` however when getting the traceback from pytest, it would show me the call stack from inside here, while with just raise, it gives the original stack trace, though i havn't tried this in python3

tests/cluster/check_remoteaccount.py
imcdo: No exception checks are the default case, so they are covered, will add the multiple case specificly

ducktape/command_line/main.py
imcdo: added to the load function method

ducktape/cluster/remoteaccount.py
imcdo: https://stackoverflow.com/questions/36153805/difference-between-raise-and-raise-e

ducktape/cluster/linux_remoteaccount.py
stan-confluent: Nah, I was mostly making sure that you don't get any unintentional consequences of passing all the args from the method above, but I don't see a real issue anywhere

ducktape/cluster/remoteaccount.py
stan-confluent: As long as it works fine in py3 (which is what newer ducktape versions are all targeting anyway), I'm fine with it.

ducktape/utils/util.py
```diff
@@ -53,3 +53,12 @@ def package_is_installed(package_name):
...
     """Return string representation of current ducktape version."""
     return __ducktape_version__
+
+def load_function(func_module_path):
+    """Loads and returns a function from a module path seperated by '.'s"""
+    module, function = func_module_path.rsplit(".", 1)
+    try:
+        return getattr(importlib.import_module(module), function)
+    except AttributeError:
+        raise Exception("Function could not be loaded from the module path {}, "
```
stan-confluent, line 63:
```python
except AttributeError as e:
    raise Exception(...) from e
```

imcdo: doesn't look like this is compatable with python2 :/ 
```
try:
...     raise Exception('hi')
... except Exception as e:
...     raise Exception('there') from e
  File "<stdin>", line 4
    raise Exception('there') from e
                                ^
SyntaxError: invalid syntax

```

stan-confluent: _mumbles swear words into the beard_

imcdo: python2 man >:(

ducktape/command_line/parse_args.py
stan-confluent: cool, as long as it is clear from the docs. I was confused because actual checkers on the cluster object are a list

stan-confluent (approved):

ducktape/cluster/remoteaccount.py
```diff
@@ -26,6 +27,21 @@
...
+def check_ssh(method):
+    def wrapper(self, *args, **kwargs):
+        try:
+            return method(self, *args, **kwargs)
+        except (SSHException, NoValidConnectionsError, socket.error) as e:
+            if self._custom_ssh_exception_checks:
+                self._log(logging.DEBUG, "starting ssh checks:")
+                self._log(logging.DEBUG, "\n".join(repr(f) for f in self._custom_ssh_exception_checks))
+                for func in self._custom_ssh_exception_checks:
+                    func(e, self)
```
ewencp, line 42: This should also be wrapped in a try/except if we support multiple checks.

ewencp (commented):
Does this actually need to be configurable, or is it something that can be baked in? Even if just providing a few options ootb?

The description is about the error logging, but that could be improved without a configurable validator. What type of validation would be happening other than trying to reconnect via the remote account object or just logging more detailed output? The main thing I can think of would be checks against e.g. a cloud provider, but since that info is injected after allocation atm, a check like that would need more substantial changes.

ducktape/cluster/remoteaccount.py
imcdo: The current assumption is that the validators will raise an exception if something is wrong.  Should this paradigm change?  Should it return some sort of result object?  We can also just try catch it but we can only raise one error in python2 (and python3 its a chain of exceptions which is a different assumption AFAIK)

imcdo:
> Does this actually need to be configurable, or is it something that can be baked in? Even if just providing a few options ootb?
I think this is better to be left as something that is configurable, as  ducktape isn't responsible for the environment at all and operates in the dark outside of ssh access.  Relying on the user to pass in validation about the environment when ssh fails as they are responsible for creating the environment allows for validation a better understanding of the environment.

ducktape/cluster/remoteaccount.py
ewencp: Yeah, you can only raise one exception, but you can make an exception type that wraps others. I just wasn't sure what types of checks you might do and in what order. Since this approach requires doing only the most specific checks, it can be more expensive (e.g. checking if a cloud instance is destroyed vs just doing a quick `nc -z` check first).

I'm not sure how many use cases there would be for using multiple checks other than a) fast -> slow ordering for checks and b) if you have multiple detectable failure modes. For (b), I'm not sure what that list would be.

ewencp (approved):
ok, guess it makes sense for the shortest path. the reason i ask is that one alternative would be a bigger change, but require less configuration -- tie the handler to the Cluster/ClusterNode and just have a native implementations for things like cloud providers. In particular, vagrant has been painful for long running clusters for a long time and we effectively bake certain implementations into ducktape anyway (in particular vagrant).

I don't mind making things pluggable, but where possible it'd be nice to keep the settings to a minimum if they're effectively going to be identical for a given cluster type. Another way to think of this in the current state is that if an instance disappears, vagrant status should be able to tell you that too, so even without adding custom cloud providers you could still bake the checks into the Cluster implementation.

More broadly, I was wasn't sure how this would work generally given the limited info that is passed along via the ClusterNode (other than trying some simpler tests like nc -z but that may fail just because of network connectivity, not only if the cloud resource was destroyed). It means you have to be able to backtrack to the original information about the node. That's possible in some cases, I'm not sure about all of them, and even if it is, could be relatively expensive.

We can make this change, but I do think we should continue to consider a more integrated and native approach for different cluster types. e.g. we had thought about a native aws implementation, but vagrant was the fastest route originally to support both local + aws.

PR #292, opened 2022-02-03
Removed tox and locked cryptography version

stan-confluent:
This is again to maintain python 2 compatibility. 

Cryptography is a transitive dep, specified by several of packages, none of which has an upper bound, so we lock it to the last py2-compatible version.
See https://cryptography.io/en/latest/changelog/#v3-4

Removing tox because it pulls virtualenv and that one pulls a bunch of stuff. But more importantly it is a test dependency, not a prod one, so I don't think it belongs to direct ducktape dependencies. And we don't run it in an automated fashion anywhere anyway.


imcdo (approved):

stan-confluent:
Merged without waiting for a PR build since it was simply a sync with 0.7.x which was also green.

PR #298, opened 2022-04-01
add `@ok_to_fail` decorator

### LINKED ISSUE #297: add decorator for tests that still runs test but not count result
It would be helpful to have a way to run tests that are known to be flakey but not have the test results be counted towards passing or failing in the final result. It is helpful to run flakey tests to see how often they fail in the context of all the other tests because the flakiness could be caused by the order of the test runs.

The `--exclude` functionality simply skips the tests and the `@ignore` decorator also does not run tests but at least mentions in the result tests not run which is helpful to have a quick summary of how many flakey tests were ignored.

In the old issue #112 there was a mention of a decorator called `@ok_to_fail` that seems to imply the functionality desired. The decorator approach helps keep the expected test behaviour close to the definition of the test.

Would need `reporter.py` to categorize the results of tests decorated as `@ok_to_fail`. I suggest having 2 additional test status `OPASS` and `OFAIL` to distinguish from existing test status `PASS`, `FAIL`, and `IGNORE`.

andrewhsu:
Closes #297

### DIFF
--- a/ducktape/mark/_mark.py
+++ b/ducktape/mark/_mark.py
@@ -107,6 +107,24 @@ class IgnoreAll(Ignore):
         self.injected_args = None
 
 
+class OkToFail(Mark):
+    """Run the test but categorize status as OPASS or OFAIL instead of PASS or FAIL."""
+
+    def __init__(self):
+        super(OkToFail, self).__init__()
+        self.injected_args = None
+
+    @property
+    def name(self):
+        return "OK_TO_FAIL"
+
+    def apply(self, seed_context, context_list):
+        assert len(context_list) > 0, "ignore annotation is not being applied to any test cases"
+        for ctx in context_list:
+            ctx.ok_to_fail = ctx.ok_to_fail or self.injected_args is None
+        return context_list
+
+
 class Matrix(Mark):
     """Parametrize with a matrix of arguments.
     Assume each values in self.injected_args is iterable
@@ -218,6 +236,7 @@ PARAMETRIZED = Parametrize()
 MATRIX = Matrix()
 DEFAULTS = Defaults()
 IGNORE = Ignore()
+OK_TO_FAIL = OkToFail()
 ENV = Env()
 
 
@@ -233,6 +252,11 @@ def parametrized(f):
 def ignored(f):
     """Is this function or object decorated with @ignore?"""
     return Mark.marked(f, IGNORE)
+
+
+def oked_to_fail(f):
+    """Is this function or object decorated with @ok_to_fail?"""
+    return Mark.marked(f, OK_TO_FAIL)
 
 
 def is_env(f):
@@ -409,6 +433,25 @@ def ignore(*args, **kwargs):
         return f
 
     return ignorer
+
+
+def ok_to_fail(*args, **kwargs):
+    """
+    Test method decorator which signals to the test runner to run test but set status as OPASS or OFAIL.
+    This will keep test results separate from the status PASS and FAIL.
+
+    Example::
+        @ok_to_fail
+        def the_test(...):
+            ...
+    """
+    if len(args) == 1 and len(kwargs) == 0:
+        # this corresponds to the usage of the decorator with no arguments
+        # @ok_to_fail
+        # def test_function:
+        #   ...
+        Mark.mark(args[0], OkToFail())
+        return args[0]
 
 
 def env(**kwargs):
--- a/ducktape/mark/_mark.py
+++ b/ducktape/mark/_mark.py
@@ -119,7 +119,7 @@ class OkToFail(Mark):
         return "OK_TO_FAIL"
 
     def apply(self, seed_context, context_list):
-        assert len(context_list) > 0, "ignore annotation is not being applied to any test cases"
+        assert len(context_list) > 0, "ok_to_fail annotation is not being applied to any test cases"
         for ctx in context_list:
             ctx.ok_to_fail = ctx.ok_to_fail or self.injected_args is None
         return context_list
--- a/ducktape/tests/reporter.py
+++ b/ducktape/tests/reporter.py
@@ -308,7 +308,7 @@ class HTMLSummaryReporter(SummaryReporter):
             'ignored_tests': ignored_result_string,
             'ofailed_tests': ofailed_result_string,
             'opassed_tests': opassed_result_string,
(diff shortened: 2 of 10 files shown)


ducktape/mark/_mark.py
```diff
@@ -107,6 +107,24 @@ def __init__(self):
...
+    def __init__(self):
+        super(OkToFail, self).__init__()
+        self.injected_args = None
+
+    @property
+    def name(self):
+        return "OK_TO_FAIL"
+
+    def apply(self, seed_context, context_list):
+        assert len(context_list) > 0, "ignore annotation is not being applied to any test cases"
```
imcdo, line 122: should be:
```suggestion
        assert len(context_list) > 0, "ok_to_fail annotation is not being applied to any test cases"
```

ducktape/tests/reporter.py
```diff
@@ -284,12 +299,16 @@ def format_report(self):
...
+            'num_ofailed': self.results.num_ofailed,
             'run_time': format_time(self.results.run_time_seconds),
             'session': self.results.session_context.session_id,
             'passed_tests': passed_result_string,
             'failed_tests': failed_result_string,
             'ignored_tests': ignored_result_string,
-            'test_status_names': ",".join(["\'%s\'" % str(status) for status in [PASS, FAIL, IGNORE]])
+            'ofailed_tests': ofailed_result_string,
+            'opassed_tests': opassed_result_string,
+            'test_status_names': ",".join(["\'%s\'" % str(status) for status in [PASS, FAIL, IGNORE, OPASS, OFAIL]])
```
imcdo, line 311: nit: brackets unnecessary
```suggestion
            'test_status_names': ",".join("\'%s\'" % str(status) for status in [PASS, FAIL, IGNORE, OPASS, OFAIL])
```

imcdo:
We might also want to think of a more robust solution to deal with flaky tests, just tracking them separately and marking them might not be the best way forward here.

andrewhsu:
@imcdo in the context of addressing flaky tests, `@ok_to_fail` can be useful in the situation where:
* The test itself is bug-free and desirable (if the test is suspect of causing the flakiness, then it should be fixed, `@ignore`, `exclude`, or removed)
* The flakey test failure is not critical enough to block all development on the codebase to fix the root cause
* Codebase that is being tested is frequently improved upon and the test may start passing reliably in the future

In practice, with a large enough codebase, there are usually several of these type of flakey tests and the combined flakiness add up to a percentage that is enough to slow down development, e.g. 5 flakey tests with each 10% chance of flakey failure.

This PR also updates the reporter to provide metrics on how many tests decorated with `@ok_to_fail` have passed and failed which can be useful for comparison across historical test runs.

I would suspect this decorator can be misused by overuse, but so can `@ignore`, `exclude`, and removing tests. I see this as just another, but more helpful, tool in the tool chest that ducktape can provide.

andrewhsu:
@imcdo If there is something else I'm not considering, let me know.

imcdo:
Hey @andrewhsu sorry for the late response.  We where thinking of doing something along the lines of https://github.com/confluentinc/ducktape/pull/299 where we rerun tests to determine flakyness.  Let me know your thoughts on this.

ducktape/mark/_mark.py
```diff
@@ -107,6 +107,24 @@ def __init__(self):
         self.injected_args = None
 
 
+class OkToFail(Mark):
```
stan-confluent, line 110: I would suggest renaming this to Flaky. It should never be ok to fail 😄

ducktape/tests/status.py
```diff
@@ -30,3 +30,5 @@ def to_json(self):
 PASS = TestStatus("pass")
 FAIL = TestStatus("fail")
 IGNORE = TestStatus("ignore")
+OPASS = TestStatus("opass")
```
stan-confluent, line 33: For readability purposes, please rename into something like FLAKY_PASS and FLAKY_FAIL

stan-confluent (commented):
Thanks Andrew! 
Some comments on implementation:
Please make sure to update the junit reporter further (because it won't mark failed flaky as a failed, its around line 210). Please also check other reporters. 

As a larger note, I don't think we should mark the run as successful if there were any flaky tests at all, regardless of whether they passed or failed. 
I'm not as familiar with that part of the codebase, but seems like `get_aggregate_success` in result.py needs updating to account for new statuses. I don't think any of the reporters actually reports aggregated status, but I might be wrong, so please check for those as well. 
Like I wonder if there's any way we can mark flaky passed as yellow or something in junit reporter.

ducktape/tests/reporter.py
```diff
@@ -176,8 +178,13 @@ def report(self):
                 testsuite['failures'] += 1
             elif result.test_status == IGNORE:
                 testsuite['skipped'] += 1
+            elif result.test_status == OPASS:
```
stan-confluent, line 181: there's also testsuite.setdefault line that likely needs updating above

stan-confluent:
My other concern is how would the results look in jenkins jobs, and if it's possible to mark them with different color if there were flaky tests there. 
I would suggest different exit codes depending on whether the run passed, failed, passed with flaky or failed with flaky, or something like that

imcdo:
Why do we need to keep the OPASS tally separate from the pass tally?  Why not just track the ok_to_fail failures, and then just keep the passed tests in `passed`

stan-confluent:
> Why do we need to keep the OPASS tally separate from the pass tally? Why not just track the ok_to_fail failures, and then just keep the passed tests in passed

I wouldn't want to keep flaky tests in the same basket as stable tests, even if they happened to pass on that particular run. This would discourage removing of @flaky annotation and mask the underlying issue.

stan-confluent:
@andrewhsu @imcdo upon reviewing #299 , I'm thinking maybe we should call this annotation `@unstable` instead of `@flaky` or `@ok_to_fail`, wdyt?

imcdo:
any updates here @andrewhsu @dotnwat?

ducktape/tests/status.py
stan-confluent: edit: UNSTABLE_PASS and UNSTABLE_FAIL

ducktape/mark/_mark.py
stan-confluent: edit: Unstable 😄

andrewhsu:
converting this PR to draft since there are feedback and merge conflicts that need resolution.

andrewhsu:
closing PR because not needed anymore

PR #299, opened 2022-04-13
Add deflake feature to Ducktape

imcdo:
This pr adds the ability to run ducktape with the flag `--deflake`.  This option allows a user to specify how many times a test can be reran once it fails to determine if the failure is flaky.  As an example if you run a ducktape test with `--deflake 5` and the test fails, it will run up to 5 times until it either passes (and will be then marked as flaky) or fail if it fails all deflake runs.

### DIFF
--- a/ducktape/command_line/main.py
+++ b/ducktape/command_line/main.py
@@ -200,7 +200,8 @@ def main():
         sys.exit(1)
 
     # Run the tests
-    runner = TestRunner(cluster, session_context, session_logger, tests)
+    deflake_num = max(1, args_dict['deflake'])
+    runner = TestRunner(cluster, session_context, session_logger, tests, deflake_num)
     test_results = runner.run_all_tests()
 
     # Report results
--- a/ducktape/command_line/main.py
+++ b/ducktape/command_line/main.py
@@ -200,9 +200,10 @@ def main():
         sys.exit(1)
 
     # Run the tests
+    deflake_num = args_dict['deflake']
     if deflake_num < 1:
         session_logger.warning("specified number of deflake runs specified to be less than 1, running without deflake.")
-    deflake_num = max(1, args_dict['deflake'])
+    deflake_num = max(1, deflake_num)
     runner = TestRunner(cluster, session_context, session_logger, tests, deflake_num)
     test_results = runner.run_all_tests()
 
--- a/ducktape/command_line/main.py
+++ b/ducktape/command_line/main.py
@@ -200,6 +200,8 @@ def main():
         sys.exit(1)
 
     # Run the tests
+    if deflake_num < 1:
+        session_logger.warning("specified number of deflake runs specified to be less than 1, running without deflake.")
     deflake_num = max(1, args_dict['deflake'])
     runner = TestRunner(cluster, session_context, session_logger, tests, deflake_num)
     test_results = runner.run_all_tests()
--- a/ducktape/services/service.py
+++ b/ducktape/services/service.py
@@ -307,6 +307,11 @@ class Service(TemplateRenderer):
             node.account.logger = None
             self.cluster.free(node)
 
+        try:
+            self.registry.remove(self)
+        except AttributeError:
+            pass
+
         self.nodes = []
 
     def run(self):
--- a/ducktape/services/service.py
+++ b/ducktape/services/service.py
@@ -102,7 +102,6 @@ class Service(TemplateRenderer):
         self._stop_time = -1
         self._stop_duration_seconds = -1
         self._clean_time = -1
-        self._registries = []
 
         self._initialized = False
         self.service_id_factory = ServiceIdFactory(self)
@@ -342,9 +341,6 @@ class Service(TemplateRenderer):
             node.account.logger = None
             self.cluster.free(node)
 
-        for registry in self._registries:
-            registry.remove(self)
-
     def run(self):
         """Helper that executes run(), wait(), and stop() in sequence."""
         self.start()
@@ -383,9 +379,6 @@ class Service(TemplateRenderer):
             svc.wait()
         for svc in args:
             svc.stop()
-
-    def add_registry(self, registry):
-        self._registries.append(registry)
 
     def to_json(self):
         return {
--- a/ducktape/services/service.py
+++ b/ducktape/services/service.py
@@ -23,6 +23,38 @@ import tempfile
 import time
 
 
+class AbstractServiceIdFactory:
+    def generate_service_id(self):
+        raise NotImplementedError()
+
+
+class ServiceIdFactory(AbstractServiceIdFactory):
+    def __init__(self, service):
+        self.service = service
+
+    def generate_service_id(self):
+        return "{service_name}-{service_number}-{service_id}".format(
+            service_name=self.service.__class__.__name__,
+            service_number=self.service._order,
+            service_id=id(self.service)
+        )
+
+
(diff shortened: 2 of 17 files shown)


stan-confluent:
Thanks Ian!
I'll do a more in depth review later, but just to brainstorm the strategy - I'm wondering if we want to stop at the first pass or actually run the test all 5 times and record how many times it passed and how many times it failed. 
Also, are the test statuses reported separately for each failure in this case and the successful run is marked as flaky?

And more complex thing to resolve would be on how to combine this with the @flaky annotation PR by Andrew.

imcdo:
I think having it run 5 times would be a little too over the top, as it would literally just be repeating the tests (which we can already do in ducktape with the `repeat` flag).  This is to determine specifically if a test fails, does it fail consistently, so no need in rerunning tests that pass the first time as they do not fail consistently.

stan-confluent:
What if the test passes on the first go, but then goes on to fail on e.g. third attempt? Because we're talking about the flaky tests, passing on the first attempt doesn't mean it is actually stable. 
In fact, you're making a good point about using repeat flag - we could simply run tests with repeat and then if the test is not stable, mark it as unstable or something.

imcdo:
I feel they have different use cases.  One is when you are trying to catch flaky tests, and the other is determine if a test is blocking you so you need to determine if it is flaky or not.  This pr is ment to address that use case where you want to validate that tests that fail have real failures or if they are just flaky, the other case can be covered by the --repeat flag.

ducktape/tests/runner_client.py
```diff
@@ -102,82 +105,89 @@ def run(self):
...
-            # Tell the server we are finished
-            self._do_safely(lambda: self.send(self.message.finished(result=result)),
-                            "Problem sending FINISHED message for " + str(self.test_metadata) + ":\n")
-            # Release test_context resources only after creating the result and finishing logging activity
-            # The Sender object uses the same logger, so we postpone closing until after the finished message is sent
-            self.test_context.close()
-            self.test_context = None
-            self.test = None
+        num_runs = self.deflake_num
+        while test_status == FAIL and num_runs > 0:
```
stan-confluent, line 113: This is a little bit difficult to read, with num_runs counting backwards and num_runs == self.deflake_num - 1 later on, when we can simply do num_runs < deflake_num and then num_runs == 1 below.

stan-confluent (commented):
Couple of implementation comments.

On the second thought, I think I would agree to stop as soon as test starts passing, since we don't care how often it passes, just the fact that its flaky.

As for how to integrate it with @flaky annotation, I'm thinking we shouldn't worry about it too much either - it's ok to have separate concepts.

We might want to consider renaming one of these concepts from flaky to unstable, either this one or the one @andrewhsu prepared.

ducktape/command_line/main.py
```diff
@@ -200,7 +200,8 @@ def main():
         sys.exit(1)
 
     # Run the tests
-    runner = TestRunner(cluster, session_context, session_logger, tests)
+    deflake_num = max(1, args_dict['deflake'])
```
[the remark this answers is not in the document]
imcdo: this also catches users specifiying 0 and -1, by default its 1, so the only case its less than 1 is if a user specifies it.

ducktape/services/service.py
```diff
@@ -307,6 +307,11 @@ def free(self):
             node.account.logger = None
             self.cluster.free(node)
 
+        try:
+            self.registry.remove(self)
```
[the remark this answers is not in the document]
imcdo: in the past, service registry tracked what services where running but didn't remove them when services stoped becuase when a test was finished there was no need to.   This gives a call back to the registry to clean up when a services is shut down.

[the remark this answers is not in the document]
imcdo: specificly the problem was that the services where still being tracked and even though the nodes where freed, the service registry still thought that the cluster was occupied when rerunning a test from the client level.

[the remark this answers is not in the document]
stan-confluent: how did it work with --repeat?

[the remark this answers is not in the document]
stan-confluent: did `--repeat` simply schedule the test multiple times on the runner (as opposed to the client)?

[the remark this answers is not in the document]
imcdo: yes

ducktape/command_line/main.py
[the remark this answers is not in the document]
stan-confluent: I would probably want to raise an exception if the user specified something wrong, but your way is fine too. Might be good to at least print a warning that user's value is ignored though.

setup.py
```diff
@@ -31,6 +31,14 @@ def run_tests(self):
         self.run_command('flake8')
         sys.exit(errno)
 
+test_req = [
```
stan-confluent, line 34: I would suggest extracting both require and test require to requirements.txt and requirements-test.txt but probably in a separate PR. 
Having them in txt would help when doing releases

imcdo: thats kinda why i did these changes.  You can now do `pip install .[test]` to install test requirements

ducktape/tests/runner_client.py
```diff
@@ -102,82 +105,91 @@ def run(self):
...
+                if hasattr(self, "services"):
+                    service_errors = self.test_context.services.errors()
+                    if service_errors:
+                        summary.extend(["\n\n", service_errors])
+                num_runs += 1
+
+        summary = "".join(summary)
+        test_status, summary = self._check_cluster_utilization(test_status, summary)
+
+        # for flaky tests, we report the start and end time of the successfull run, and not the whole run period
```
stan-confluent, line 186: Typo: single `l`

ducktape/tests/runner_client.py
```diff
@@ -102,82 +107,105 @@ def run(self):
...
+                for service in self.test_context.services:
+                    if self.deflake_num > 1:
+                        service.service_id_factory = MultiRunServiceIdFactory(service, num_runs)
+                    all_services.append(service)
+
+                self.teardown_test(teardown_services=not self.session_context.no_teardown, test_status=test_status)
+
+                stop_time = time.time()
+
+                if hasattr(self.test_context, "services"):
```
stan-confluent, line 168: move it before the teardown, otherwise service registry is empty by the time you get here (its wiped in teardown I believe)

imcdo: Ive moved the free_nodes out of teardown, and placed it after here, so that way we can capture errors from cleanup as well

ducktape/tests/runner_client.py
```diff
@@ -304,7 +329,7 @@ def send(self, event, blocking=True):
 
                 if sockets.get(self.socket) == zmq.POLLIN:
                     reply = self.socket.recv()
-                    if reply:
+                    if reply:
```
stan-confluent: nit: extra space at the end. Maybe run it through autopep8?

ducktape/services/service.py
```diff
@@ -23,6 +23,38 @@
 import time
 
 
+class AbstractServiceIdFactory:
+    def generate_service_id(self):
+        raise NotImplementedError()
+
+
+class ServiceIdFactory(AbstractServiceIdFactory):
```
stan-confluent, line 31: You can make this object shared between instances (or even a singleton) if you make `generate_service_id()` method accept service as an argument.
Same for multirun factory, you can create it once per run in the runner_client rather than once per service per run.

stan-confluent (approved):

PR #300, opened 2022-04-22
wrap long stack trace lines

stan-confluent:
- wrap long stack trace lines:
Screen Shot 2022-04-22 at 2 37 25 PM

- update vagrantfile to work with latest virtualbox:
```
==> ducktape1: Clearing any previously set network interfaces...
The IP address configured for the host-only network is not within the
allowed ranges. Please update the address used to be within the allowed
ranges and run the command again.

  Address: 192.168.50.151
  Ranges: 192.168.56.0/21

Valid ranges can be modified in the /etc/vbox/networks.conf file. For
more information including valid format see:

  https://www.virtualbox.org/manual/ch06.html#network_hostonly

```

### DIFF
--- a/Vagrantfile
+++ b/Vagrantfile
@@ -51,7 +51,7 @@ Vagrant.configure(VAGRANTFILE_API_VERSION) do |config|
     name = "ducktape" + i.to_s
     config.vm.define name do |worker|
       worker.vm.hostname = name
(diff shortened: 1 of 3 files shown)


imcdo (approved):
Looks good tysm!

PR #308, opened 2022-05-11
relative paths when printing tests to rerun

stan-confluent:
Different jobs on jenkins have different absolute paths; so this change makes it print relative paths instead.
Tested with vagrant locally; have NOT tested it with jenkins actually :) Can do it later today if needed, but I think it should be ok anyway


imcdo:
Should this target 0.8.x?

stan-confluent:
>  Should this target 0.8.x?

yes!

PR #310, opened 2022-05-12
fixed #284 properly

### LINKED ISSUE #284: Tests in explicitly provided files ignored

stan-confluent:
Previous fix for #284 broke glob loader for test suites - if you specified /folder/* in the test suite, *all* files under that glob will be attempted to be loaded, which would break when we try to load non-python or non-test files. 
This PR introduces a proper fix, which checks if a symbol refers to a single file first, if so, just load it directly, else use the usual file name pattern matching.

### DIFF
--- a/ducktape/tests/loader.py
+++ b/ducktape/tests/loader.py
@@ -389,15 +389,22 @@ class TestLoader(object):
         """
         test_files = []
         self.logger.debug('Looking for test files in {}'.format(path_or_glob))
+        # glob is safe to be called on non-glob path - it would just return that same path wrapped in a list
         expanded_glob = glob.glob(path_or_glob)
         self.logger.debug('Expanded {} into {}'.format(path_or_glob, expanded_glob))
-        # glob is safe to be called on non-glob path - it would just return that same path wrapped in a list
+
+        def maybe_add_test_file(f):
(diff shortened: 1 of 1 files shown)


stan-confluent:
cc @vp-elitnet

imcdo (approved):
LGTM thanks stan for the fix

stan-confluent:
Now the more I think about it, I think another fix is needed, for the cases where you specify smth like this in the test suite:
```yaml
- /some/path/*.py
- /some/path/*/whatever.py
- /some/path/file.py
```
- should the first case attempt to load all python files or only the ones matching the wildcard? 
- the second and third cases should load the paths without applying the regex matcher.

Clearly we need a more complex logic in there.

PR #311, opened 2022-05-12
support ssh timeout

stan-confluent:
Add support to the `connecttimeout` param (see https://linux.die.net/man/5/ssh_config) in RemoteAccount ssh configuration, and use that value when creating ssh client.
Primary use is to speed up our ssh checker test. Mocking ssh client for the checker is possible, but why not expose an extra configuration parameter? 
This, however, means that any clusters loaded from json or vagrant will respect connecttimeout parameter, hence I'd like to test it with real jenkins jobs before merging. This is also why I don't want to target 0.7.x with this change, maybe would be a good idea to bump this to 0.9?


imcdo (approved):
LGTM thanks stan

stan-confluent:
I'm merging it into master and will release as part of 0.9.x, so that if it does cause issues (which is unlikely, but still), people can still use 0.8.x branch.

PR #313, opened 2022-05-20
cleaner logging for a passing test

stan-confluent:
We don't need the `TEST` bit; and we can simply print the string value rather than 'to_json' - because it evaluates to that string value anyway. If we wanted to future-proof this, we would change all the places where we print the status, but I don't think there's any point in doing so

### DIFF
--- a/ducktape/tests/runner_client.py
+++ b/ducktape/tests/runner_client.py
@@ -125,7 +125,7 @@ class RunnerClient(object):
                 if test_status == PASS and num_runs > 1:
                     test_status = FLAKY
 
-                msg = "{}".format(test_status.to_json())
+                msg = str(test_status.to_json())
                 if summary:
                     msg += ": {}".format(summary)
                 if num_runs != self.deflake_num:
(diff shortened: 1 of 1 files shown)


ducktape/tests/runner_client.py
```diff
@@ -175,7 +175,8 @@ def _do_run(self, num_runs):
             data = self.run_test()
 
             test_status = PASS
-            self.log(logging.INFO, "{} TEST".format(test_status.to_json()))
+            # test_status.to_json() simply prints PASS - str(self).upper() - so might as well just print PASS directly
+            self.log(logging.INFO, "PASS")
```
imcdo, line 179: maybe this print should be moved, as it should print flaky for a flaky test not pass.

ducktape/tests/runner_client.py
```diff
@@ -121,11 +121,21 @@ def run(self):
                 self.log(logging.INFO, "on run {}/{}".format(num_runs, self.deflake_num))
                 start_time = time.time()
                 test_status, summary, data = self._do_run(num_runs)
+
+                if test_status == PASS and num_runs > 1:
+                    test_status = FLAKY
+
+                msg = "{}".format(test_status.to_json())
```
imcdo, line 128: nit: this is just calling `str(test_status.to_json())`

imcdo (approved):
Looks good thanks stan! Left a single nit.

PR #314, opened 2022-05-25
update fetch_externally_routable_ips to use any network device

imcdo:
instead of expecting a explicit network device, use a generic one based on the devices found on the remote machine.

### DIFF
--- a/ducktape/cluster/linux_remoteaccount.py
+++ b/ducktape/cluster/linux_remoteaccount.py
@@ -12,8 +12,9 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
+from deprecated import deprecated
 from ducktape.cluster.cluster_spec import LINUX
-from ducktape.cluster.remoteaccount import RemoteAccount
+from ducktape.cluster.remoteaccount import RemoteAccount, RemoteAccountError
 
 
 class LinuxRemoteAccount(RemoteAccount):
@@ -30,11 +31,37 @@ class LinuxRemoteAccount(RemoteAccount):
         This is an imperfect heuristic, but should work for simple local testing."""
         return self.hostname == "localhost" and self.user is None and self.ssh_config is None
 
-    def fetch_externally_routable_ip(self, is_aws):
-        if is_aws:
-            cmd = "/sbin/ifconfig eth0 "
-        else:
-            cmd = "/sbin/ifconfig eth1 "
-        cmd += r"| grep 'inet ' | tail -n 1 | egrep -o '[0-9\.]+' | head -n 1 2>&1"
-        output = "".join(self.ssh_capture(cmd))
-        return output.strip()
+    # deprecated, please just use the remove externally routable device
+    @deprecated
+    def fetch_externally_routable_ip(self, is_aws=None):
+        if is_aws is not None:
+            self.logger.warning("fetch_externally_routable_ip: is_aws is a depricated flag, and does nothing")
+
+        devices = [
+            device
+            for device in self.sftp_client.listdir('/sys/class/net')
+            if device != 'lo' # do not include local device
+        ]
+
+        self.logger.debug("found devices: {}".format(devices))
+
+        if len(devices) == 0:
+            raise RemoteAccountError("Couldn't find any network devices")
+
+        fmt_cmd = (
+            "/sbin/ifconfig {device} | " 
+            "grep 'inet ' | "
+            "tail -n 1 | "
+            r"egrep -o '[0-9\.]+' | "
+            "head -n 1 2>&1"
+        )
+
+        ips = [
+            "".join(
+                self.ssh_capture(fmt_cmd.format(device=device))
+            ).strip()
+            for device in devices
+        ]
+        self.logger.debug("found ips: {}".format(ips))
+        self.logger.debug("returning the first ip found")
+        return next(iter(ips))
--- a/requirements.txt
+++ b/requirements.txt
@@ -17,3 +17,4 @@ six==1.12.0
 pynacl==1.4.0
 filelock==3.2.1
 cryptography==3.3.2
+Python-Deprecated~=1.1.0


ducktape/cluster/linux_remoteaccount.py
```diff
@@ -30,11 +31,37 @@ def local(self):
...
+
+        devices = [
+            device
+            for device in self.sftp_client.listdir('/sys/class/net')
+            if device != 'lo' # do not include local device
+        ]
+
+        self.logger.debug("found devices: {}".format(devices))
+
+        if len(devices) == 0:
```
stan-confluent, line 48: nit: `if not devices` is more pythonic

ducktape/cluster/linux_remoteaccount.py
```diff
@@ -30,11 +31,37 @@ def local(self):
...
+
+        fmt_cmd = (
+            "/sbin/ifconfig {device} | " 
+            "grep 'inet ' | "
+            "tail -n 1 | "
+            r"egrep -o '[0-9\.]+' | "
+            "head -n 1 2>&1"
+        )
+
+        ips = [
```
stan-confluent, line 59: I'm not convinced of this formatting. Try running this through autopep8

ducktape/cluster/linux_remoteaccount.py
```diff
@@ -30,11 +31,37 @@ def local(self):
...
+            "".join(
+                self.ssh_capture(fmt_cmd.format(device=device))
+            ).strip()
+            for device in devices
+        ]
+        self.logger.debug("found ips: {}".format(ips))
+        self.logger.debug("returning the first ip found")
+        return next(iter(ips))
```
stan-confluent, line 67: why not just `ips[0]`?

ducktape/cluster/linux_remoteaccount.py
```diff
@@ -30,11 +31,37 @@ def local(self):
...
 
-    def fetch_externally_routable_ip(self, is_aws):
-        if is_aws:
-            cmd = "/sbin/ifconfig eth0 "
-        else:
-            cmd = "/sbin/ifconfig eth1 "
-        cmd += r"| grep 'inet ' | tail -n 1 | egrep -o '[0-9\.]+' | head -n 1 2>&1"
-        output = "".join(self.ssh_capture(cmd))
-        return output.strip()
+    # deprecated, please just use the remove externally routable device
```
stan-confluent, line 34: what's `remove externally routable device`? A bit more details on comment would be nice

stan-confluent (commented):
Thanks Ian! Looks ok, but the tests fail and I'm not sure about this formatting (try autopep8)

stan-confluent:
A more meta question is this - what specific issue is this resolving and do we want it in all branches or maybe make it 0.9.x only?

ducktape/cluster/linux_remoteaccount.py
imcdo: i tend to prefer this as its more generic, it can take any iterable and get the first one, and doesn't require indexability in the case that ips gets changed from a list.

ducktape/cluster/linux_remoteaccount.py
imcdo: it's fine with it.

imcdo:
> A more meta question is this - what specific issue is this resolving and do we want it in all branches or maybe make it 0.9.x only?
seems that some people use this in their test, and when changing ec2 instance types/base amis the network device name changes and is rarely actually these two devices.  Realistically people should only use the externally routable ip that is set in their cluster (hence the deprecation) but this fix is for people with current tests that use it and who are trying to upgrade their hardware.

stan-confluent (approved):

ducktape/cluster/linux_remoteaccount.py
```diff
@@ -30,11 +31,38 @@ def local(self):
...
-        else:
-            cmd = "/sbin/ifconfig eth1 "
-        cmd += r"| grep 'inet ' | tail -n 1 | egrep -o '[0-9\.]+' | head -n 1 2>&1"
-        output = "".join(self.ssh_capture(cmd))
-        return output.strip()
+    # deprecated, please use the self.externally_routable_ip that is set in your cluster
+    @deprecated
+    def fetch_externally_routable_ip(self, is_aws=None):
+        if is_aws is not None:
+            self.logger.warning("fetch_externally_routable_ip: is_aws is a depricated flag, and does nothing")
```
lbradstreet, line 38: nit: s/depricated/deprecated/

lbradstreet:
This LGTM, but I fired off a run that previously failed to be sure. I'll approve once it's done.

lbradstreet (approved):
LGTM, thanks for the fix!

lbradstreet:
There's some conflicts with requirements btw.

ducktape/cluster/windows_remoteaccount.py
```diff
@@ -96,10 +96,7 @@ def winrm_client(self):
 
         return self._winrm_client
 
-    def fetch_externally_routable_ip(self, is_aws):
-        if not is_aws:
-            raise NotImplementedError("Windows is only supported in AWS.")
-
+    def fetch_externally_routable_ip(self, is_aws=None):
```
stan-confluent, line 99: So do we support windows in aws now?

stan-confluent (commented):
Seeing how we ran into issues with other changes in the past, I'm not convinced we want such changes in 0.7 or 0.8 tbh. 
We can discuss this offline too.

ducktape/cluster/windows_remoteaccount.py
imcdo: in the past we only supported windows in aws for vagrant clusters, whats not made clear is this code path is only executed from the vagrant cluster file, so its not a great representation of whats actually supported.  This pr makes things a bit more generic.

imcdo:
the expectation before had tied the hardware to vagrant clusters, and this pr unties them, so I believe its worth wile to merge this in old branches as imo ducktape shouldn't make any assumptions about hardware in any version

stan-confluent:
I totally support the motivation behind this change, ducktape shouldn't make assumptions about hardware indeed. However, I don't want us making potentially breaking changes in the patch versions, especially since ducktape is used in other projects outside of confluent as well. I may be wrong here, of course, so let's continue this conversation offline.

lbradstreet:
@imcdo I tested this change with my branch where the IP discovery and interface discovery were previously failing and it works well and fixes the issues. For some of the tests where I need to detect the default interface, I ended up doing `node.account.get_external_accessible_network_devices()[0]` which is slightly awkward but works fine.

PR #320, opened 2022-06-01
command to print total number of nodes

stan-confluent:
```shell
 ducktape systests/cluster/test_remote_account.py::TestClusterSpec --total-size --collect-only 2>/dev/null
5
```

### DIFF
--- a/ducktape/command_line/main.py
+++ b/ducktape/command_line/main.py
@@ -160,6 +160,11 @@ def main():
         sys.exit(1)
 
     if args_dict["collect_only"]:
+        if args_dict["total_size"]:
+            total_nodes = sum(test.expected_cluster_spec.size() for test in tests)
+            print(total_nodes)
+            sys.exit(0)
+
         print("Collected %d tests:" % len(tests))
         for test in tests:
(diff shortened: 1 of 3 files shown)


imcdo (commented):
Why have a seperate flag thats part of collect only, it seems wierd to have the dependancy, why not just have:
```
ducktape tests --collect-num-nodes ... 
``` 
instead

ducktape/command_line/main.py
```diff
@@ -165,6 +165,11 @@ def main():
             print("    " + str(test))
         sys.exit(0)
 
+    if args_dict["collect_num_nodes"]:
+        total_nodes = sum(test.expected_cluster_spec.size() for test in tests)
```
abolipai99, line 169: We can use test.expected_num_nodes here. It's returns the same value right now but would be helpful if there's any change in it in the future.

stan-confluent: good idea, changed

imcdo (approved):

imcdo:
@stan-confluent is there anything blocking this from being merged?

PR #328, opened 2022-06-17
updated vagrant to ubuntu 20 and fixed network discovery 

stan-confluent:
Updated vagrantfile to use ubuntu focal, rather than the ancient ubuntu 14
This required updating the network discovery code to account for other possible interface names, such as enp or enx.
See this stackexchange post - https://unix.stackexchange.com/q/134483
Tested by running ducktape systests on vagrant locally on a mac

### DIFF
--- a/Vagrantfile
+++ b/Vagrantfile
@@ -23,7 +23,7 @@ VAGRANTFILE_API_VERSION = "2"
 enable_dns = false
 num_workers = 3
 ram_megabytes = 300
-base_box = "ubuntu/trusty64"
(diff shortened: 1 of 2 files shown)


imcdo (approved):
looks good thanks stan!  Will this allow for updated paramiko?
(not that we should for 0.7)

stan-confluent:
yeah, I've tested it locally with paramiko 2.10 on master before pushing the PR back onto 0.7.x and testing there with older paramiko

PR #340, opened 2022-07-15
Stan 0.10.0

stan-confluent:


ducktape/cluster/json.py
```diff
@@ -116,7 +116,7 @@ def do_alloc(self, cluster_spec):
         try:
             good_nodes, bad_nodes = self._available_accounts.remove_spec(cluster_spec)
         except InsufficientHealthyNodesError as e:
-            self._bad_accounts.add_nodes(e)
+            self._bad_accounts.add_nodes(e.bad_nodes)
```
stan-confluent, line 119: this was a bug, plain and simple. I'm surprised unit tests didn't catch it

ducktape/tests/runner.py
```diff
@@ -210,19 +210,23 @@ def run_all_tests(self):
...
+                            f"Couldn't schedule test context {next_test_context} but we'll keep trying",
+                            exc_info=True
+                        )
                         self._check_unschedulable()
-                        continue
-
-                    # only remove the test from the scheduler when we have successfully allocated a subcluster for it
-                    self.scheduler.remove(next_test_context)
-                    self._run_single_test(next_test_context)
+                    else:
```
stan-confluent, line 226: I feel like this construct is cleaner than `continue`, though functionally should be identical

imcdo (approved):
Thanks stan

stan-confluent:
Thanks for the review Ian.
I'm currently running a full set of kafka and confluent system tests, just to make sure we don't repeat the same mistakes as before. Once those are done, I'll merge this and cut the new release, but it's safer to do it on Mondays.

PR #358, opened 2022-08-10
Fix kill_process system test

stan-confluent:
I'm not sure what changed - it was certainly not a ducktape code change, nor it was a vagrant ubuntu base box change (because I've tested with older boxes explicitly, unless vagrant can update some parts of older boxes without bumping their version). It could be that something silently updated vagrant itself? Or a MacOS change? Unlikely though.
Anyways, right now if I run ducktape system tests on my local machine, then the `test_kill_process` would fail saying `operation not permitted`. Turns out when the test is doing `grep -i nc`, the first line in the output is `/usr/sbin/irqbalance --foreground`, which cannot be killed (well, maybe with sudo? not through normal means). So I'm updating the test to be more precise about what it `grep`s for, even if I don't understand where the `irqbalance` is even coming from.


systests/cluster/test_remote_account.py
```diff
@@ -579,9 +579,10 @@ def test_monitor_log_exception(self):
     @cluster(num_nodes=1)
     def test_kill_process(self):
         """Tests that kill_process correctly works"""
+        grep_str = '"nc -l -p 5000"'
 
         def get_pids():
-            pid_cmd = "ps ax | grep -i nc | grep -v grep | awk '{print $1}'"
+            pid_cmd = f"ps ax | grep -i {grep_str} | grep -v grep | awk '{{print $1}}'"
```
imcdo, line 585: nit: should we quote the grep string here for clairity up to you will approve regradless?

imcdo (approved):

PR #359, opened 2022-08-16
Add debugging for hung tests

ajfabbri:
To allow us to debug hard-to-reproduce "hung test" failures, this PR adds special handling of test timeouts which, for any still-running child processes:

- Sends a SIGUSR1 which causes them to print stack traces of all
threads.
- Sends a SIGINT and wait for children processes to exit.

The goals are to provide diagnostics on where child test clients are stuck, and to allow the ducktape processes to exit, which should allow gathering logs.

### DIFF
--- a/ducktape/command_line/main.py
+++ b/ducktape/command_line/main.py
@@ -18,12 +18,17 @@ import importlib
 import json
 import os
 import random
+import signal
+from time import sleep
+from typing import Optional
 from six import iteritems
 import sys
 import traceback
 
 from ducktape.command_line.defaults import ConsoleDefaults
 from ducktape.command_line.parse_args import parse_args
+from ducktape.errors import TimeoutError
+from ducktape.tests.result import TestResults
 from ducktape.tests.loader import TestLoader, LoaderException
 from ducktape.tests.loggermaker import close_logger
 from ducktape.tests.reporter import SimpleStdoutSummaryReporter, SimpleFileSummaryReporter, \
@@ -189,7 +194,26 @@ def main():
         session_logger.warning("specified number of deflake runs specified to be less than 1, running without deflake.")
     deflake_num = max(1, deflake_num)
     runner = TestRunner(cluster, session_context, session_logger, tests, deflake_num)
-    test_results = runner.run_all_tests()
+
+    # Detect hung tests
+    test_results: Optional[TestResults] = None
+    try:
+        test_results = runner.run_all_tests()
+    except TimeoutError as e:
+        print(f"Timeout: assuming hung test.\n{e}")
+        for p in runner._client_procs.values():
+            assert p.pid != os.getpid(), "Main pid shouldn't be in client_procs"
+            if p.is_alive():
+                print(f"Hung test stacktrace: Sending SIGUSR1 to child pid {p.pid}")
+                os.kill(p.pid, signal.SIGUSR1)
+
+        sleep(0.2)  # wait for signal handlers to print, might not be necessary?
+        # try to force stuck clients to do cleanup and exit
+        for p in runner._client_procs.values():
+            if p.is_alive():
+                print(f"Sending SIGINT to child pid {p.pid}")
+                os.kill(p.pid, signal.SIGINT)
+                p.join()
 
     # Report results
     reporters = [
@@ -200,12 +224,26 @@ def main():
         JUnitReporter(test_results),
         FailedTestSymbolReporter(test_results)
     ]
+=======
+    runner = TestRunner(cluster, session_context, session_logger, tests)
+
+    # Report results
+    reporters: list[SummaryReporter] = []
+    if test_results:
+        reporters += [
+            SimpleStdoutSummaryReporter(test_results),
+            SimpleFileSummaryReporter(test_results),
+            HTMLSummaryReporter(test_results),
+            JSONReporter(test_results),
+            JUnitReporter(test_results)
+        ]
+>>>>>>> 80360b1 (main: Add debugging for hung tests)
 
     for r in reporters:
         r.report()
 
     update_latest_symlink(args_dict["results_root"], results_dir)
     close_logger(session_logger)
-    if not test_results.get_aggregate_success():
+    if not (test_results and test_results.get_aggregate_success()):
         # Non-zero exit if at least one test failed
         sys.exit(1)
--- a/ducktape/tests/runner.py
+++ b/ducktape/tests/runner.py
@@ -66,10 +66,12 @@ class Receiver(object):
             # use default value of 1800000 or 30 minutes
             timeout = 1800000
         self.socket.RCVTIMEO = timeout
+        t_0 = time.time()
         try:
             message = self.socket.recv()
         except zmq.Again:
-            raise TimeoutError("runner client unresponsive")
+            t_x = time.time()
+            raise TimeoutError(f"runner client unresponsive after {t_x - t_0:.2f} seconds.")
         return self.serde.deserialize(message)
 
     def send(self, event):
--- a/ducktape/tests/runner.py
+++ b/ducktape/tests/runner.py
@@ -255,9 +255,6 @@ class TestRunner(object):
                         self._log(logging.ERROR, err_str)
 
                         # All processes are on the same machine, so treat communication failure as a fatal error
-                        for proc in self._client_procs.values():
-                            proc.terminate()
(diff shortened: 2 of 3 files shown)


ducktape/command_line/main.py
```diff
@@ -189,7 +194,26 @@ def main():
...
     deflake_num = max(1, deflake_num)
     runner = TestRunner(cluster, session_context, session_logger, tests, deflake_num)
-    test_results = runner.run_all_tests()
+
+    # Detect hung tests
+    test_results: Optional[TestResults] = None
+    try:
+        test_results = runner.run_all_tests()
+    except TimeoutError as e:
+        print(f"Timeout: assuming hung test.\n{e}")
```
imcdo, line 203: is this a safe assumption? I think its generally probably going to be the case, but shouldn't this be handled on the test_run client level to check if a test itself is hung?

ducktape/command_line/main.py
```diff
@@ -189,7 +194,26 @@ def main():
...
+                print(f"Hung test stacktrace: Sending SIGUSR1 to child pid {p.pid}")
+                os.kill(p.pid, signal.SIGUSR1)
+
+        sleep(0.2)  # wait for signal handlers to print, might not be necessary?
+        # try to force stuck clients to do cleanup and exit
+        for p in runner._client_procs.values():
+            if p.is_alive():
+                print(f"Sending SIGINT to child pid {p.pid}")
+                os.kill(p.pid, signal.SIGINT)
+                p.join()
```
imcdo, line 216: generally this is great!  would be very useful, but i think the logic should be moved into the termination of the runner object itslef then living in the main script.

ducktape/command_line/main.py
```diff
@@ -200,12 +224,26 @@ def main():
...
+    reporters: list[SummaryReporter] = []
+    if test_results:
+        reporters += [
+            SimpleStdoutSummaryReporter(test_results),
+            SimpleFileSummaryReporter(test_results),
+            HTMLSummaryReporter(test_results),
+            JSONReporter(test_results),
+            JUnitReporter(test_results)
+        ]
+>>>>>>> 80360b1 (main: Add debugging for hung tests)
```
imcdo, line 240: nit: merge?

ducktape/command_line/main.py
```diff
@@ -200,12 +224,26 @@ def main():
         JUnitReporter(test_results),
         FailedTestSymbolReporter(test_results)
     ]
+=======
```
imcdo, line 227: nit: merge?

ducktape/command_line/main.py
imcdo: probably should be removed no?

ducktape/tests/runner.py
```diff
@@ -118,11 +123,22 @@ def __init__(self, cluster, session_context, session_logger, tests, deflake_num,
         # This immutable dict tracks test_id -> test_context
         self._test_context = persistence.make_dict(**{t.test_id: t for t in tests})
         self._test_cluster = {}  # Track subcluster assigned to a particular TestKey
-        self._client_procs = {}  # track client processes running tests
+        # track client processes running tests
+        self._client_procs: dict[str, multiprocessing.Process] = {}
```
imcdo, line 127: nit: should be typing dict
```suggestion
        self._client_procs: Dict[str, multiprocessing.Process] = {}
```

ducktape/tests/runner.py
```diff
@@ -50,6 +51,8 @@ def __init__(self, min_port, max_port):
 
         self.zmq_context = zmq.Context()
         self.socket = self.zmq_context.socket(zmq.REP)
+        # Set a reasonable linger time to help with unclean shutdown
+        self.socket.setsockopt(zmq.LINGER, self.LINGER_MS)
```
imcdo, line 55: Just for my understanding, what are the implications for linger, does it just keep the connection open for LINGER_MS amount of time before raising the zmq.Again?

imcdo: How is this differnet from the `self.socket.RCVTIMEO`

ducktape/tests/runner.py
```diff
@@ -234,13 +250,11 @@ def run_all_tests(self):
                         self._handle(event)
                     except Exception as e:
                         err_str = "Exception receiving message: %s: %s" % (str(type(e)), str(e))
-                        err_str += "\n" + traceback.format_exc(limit=16)
+                        # this is redundant output
+                        # err_str += "\n" + traceback.format_exc(limit=16)
```
imcdo, line 254: agreed, lets just remove the line!

ducktape/tests/runner.py
```diff
@@ -234,13 +250,11 @@ def run_all_tests(self):
...
                         self._log(logging.ERROR, err_str)
 
                         # All processes are on the same machine, so treat communication failure as a fatal error
-                        for proc in self._client_procs.values():
-                            proc.terminate()
-                        self._client_procs = {}
```
imcdo, line 243: this is what previously propogated the SIGTERM's (if you read the docs thats what multiprocessing terminate does with some more redundancy i believe ).  Anyway this is probably where the new cleanup logic should go

ducktape/tests/runner_client.py
```diff
@@ -43,6 +46,8 @@ class RunnerClient(object):
     def __init__(self, server_hostname, server_port, test_id,
                  test_index, logger_name, log_dir, debug, fail_bad_cluster_utilization, deflake_num):
         signal.signal(signal.SIGTERM, self._sigterm_handler)  # register a SIGTERM handler
+        # for debugging stuck processes
+        faulthandler.register(signal.SIGUSR1)
```
imcdo, line 50: nice!

imcdo (changes requested):
Hi @ajfabbri thanks so much for the pr!  This is definitely something that is needed by ducktape today and am excited that you picked this up.  Left a few comments but my biggest request here is that we move the majority of the logic out of main and into runner if possible.  Thanks again

ducktape/tests/runner.py
stan-confluent: technically both would work, but I'd also prefer typing.Dict

ducktape/tests/runner_client.py
```diff
@@ -294,14 +299,14 @@ def log(self, log_level, msg, *args, **kwargs):
 
 
 class Sender(object):
-    REQUEST_TIMEOUT_MS = 3000
+    REQUEST_TIMEOUT_MS = 1000
```
stan-confluent, line 302: Why reduce the request timeout?

stan-confluent (changes requested):
Thanks a lot for adding this, super useful!
As Ian suggested, probably a good idea to put it inside the `runner.py` itself.
Also, please add either a unit test (although I admit it might be difficult) or a hanging systest so that we can validate this functionality before releases.
Thanks!

ducktape/command_line/main.py
ajfabbri: Yep.. missed merge conflict marker, thanks.

ducktape/tests/runner.py
ajfabbri: Yeah, we only work with python >= 3.9, so we use the lower case variants for collections. I'll switch to the python >3.6 style and make these capitalized.

ducktape/tests/runner.py
ajfabbri: Cool. This was my first stab at merging w/ some conflicts. I'll move the hung test logic to the runner and we can see how it looks.

ajfabbri:
Thanks for the reviews!  I will push an update when I get a chance, which will:
- Move hung test handling to runner.
- Explain or remove ZMQ / TCP constant changes.
- Add a test.
- Fix neglected merge markers.
- Address other comments (typing syntax, remove dead code comment, etc.)

ducktape/command_line/main.py
ajfabbri: Based on my search of the code at the time, I think so. Should be clearer when moved to the runner, though, thanks.

imcdo:
Again thanks so much @ajfabbri this is really a much appreciated feature

imcdo:
@ajfabbri any update on this?

ajfabbri:
Hi there. I am no longer working on Kafka stuff so I'm unable to work on this. I will close this PR but I welcome anyone to take over and finish developing the patch.

PR #360, opened 2022-08-23
Store summary of previous runs when deflaking

imcdo:
Motivation:
when getting a flaky test, the summaries are clear due to the last run passing, loosing all data of failed runs.  This PR intends to remedy this by changing the summary behavior of deflake runs.  In a deflake run, all summaries are included, and demarked by the run number and a ~~~~~:

image

Testing:
ran unit tests and ran all system tests

### DIFF
--- a/ducktape/tests/runner_client.py
+++ b/ducktape/tests/runner_client.py
@@ -12,11 +12,13 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
+from collections import defaultdict
 import logging
 import os
 import signal
 import time
 import traceback
+from typing import List, Mapping
 import zmq
 
 from ducktape.services.service import MultiRunServiceIdFactory, service_id_factory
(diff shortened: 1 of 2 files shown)


nagpavan-chilakam (approved):
Just a thought, if the error summary is same for all the runs, it would be good if we can just display that 1 error and number of times it failed with the same reason. If a test failed all 5 times with same reason, displaying the same error summary 5 times may be cumbersome to read, especially if we are running a bunch of tests.
Otherwise looks good

imcdo:
> Just a thought, if the error summary is same for all the runs, it would be good if we can just display that 1 error and number of times it failed with the same reason. If a test failed all 5 times with same reason, displaying the same error summary 5 times may be cumbersome to read, especially if we are running a bunch of tests. Otherwise looks good

This is a good idea but Im not sure a good way to display this, 
would something like this work:
```
summary:
run 1, 3, 5 summary:
....
run 2, 4 summary:
...
run 6: PASSED
```
? Thoughts @nagpavan-chilakam @stan-confluent

stan-confluent:
Looks good overall, thanks Ian!
- your way of printing similar results makes sense (aggregate them using a dict)
- worth adding a unit test

imcdo:
example fail case:
image

example flaky case:
image

stan-confluent (approved):

PR #366, opened 2022-10-13
Runner Client Minor Refactor and Test

imcdo:
The biggest change here to operations is organizational, the runner still calls `run_client` which does the same thing as before, but now `ready` has been extracted from the constructor, this way the constructor doesn't make any network calls.

Also for modern python development typing is now being used, and we have a specific runner client unit test.

Also for this tests mocks have been improved.

For typing reasons, Sender has been moved to the top of the file.

### DIFF
--- a/ducktape/tests/runner_client.py
+++ b/ducktape/tests/runner_client.py
@@ -12,7 +12,7 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-from collections import defaultdict
+from collections import Counter, defaultdict
 import logging
 import os
 import signal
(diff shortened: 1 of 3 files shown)


stan-confluent:
A proper review is underway, but a couple of quick points:
- thanks for taking this on! I might be a pain during the review process, but doesn't mean this effort is not appreciated ;)
- definitely need a minor version bump for this

stan-confluent (approved):
Approved, but please release this as a minor version bump, not patch version. Looks good though, thanks!

PR #367, opened 2022-10-16
print ssh command output on debug level

stan-confluent:
We have a lot of tests that go like:
```python
assert "whatever" in node.account.ssh_output(...)
```
which makes it harder to debug what happened if it failed. With this change, we'll always have both the command and it's output printed on the debug level.


imcdo (approved):

imcdo:
looks good, thanks stan.

PR #371, opened 2022-11-30
handle edge case when wait is called on BackgroundThread that has never been started

stan-confluent:
If BackgroundThread object has never been started, calling `wait()` on it will crash since worker_threads won't contain a thread for this worker. This PR adds a safety check around that case.

We discovered it the hard way, with confluent kafka system tests. Debugging this issue was not straightforward for some reason, since the exception is silenced by ducktape (this is another problem not tackled by this PR).

This PR also adds a couple of useful debug lines which we added when debugging an issue with confluent system tests.

### DIFF
--- a/ducktape/services/background_thread.py
+++ b/ducktape/services/background_thread.py
@@ -87,9 +87,14 @@ class BackgroundThreadService(Service):
 
     def wait_node(self, node, timeout_sec=600):
         idx = self.idx(node)
-        worker_thread = self.worker_threads[idx]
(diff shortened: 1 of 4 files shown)


imcdo (approved):
Thanks stan!

stan-confluent:
Thanks @imcdo !
As a heads up, I'll merge this into 0.11.x and release as a hotfix 0.11.2 version - there are other big changes in the next release

PR #377, opened 2023-01-24
Update to use package_root to fix 0.8.x

imcdo:
follow up to https://github.com/confluentinc/ducktape/pull/376, but targeting the python 3.8 target


stan-confluent (approved):
thanks Ian

PR #378, opened 2023-02-07
Adding nodes used in test summary for HTML report

abolipai99:
This adds the number of nodes used by the test in the HTML report.
Screenshot 2023-02-07 at 10 35 56 PM

### DIFF
--- a/ducktape/tests/reporter.py
+++ b/ducktape/tests/reporter.py
@@ -222,10 +222,10 @@ class JUnitReporter(object):
 class HTMLSummaryReporter(SummaryReporter):
 
     def format_test_name(self, result):
-        lines = ["Module: " + result.module_name,
-                 "Class:  " + result.cls_name,
-                 "Method: " + result.function_name,
-                 "Nodes:  " + str(result.nodes_used)]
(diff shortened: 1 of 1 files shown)


ducktape/tests/reporter.py
```diff
@@ -224,7 +224,8 @@ class HTMLSummaryReporter(SummaryReporter):
     def format_test_name(self, result):
         lines = ["Module: " + result.module_name,
                  "Class:  " + result.cls_name,
-                 "Method: " + result.function_name]
+                 "Method: " + result.function_name,
+                 "Nodes:  " + str(result.nodes_used)]
```
imcdo, line 228: Nice!  Why not show the utilization as well to see if people are effectively using their nodes?
```suggestion
                 "Method: " + result.function_name,
                f "Nodes: {result.nodes_used}/{result.nodes_allocated}"]
```

imcdo (approved):
This is great thanks!  Ill go ahead and approve because it looks good thanks!  Id suggest also showing the test node allocation so people can see if they are effecivly using their nodes.

PR #395, opened 2023-12-06
Parse args from config files independently

imcdo:
# The Problem

Currently if you add any arguments in your config files that take n arguments, such as exclude, and you have no other arguments, these could potentially get merged with your test path arguments.

For example:
.ducktape/config:
```
--exclude c/a_test.py, c/b_test.py
```
and you run:
```
ducktape c
```

the arg parse will interpret c as an additional argument to the exclude, so you will now be excluding all tests.

# The Solution
Simply parse each set of args independently, then merge them into one combined group of args!

# Testing
Unit test added, also tested with our system tests: 
Set config to:
```
--exclude systests/cluster/test_runner_operations.py::SimpleRunnerTest.timeout_test
```

and executed
```
ducktape systests/cluster/test_runner_operations.py --max-parallel 9000
```
which successfully ran our tests, in parallel.

### DIFF
--- a/ducktape/command_line/parse_args.py
+++ b/ducktape/command_line/parse_args.py
@@ -128,6 +128,19 @@ def config_file_to_args_list(config_file):
     return list(itertools.chain(*[line.split() for line in config_lines]))
 
 
+def parse_non_default_args(parser: argparse.ArgumentParser, defaults: dict, args: list) -> dict:
+    """
+    Parse and remove default args from a list of args, and return the dict of the parsed args.
+    """
+    parsed_args = vars(parser.parse_args(args))
+
+    # remove defaults
+    for key, value in defaults.items():
+        if parsed_args[key] == value:
+            del parsed_args[key]
+
+    return parsed_args
+
 def parse_args(args):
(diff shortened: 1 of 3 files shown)


ducktape/command_line/parse_args.py
```diff
@@ -128,6 +128,19 @@ def config_file_to_args_list(config_file):
     return list(itertools.chain(*[line.split() for line in config_lines]))
 
 
+def parse_non_default_args(parser: argparse.ArgumentParser, defaults: dict, args: list) -> dict:
```
stan-is-hate, line 131: why do you need to remove the defaults?

stan-is-hate (dismissed):
Looks good, thanks Ian! Just one comment out of curiosity, but looks good overall and thanks for adding a test.

ducktape/command_line/parse_args.py
imcdo: otherwise the merging doesn't work, as the defaults will override the pervious files contributions. (the parser will always include the defaults when parsing the args)

PR #436, opened 2025-06-08
CPTF-1049 | report.html: Report expected test count in the summary

tsrivatsavs:
### Background
There have been cases where ducktape fails to report all the tests that were part of the list of tests to run. This is usually seen in parallel to the issue of `runner_client unresponsive`, and the corresponding test handled by that runner_client will not be part of any of the reports (report.txt, report.html etc). For example, this is a report of a run that initially collected 1232 tests, but only reported 975:
Screenshot 2025-06-09 at 00 08 28

Hence this code change: When the above issue is hit, we want to flag a warning in the logs, ducktape should exit with return code 1, and explicitly mention this in report.html. Here's an edited version of the above snapshot of report.html:
Screenshot 2025-06-09 at 00 09 26
Screenshot 2025-06-08 at 23 49 02

### Code changes
* main.py: Introduce a new variable: `expected_test_count`, which holds the total number of tests expected to run. Pass this while initiating an instance of `HTMLSummaryReporter`.
* reporter.py: Modify the `HTMLSummaryReporter` to accept this new variable to be displayed in the summary panel of report.html. Also update the `report.html` template.

Snapshot from an actual run:
Screenshot 2025-06-09 at 00 13 06
job: https://semaphore.ci.confluent.io/jobs/4e032d5e-56f4-4543-a0e9-a81f3b6dbcec (accessible only by Confluent engineers)

### Out of scope
Please note: This change only flags if any test is missing from the reports. Ducktape still WOULD NOT list out the tests that are missing.

### DIFF
--- a/ducktape/command_line/main.py
+++ b/ducktape/command_line/main.py
@@ -216,7 +216,8 @@ def main():
     update_latest_symlink(args_dict["results_root"], results_dir)
 
     if len(test_results) < expected_test_count:
-        session_logger.warning(f"All tests were NOT run. Expected {expected_test_count} tests, only {len(test_results)} were run.")
+        session_logger.warning(
+            f"All tests were NOT run. Expected {expected_test_count} tests, only {len(test_results)} were run.")
         close_logger(session_logger)
         sys.exit(1)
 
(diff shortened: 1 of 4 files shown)


milindbishnoi (approved):

saurabhcnf (approved):
Thanks Tarun, LGTM.

PR #470, opened 2026-05-27
[CPTF-1609] Add nested result directory feature to solve OS 255 Byte filename limit 

saurabhcnf:
Jira: https://confluentinc.atlassian.net/browse/CPTF-1609

## Summary
- Adds opt-in `--nested-result-dirs` CLI flag. When set, `TestContext.results_dir()` produces one nested directory per `@matrix`/`@parametrize` argument (`<cls>/<method>/k1=v1/k2=v2/`) instead of the single dotted basename (`<cls>/<method>/k1=v1.k2=v2/`).
- Default behavior unchanged; `test_id`, `test_name`, and `report.json` schema preserved regardless of the flag.
- Keeps each path segment short enough to avoid the OS 255-byte filename limit hit by heavily-parameterized tests (e.g. `ConsumerProtocolMigrationTest::test_consumer_offline_migration` on kafka `8.3.x` / `master`).

## Test plan
- [x] Unit tests in `tests/tests/check_test_context.py` covering flat default, nested basic, sorted-key stability, no/empty/single args, `test_index` placement, per-segment escaping, and `injected_args_name` flag-independence.
- [x] End-to-end verified via muckrake branch-builder against the offending parameterization (a separate muckrake-infra change consumes the flag).
https://semaphore.ci.confluent.io/workflows/9bae5628-810f-40ce-b640-37c5f56c62fe?pipeline_id=747fc6da-e239-4376-83a5-4406bcf8b2e5

Design one-pager: https://confluentinc.atlassian.net/wiki/spaces/OAAC/pages/5600575730

### DIFF
--- a/ducktape/tests/test_context.py
+++ b/ducktape/tests/test_context.py
@@ -187,8 +187,15 @@ class TestContext(object):
             d = os.path.join(d, test_context.cls.__name__)
         if test_context.function is not None:
             d = os.path.join(d, test_context.function.__name__)
-        if test_context.injected_args is not None:
-            d = os.path.join(d, test_context.injected_args_name)
+        if test_context.injected_args:
+            if test_context.session_context.nested_result_dirs:
+                # One nested directory per parameter, sorted by key for stable layout
+                # across reruns and reorderings of @matrix/@parametrize arguments.
+                for k in sorted(test_context.injected_args):
+                    segment = _escape_pathname("%s=%s" % (k, test_context.injected_args[k]))
+                    d = os.path.join(d, segment)
+            else:
+                d = os.path.join(d, test_context.injected_args_name)
         if test_index is not None:
             d = os.path.join(d, str(test_index))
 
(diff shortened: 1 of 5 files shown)


ducktape/tests/test_context.py
```diff
@@ -187,8 +187,15 @@ def results_dir(test_context, test_index):
...
         if test_context.function is not None:
             d = os.path.join(d, test_context.function.__name__)
-        if test_context.injected_args is not None:
-            d = os.path.join(d, test_context.injected_args_name)
+        if test_context.injected_args:
+            if test_context.session_context.nested_result_dirs:
+                # One nested directory per parameter, sorted by key for stable layout
+                # across reruns and reorderings of @matrix/@parametrize arguments.
+                for k in sorted(test_context.injected_args):
+                    segment = _escape_pathname("%s=%s" % (k, test_context.injected_args[k]))
```
mihirchandna, line 195: Minor: Can we add the injected args is not None check again, although we should be fine but let's keep it and let's add/update test to lock this in

mihirchandna (approved):
LGTM, thanks

PR #472, opened 2026-06-23
[CPTF-1633] Add --ignore-node-type flag to bypass heterogeneous node matching

saurabhcnf:
Jira: https://confluentinc.atlassian.net/browse/CPTF-1633

Adds `--ignore-node-type` to ignore the `node_type` from `@cluster(node_type=...)` / `cluster_spec` and allocate any available node of the matching OS — lets tests written for a heterogeneous cluster run on a homogeneous one. Node count and OS are still honored; defaults to off, so existing behavior is unchanged.

Stripping happens at `TestContext.expected_cluster_spec`, the single spec both allocation and scheduling read, so behavior stays consistent. 

Docs updated; 3 unit tests added.


mihirchandna (approved):
LGTM

PR #473, opened 2026-08-10
[CPTF-1662] Report all tests on driver abort, join clients in parallel

saurabhcnf:
### Description
The driver only handled `TimeoutError` from the client socket gracefully — every other failure re-raised and discarded the entire test report, so a run that died partway through was published as green with the missing tests simply absent rather than failed. Any failure receiving or handling a client message now shuts the run down the same way a timeout does: signal the clients so they can tear down and collect logs, mark the tests that were still running or had not yet started as failed, and return a complete report. 

Because the clients are now always signalled, they also stop blocking on sends to a driver that is no longer listening, which previously aborted their teardown and lost log collection. Separately, the driver now waits on all client processes against a single shared deadline instead of one at a time, since sequential waiting made shutdown cost the timeout multiplied by the number of stuck clients; the shared timeout is raised from 120s to 300s because it is now a total budget rather than per-process.

Also included is a test-only fixture that destroys each `Receiver`'s zmq context after every test under `tests/runner`. A `TestRunner` that is built but never run leaves its context bound with an open socket, and garbage collecting one later blocks forever in `Context.term()` — stalling whichever unrelated test happens to trigger the collection. This was already latent on master and was hanging CI non-deterministically. Verified on Linux with the fork start method across Python 3.11, 3.12 and 3.13.

### Issue
Jira: https://confluentinc.atlassian.net/browse/CPTF-1662
Jira: https://confluentinc.atlassian.net/browse/CPTF-1664

### DIFF
--- a/ducktape/tests/runner.py
+++ b/ducktape/tests/runner.py
@@ -48,7 +48,7 @@ from ducktape.tests.test_context import TestContext
 from ducktape.utils.terminal_size import get_terminal_size
 
 DEFAULT_MP_JOIN_TIMEOUT = 30
-DEFAULT_TIMEOUT_EXCEPTION_JOIN_TIMEOUT = 120
+DEFAULT_TIMEOUT_EXCEPTION_JOIN_TIMEOUT = 300
 
 
 class Receiver(object):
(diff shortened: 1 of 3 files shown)


ducktape/tests/runner.py
```diff
@@ -162,21 +162,39 @@ def _terminate_process(self, process: multiprocessing.Process):
...
-        start = time.time()
-        while time.time() - start <= timeout:
-            if not process.is_alive():
-                self.client_report[process_key]["status"] = "FINISHED"
-                break
-            time.sleep(0.1)
-        else:
-            # Note: This can lead to some tmp files being uncleaned, otherwise nothing else should be executed by the
-            #       client after this point.
+        self._join_test_processes([process_key], timeout)
```
mihirchandna, line 165: Old code is raising key error in case of missing process key which this is not, we should preserve this i think

ducktape/tests/runner.py
```diff
@@ -320,6 +338,34 @@ def _report_active_as_failed(self, reason):
...
+
+        try:
+            self._join_test_processes(list(self._client_procs.keys()), self.timeout_exception_join_timeout)
+        except Exception:
+            self._log(logging.ERROR, f"Failed to join client processes\n{traceback.format_exc(limit=16)}")
+        finally:
+            self._client_procs = {}
+
+        self._report_active_as_failed(reason)
+        self._report_remaining_as_failed(reason)
```
mihirchandna, line 362: Should we add exception handling here either failure cause failure of the other one

ducktape/tests/runner.py
```diff
@@ -162,21 +162,39 @@ def _terminate_process(self, process: multiprocessing.Process):
...
+                logging.ERROR,
+                f"after waiting {timeout}s, process {self._client_procs[key].name} failed to complete.  "
+                f"Terminating...",
             )
+
+        for key in process_keys:
+            if key in self._client_procs:
+                self._reap_test_process(key)
+
+    def _reap_test_process(self, process_key):
```
mihirchandna, line 189: Again here should we add exception for each reap such that one bad key does not abort the batch, wdyt?

tests/runner/conftest.py
```diff
@@ -0,0 +1,44 @@
...
+    only run_all_tests() closes it. Collecting such a context later blocks forever in
+    Context.term(), which stalls whichever unrelated test happens to trigger the collection.
+    """
+    receivers = []
+    real_init = Receiver.__init__
+
+    def tracking_init(receiver, *args, **kwargs):
+        real_init(receiver, *args, **kwargs)
+        receivers.append(receiver)
+
```
mihirchandna, line 36: Nit: realistically this won't be hit but ideally we should append to receivers before calling real_init.
We can ignore this as well just wanted to get your thoughts

mihirchandna (approved):

PR #474, opened 2026-08-19
[CPTF-1671] Record the test's own node count in driver-built results

saurabhcnf:
### Description
When the driver builds a `TestResult` itself — because a test never started, was still running when the run aborted, or the whole run is unschedulable — it used the driver's own `test_context`, whose `.cluster` is the entire cluster rather than the subcluster allocated to that test. `TestResult` derives `nodes_allocated` and `nodes_used` from `len(test_context.cluster)`, so every driver-fabricated result recorded the full cluster size instead of the test's declared node count. In a 200-node run this showed up as a test declaring `@cluster(num_nodes=8)` reporting `nodes_allocated=200`, and it corrupts anything that sizes clusters from historical reports — muckrake's node optimizer read that value as the test's requirement and provisioned 155 pods for a single 8-node test.

`TestResult` now accepts optional `nodes_allocated`/`nodes_used` overrides, defaulting to the existing behavior so client-built results are unaffected. The three driver paths pass explicit values: the subcluster size (falling back to the test's declared node count if none was allocated yet) for a test that was actively running, and the declared node count with `nodes_used=0` for tests that never ran at all.

### Issue
Jira: https://confluentinc.atlassian.net/browse/CPTF-1671

### DIFF
--- a/ducktape/tests/result.py
+++ b/ducktape/tests/result.py
@@ -41,15 +41,21 @@ class TestResult(object):
         data=None,
         start_time=-1,
         stop_time=-1,
+        nodes_allocated=None,
+        nodes_used=None,
     ):
         """
-        @param test_context  standard test context object
-        @param test_status   did the test pass or fail, etc?
-        @param summary       summary information
-        @param data          data returned by the test, e.g. throughput
+        @param test_context     standard test context object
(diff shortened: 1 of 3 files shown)


tsrivatsavs (approved):
LGTM

PR #475, opened 2026-08-19
[CPTF-1689] Drop Python 3.8 support and split CI matrix into one env per job

saurabhcnf:
### Description
This PR drops python 3.8 support from ducktape.

Separately, the "Test Python 3.13" job ran bare `tox` with no `-e`, so it executed the entire envlist — three redundant full suite runs (py38/py39/py310, most of which fail immediately from missing interpreters) plus `cover`, `style`, and `docs` — all inside one job, taking close to 700 seconds. It now runs `tox -e py313` like every other version job, and `cover`/`style`/`docs` move into their own new "Quality" block with one job each, matching the existing one-env-per-job pattern.

### Issue
Jira: https://confluentinc.atlassian.net/browse/CPTF-1689

### DIFF
--- a/.semaphore/semaphore.yml
+++ b/.semaphore/semaphore.yml
@@ -18,12 +18,6 @@ blocks:
     dependencies: []
     task:
       jobs:
-        - name: Test Python 3.8
-          commands:
-            - sem-version python 3.8
-            - pip install tox
-            - export PYTESTARGS='--junitxml=test/results-py38.xml'
-            - tox -e py38
         - name: Test Python 3.9
           commands:
             - sem-version python 3.9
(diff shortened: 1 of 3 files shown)


.semaphore/semaphore.yml
tsrivatsavs: do you think we should move to a cheaper `arm64-0` instance?

.semaphore/semaphore.yml
```diff
@@ -53,4 +47,25 @@ blocks:
...
+        - name: Style
+          commands:
+            - sem-version python 3.13
+            - pip install tox
+            - tox -e style
+        - name: Docs
+          commands:
+            - sem-version python 3.13
+            - pip install tox
+            - tox -e docs
```
tsrivatsavs: is this a better idea than having one single block to check coverage, style and docs?

saurabhcnf: Yes, compiled to a single block.

.semaphore/semaphore.yml
saurabhcnf: yes, updated in new revision.

tsrivatsavs (approved):

